Issue #13, opened 2022-12-26
Python type hints


shinich1:
**Is your feature request related to a problem? Please describe.**
enhancement request

**Describe the solution you'd like**
Add type hints to code

**Describe alternatives you've considered**
na

**Additional context**
e.g. https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html

prakhar-kt:
@shinich1 Hi is this issue still open? Can I try to do this?

shinich1:
Hi @prakhar-kt, 
yes it is! Please go ahead - I just created a branch for this issue which I believe you can check out locally :)
forking this repo would also be a good way, do choose whichever you feel comfortable.

prakhar-kt:
thanks @shinich1 . I have started working on it

PR #22, opened 2023-01-07
Added python type hints to transpiler.py

prakhar-kt:

Then, please fill in below:

**Context (if applicable):**

Added python type hints to the transpiler.py file

**Related issue:**

Python type hints #13

### DIFF
--- a/graphix/transpiler.py
+++ b/graphix/transpiler.py
@@ -23,7 +23,7 @@ class Circuit:
         List containing the gate sequence applied.
     """
 
-    def __init__(self, width):
+    def __init__(self, width: int) -> None:
         """
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


prakhar-kt:
Hi, I have added type hints to the transpiler.py file. If this works, I can add the same type hints to the other files as well.
Thank you

shinich1:
thanks - looks nice and comprehensive but it seems the tests failed, I think you'd want to use Tuple and List by `from typing import Tuple, List`, instead of builtin `list` and `tuple` for type hints. Please debug and check by running `pytest` - also run `black -l 120 <filename>` so the lint check passes. then yes it'd be great if you could do the other codes as well.

prakhar-kt:
Thanks for the feedback. I will do as you suggested and let you know.

shinich1:
closing because another pr was created with same topic

Issue #40, opened 2023-03-31
improve pattern minimize space method


masato-fuk:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is.

According to [M. Houshmand et al., PRA 98, 012318 (2018)](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.98.012318), the minimal required number of qubits is n + 1 for graphs with a causal flow. However, current `minimize_space()` method sometimes returns bad answers(require more than n+1 qubits).

**Describe the solution you'd like**

Implement a flow-based-measurement-order generation method which absolutely minimizes required number to n+1.

**Describe alternatives you've considered**

na

**Additional context**

na

PR #41, opened 2023-04-03
40 improve pattern minimize space method

masato-fuk:

Then, please fill in below:

**Context (if applicable):**

na

Reversed the dependency structure between gflow and pattern. In the original version, gflow.py dependes on the Pattern. However, gflow is purely a graph theory concept, so this structure is not preferable.
An improved `minimize_space` method was implemented which ensures `max_space` equals (width+1) if a graph has flow and after the `minimize_space` method applied. See `Pattern.get_measurement_order_from_flow()` for details.

**Related issue:**

#40 

### DIFF
--- a/graphix/pattern.py
+++ b/graphix/pattern.py
@@ -674,6 +674,31 @@ class Pattern:
             edges -= removable_edges
         return meas_order
 
+    def get_measurement_order_from_flow(self):
+        """Return a measurement order generated from flow. If a graph has flow, the minimum 'max_space' of a pattern is guaranteed to width+1.
+
+        Returns
+        -------
+        meas_order: list of int
+            measurement order
+        """
+        nodes, edges = self.get_graph()
+        G = nx.Graph()
+        G.add_nodes_from(nodes)
+        G.add_edges_from(edges)
+        vin = {i for i in range(self.width)}
+        vout = set(self.output_nodes)
+        f, l_k = flow(G, vin, vout)
+        if f is None:
+            return None
+        depth, layer = get_layers(l_k)
+        meas_order = []
+        for i in range(depth):
+            k = depth - i
+            nodes = layer[k]
+            meas_order += nodes
+        return meas_order
+
     def get_measurement_order_from_gflow(self):
         """Returns a list containing the node indices,
         in the order of measurements which can be performed with minimum depth.
@@ -726,7 +751,7 @@ class Pattern:
                 target += 1
         return meas_cmds
 
-    def get_measurement_commnads(self):
+    def get_measurement_commands(self):
         """Returns the list containing the measurement commands,
         in the order of measurements
 
@@ -916,7 +941,9 @@ class Pattern:
         """
         if not self.is_standard():
             self.standardize()
-        meas_order = self._measurement_order_space()
+        meas_order = self.get_measurement_order_from_flow()
+        if meas_order is None:
+            meas_order = self._measurement_order_space()
         self._reorder_pattern(self.sort_measurement_commands(meas_order))
 
     def _reorder_pattern(self, meas_commands):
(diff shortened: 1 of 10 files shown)


graphix/pattern.py
```diff
@@ -674,48 +674,132 @@ def _measurement_order_space(self):
...
+            12,
+            20,
+            21,
+            22,
+            23,
+        ]  # clifford indices that change meas plane from xy to xz
+        for cmd in self.seq:
+            if cmd[0] == "M":
+                if len(cmd) == 7:
+                    if cmd[6] in XYtoXZ_list:
```
shinich1, line 796: don't we need checks of measurement plane in the measurement command?

graphix/pattern.py
```diff
@@ -641,13 +641,13 @@ def connected_edges(self, node, edges):
         return connected
 
     def _measurement_order_space(self):
-        """Determine the optimal measurement order
-         which minimize the `max_space` of the pattern.
+        """Determine a sub-optimal measurement order
```
shinich1, line 644: entire doctoring should read:

Determine measurement order that heuristically optimises the `max_space` of a pattern

shinich1:
Should we add CHANGELOG.md in this repository to keep track of API changes? this time we are changing the way `generate_from_flow()` is imported and it might be good to make this kind of breaking changes listed (also add to readthedocs, c.f. https://qutip.org/docs/latest/changelog.html)

shinich1:
if we agree changelog.md will be useful, I will update PR template to explicitly ask contributors to add to changelog.md

graphix/pattern.py
```diff
@@ -674,48 +674,132 @@ def _measurement_order_space(self):
...
+        for i in range(depth):
+            k = depth - i
+            nodes = layer[k]
+            meas_order += nodes
+        return meas_order
+
+    def get_measurement_order_from_gflow(self):
+        """Returns a list containing the node indices,
+        in the order of measurements which can be performed with minimum depth.
+        Input must be a simple connected graph for the gflow-finding algorithm to work.
```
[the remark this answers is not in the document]
masato-fuk: This 'input' indicates a graph of a pattern. As you may think, this caution statement shouldn't be put here, so I remove it.

tests/test_generator.py
```diff
@@ -0,0 +1,38 @@
...
+        for cmd in pattern.get_measurement_commands():
+            angles[cmd[1]] = cmd[3]
+        pattern2 = generate_from_graph(g, angles, input, pattern.output_nodes)
+        # check that the new one runs and returns correct result
+        pattern2.standardize()
+        pattern2.shift_signals()
+        pattern2.minimize_space()
+        state = circuit.simulate_statevector()
+        state_mbqc = pattern2.simulate_pattern()
+        np.testing.assert_almost_equal(np.abs(np.dot(state_mbqc.flatten().conjugate(), state.flatten())), 1)
```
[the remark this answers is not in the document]
masato-fuk: I think a test for the determinismof a pattern should be separated from the `test_pattern_generation_flow` because their purposes are slightly different from each other. So I'll add a new test.

masato-fuk:
I totally agree with your idea. The CHANGELOG wil surely be helpful for newcomers who are following up with the graphix. By the way, where should I put the CHANGELOG.md file? Also, could you please guide me on how to add it to readthedocs?

graphix/pattern.py
masato-fuk: I updated this method for general measurement planes and also added test for it.

shinich1 (approved):

shinich1:
@masato-fuk please squash and merge. commit message should read, e.g., flow-based pattern minimization method (#40).

PR #44, opened 2023-04-17
add IBMQ interface

d1ssk:
 
Then, please fill in below:

**Context (if applicable):**
This change will eventually be implemented in a separate repository from graphix.

added PatternExecutor class and IBMQBackend class to run the qiskit circuit corresponding to the pattern on IBMQ devices .

**Related issue:**


graphix/clifford.py
```diff
@@ -219,3 +219,30 @@
     ["h", "x", "sdg"],
     ["h", "x", "s"],
 ]
+
+CLIFFORD_TO_QISKIT = [
```
shinich1, line 223: what's the difference from CLIFFORD_TO_QASM3?

d1ssk: Initially, I thought the identity operation in qiskit would be written as "I", so I created a new list, but eventually I found out that it is also "id" in qiskit, so the two became identical.
I think we can delete it.

shinich1:
@d1ssk could you add documentations?

graphix/clifford.py
shinich1: @d1ssk shall we delete this CLIFFORD_TO_QISKIT and use CLIFFORD_TO_QASM3, if they're identical?

d1ssk: I deleted it.

graphix/pattern.py
```diff
@@ -1166,6 +1167,26 @@ def simulate_pattern(self, backend="statevector", **kwargs):
         state = sim.run()
         return state
 
+    def execute_pattern(self, backend, **kwargs):
```
shinich1, line 1170: could you change this to `run_pattern` ?

shinich1: other parts are all looking good to me!

shinich1 (approved):

shinich1:
@d1ssk could you squash and merge?

Issue #62, opened 2023-06-20
[Bug]: pattern optimization result in wrong output state in some cases


shinich1:
**Describe the bug**
#55 did not fully resolve the issue. some class of random Clifford+T circuits sometimes result in wrong output state.

**To Reproduce**
run below multiple times:
```python
from graphix import Circuit
import numpy as np
def genpair(n_qubits, count):
    pairs = []
    for i in range(count):
        choice = [j for j in range(n_qubits)]
        x = np.random.choice(choice)
        choice.pop(x)
        y = np.random.choice(choice)
        pairs.append((x, y))
    return pairs

def get_circuit():
    layers = 3
    n_qubits = 4
    circuit = Circuit(n_qubits)
    clifford_T_choice = [0,1,2,3,4,5,6]
    for i in range(layers):
        for j, k in genpair(n_qubits, 2):
            circuit.cnot(j, k)
        for j in range(n_qubits):
            k = np.random.choice(clifford_T_choice)
            if k == 0: # T gate
                circuit.rz(j, np.pi/4)
            elif k == 1: # T^daggar gate
                circuit.rz(j, -np.pi/4)
            elif k == 2: # H
                circuit.h(j)
            elif k == 3: # S
                circuit.s(j)
            elif k == 4: # X
                circuit.x(j)
            elif k == 5: # Z
                circuit.z(j)
            else:
                pass
    return circuit

circuit = get_circuit()
pattern = circuit.transpile()
pattern.standardize()
pattern.shift_signals()
pattern.perform_pauli_measurements()
pattern.minimize_space()
out_state = pattern.simulate_pattern()
state = circuit.simulate_statevector()
print("overlap of states: ", np.abs(np.dot(state.psi.flatten().conjugate(), out_state.psi.flatten())))
```

**Expected behavior**
statevector simulation of original circuit and simulation of MBQC result in same state.

**Environment (please complete the following information):**
 - OS: [macOS Mojave]
 - Python version: [3.8.1]

**Additional context**

PR #56, opened 2023-06-05
Implement measurement pattern for Toffoli gate

king-p3nguin:
Before submitting, please check the following:
- Make sure you have tests for the new code and that test passes (run `pytest`)
- format added code and tests by `black -l 120 <filename>`
- If applicable, add a line to the [unreleased] part of CHANGELOG.md, following [keep-a-changelog](https://keepachangelog.com/en/1.0.0/).
 
Then, please fill in below:

**Context (if applicable):**

**Description of the change:**

**Related issue:**
- Resolves #49 

also see that checks (github actions) pass.
If lint check keeps failing, try installing black==22.8.0 as behavior seems to vary across versions.

PR #63, opened 2023-06-21
Fix Pauli measurement routine

shinich1:
 
Then, please fill in below:

**Context (if applicable):**

1. Fixed bug in `GraphState` graph state simulator, specifically, 
- `equivalent_graph_E2` method
- bug in toggling 'loop' in `s` method
- behaviour of `measure_x` and `measure_z` methods for isolated nodes
2. Fixed `measure` method of `StatevectorBackend`: 
- now we apply s and t signals by Clifford gate, not change of angles.
- effect of vops in `meas_op` is now done by application of Clifford gate conjugation on measurement operator, instead of interchanging of measurement axes 
3. Fixed pauli measurement preprocessing routine in `pattern.py`
- we treat the signals by applying Clifford gates, not by the change of angles

**Related issue:**
#62 
(#56)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Quantum classifier demo (#57) by @Gopal-Dahale
 
-
 ### Changed
 - fixed a bug in a code snippet isn docs (#59), as pointed out by @zilkf92
 - fixed issue building docs on readthedocs (#61)
+- fixed bug in pauli preprocessing routine and graph state simulator (#63)
+- Second output of `pattern.pauli_nodes` (`non_pauli_node` list) is now list of nodes, not list of lists (commands).
 
 ## [0.2.2] - 2023-05-25
 ### Added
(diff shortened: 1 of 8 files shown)


masato-fuk:
@shinichi1 All changes appear to be good. By the way, I find `vop` makes the code a little complicated. As far as I understand, there shouldn't be any issues with merging vops into measurement angles and planes right after Pauli measurements(if my assumption is incorrect, please disregard this). I  believe it would be better to do so because simple structure let users follow up easily. But, I think this modification should be done in another issue...

masato-fuk:
I'll fix a bug in flow finding algorithm, so please wait a moment.

shinich1:
> By the way, I find `vop` makes the code a little complicated. As far as I understand, there shouldn't be any issues with merging vops into measurement angles and planes right after Pauli measurements(if my assumption is incorrect, please disregard this). I believe it would be better to do so because simple structure let users follow up easily. But, I think this modification should be done in another issue...

I agree, we will need to update angles as well as measurement planes, but that should be doable. I agree it would make things look much simpler.

shinich1:
> I'll fix a bug in flow finding algorithm, so please wait a moment.

let's do this in a separate issue.

PR #64, opened 2023-06-25
Density matrix backend

### LINKED ISSUE #31: Density matrix backend
We cannot simulate with noise/error at the moment.

**Describe the solution you'd like**
Add density matrix backend so that we can simulate various quantum channels (ones described by Kraus ops).

**Describe alternatives you've considered**
na

**Additional context**
Consider how this could be done in TN backend.

shinich1:
 
Then, please fill in below:

**Context (if applicable):**
na

Add density matrix simulator backend and interface for custom noise model.

**Related issue:**
closes #31


tests/test_density_matrix.py
```diff
@@ -0,0 +1,397 @@
...
+            backend.dephase(p)
+            backend.finalize()
+            return backend
+
+        circ = Circuit(1)
+        circ.rx(0, np.pi / 4)
+        pattern = circ.transpile()
+        backend1 = simulate(0, pattern)
+        backend2 = simulate(1, pattern)
+        np.testing.assert_allclose(backend1.state.rho, backend2.state.rho)
```
shinich1, line 379: what is this testing?

tests/test_density_matrix.py
```diff
@@ -0,0 +1,397 @@
...
+        sv_backend.entangle_nodes((0, 1))
+        sv_backend.entangle_nodes((1, 2))
+        sv_backend.measure(pattern.seq[-4])
+        sv_backend.measure(pattern.seq[-3])
+        sv_backend.correct_byproduct(pattern.seq[-2])
+        sv_backend.correct_byproduct(pattern.seq[-1])
+        sv_backend.finalize()
+        np.testing.assert_allclose(dm_backend.state.fidelity(sv_backend.state.psi), 0.25)
+
+
```
shinich1, line 396: could you add tests for larger simulations (see tests for pattern and sv sim)? compare results with circuit.simulat_statevector at no-noise case, and with noisy case, fix the np.random.seed and check the results with some known result you calculate without using the backend?

tests/test_density_matrix.py
```diff
@@ -0,0 +1,397 @@
...
+        circ.rx(0, np.pi / 2)
+        pattern = circ.transpile()
+
+        backend = DensityMatrixBackend(pattern)
+        backend.add_nodes([0, 1, 2])
+        backend.entangle_nodes((0, 1))
+        backend.entangle_nodes((1, 2))
+        backend.measure(backend.pattern.seq[-4])
+
+        expected_matrix_1 = np.zeros((4, 4))
```
shinich1, line 308: how about np.ones((4, 4))/2. ?

tests/test_density_matrix.py
nabe98: This is checking the concrete case where both resulting states becomes equal(no dephase case(p=0) and p=1 case).

tests/test_density_matrix.py
nabe98: If you mean np.ones((2,2))/2, can be refactored as one-liner. Thanks! (in the next commit I'll push this change)
`expected_matrix_1 = np.kron(np.array([[1, 0], [0, 0]]), np.ones((2, 2)) / 2)`

w.r.t this, `expected_matrix_2` is also refactored as one-liner
`expected_matrix_2 = np.kron(np.array([[0, 0], [0, 1]]), np.array([[0.5, -0.5], [-0.5, 0.5]]))`

tests/test_density_matrix.py
nabe98: I added the test case for 3-qubit QFT. Please see that if it's enough.

shinich1: could you use https://github.com/TeamGraphix/graphix/blob/master/tests/random_circuit.py#L56 instead?

nabe98: After the commit [a6745bb](https://github.com/TeamGraphix/graphix/pull/64/commits/a6745bb6e09a039bab0d8b30d88f7d79294d7165), I'll fix this part.

nabe98: What do you mean by this? I think it's not the test for `dephase` if you mean something like `np.testing.assert_allclose(dm_backend.state.fidelity(random_state), dm_backend.state.fidelity(exact_qft_state))` instead of
`np.testing.assert_allclose(dm_backend.state.fidelity(noisy_state), dm_backend.state.fidelity(exact_qft_state))`

shinich1: I mean something like `np.testing.assert_allclose(1, dm_backend.state.fidelity(known_noisy_state_for_noise_realization))` (fix the noise realizatino by np.random.seed())

thierry-martinez (commented):
I didn't have time to reread all the code yet, so here are only very minor comments. I hope I will have time to read more code in the following days!

.gitignore
```diff
@@ -9,3 +9,7 @@ docs/source/gallery/**
 docs/source/benchmarks/**
 .tox/
 .python-version
+.pre-commit-config.yaml
+graphix/sim/graphix.code-workspace
+
+.vscode/settings.json
```
shinich1, line 15: do we need to specify them?

graphix/channels.py
```diff
@@ -0,0 +1,208 @@
...
+            {"coef": np.sqrt(prob / 15.0), "operator": np.kron(Ops.x, Ops.z)},
+            {"coef": np.sqrt(prob / 15.0), "operator": np.kron(Ops.y, Ops.x)},
+            {"coef": np.sqrt(prob / 15.0), "operator": np.kron(Ops.y, Ops.z)},
+            {"coef": np.sqrt(prob / 15.0), "operator": np.kron(Ops.z, Ops.y)},
+            {"coef": np.sqrt(prob / 15.0), "operator": np.kron(Ops.z, Ops.x)},
+        ]
+    )
+
+
+def two_qubit_depolarising_tensor_channel(prob: float) -> KrausChannel:
```
shinich1, line 172: how about adding two qubit dephasing channel and pauli channel?

mgarnier59: Random Pauli channel added in [5501114](https://github.com/TeamGraphix/graphix/pull/64/commits/5501114f4c9135bb49fee29db37721836a9e71be). 
Isn't the 2-qubit dephasing channel just the tensor product of single-qubt dephasing channels? Indedd, some channel manipulation utilities could be added (tensor, composition).

.gitignore
mgarnier59: Just my local files that I wanted to ignore.

graphix/channels.py
shinich1: great, how does it work? shall we move it to `channels` instead, that seems useful?

mgarnier59: How does what work?

I have no opinion on whether to move the random Pauli channels to `channel` or leave it in `random_objects` to me it made sense to gather all random stuff at the same place. 

I planned to add all Channel manipulation stuff a bit later.

mgarnier59: > > How does what work?
> 
> `rand_Pauli_channel_kraus` function.
You select the number of operators you want in the Channel (rank) from all possible Pauli tensor products that are precomputed, you build a discrete probability distribution with this number of possibilities and then you choose randomly the correct number of operators.


> Let's move things in `random_objects.py` outside of `tests`, perhaps? functions in `random_circuit` is not so generalizable (beyond unit tests) but I feel that things in `random_objects.py` that you added could be quite useful.

Agreed. Random channels can also be useful.

graphix/channels.py
```diff
@@ -0,0 +1,208 @@
...
+        # check that the channel is properly normalized i.e
+        # \sum_K_i^\dagger K_i = Identity
+        assert check_data_normalization(kraus_data)
+
+        self.nqubit = int(np.log2(kraus_data[0]["operator"].shape[0]))
+        self.kraus_ops = kraus_data
+
+        # np.asarray(data, dtype=np.complex128)
+        # number of Kraus operators in the Channel
+        assert check_rank(kraus_data)
```
nabe98, line 70: This is better since, if `python assert check_rank(kraus_data)` fails, the first two lines are meaningless.

```python
        assert check_rank(kraus_data)
        self.nqubit = int(np.log2(kraus_data[0]["operator"].shape[0]))
        self.kraus_ops = kraus_data
```

setup.py
```diff
@@ -37,7 +37,7 @@
     ],
     "python_requires": ">=3.8,<3.12",
     "install_requires": requirements,
-    "extras_require": {"extra": ["graphix-ibmq", "graphix-perceval"]},
+    "extras_require": {'extra': ['graphix-ibmq', "graphix-perceval"], 'test': ['graphix-ibmq','scipy','random']},
```
nabe98, line 40: Please use " instead of ' if there is no reason!

nabe98 (commented):
Hello! I've left some simple comments (@shinich1 asked me to). Please check them.
As soon as I have time, I might leave some more comments.

graphix/channels.py
mgarnier59: Ok. We'll also change all dimensions check for something more efficient soon.

setup.py
mgarnier59: Don't know what you mean. If I remember correctly, random is no longer needed. I'll remove it.

graphix/random_objects.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    tmp = np.random.rand(l, l) + 1j * np.random.rand(l, l)
+    return tmp + tmp.conj().T
+
+
+def rand_unit(l: int):
+    """
+    generate random unitary matrix of size l*l from hermitian matrix
+    """
+    return scipy.linalg.expm(1j * rand_herm(l))
```
masa10-f, line 22: How about using `scipy.stats.unitary_group`?
https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.unitary_group.html

mgarnier59: Used it in the beginning then tried to remove it to avoid using `scipy` but couldn't and used this more simple version. I opted for the easy version since I had no specific requirement on the measure used for the sampling but we might want to make sure that we use the Haar measure.

graphix/random_objects.py
mgarnier59, line 1: Moved this file to `graphix/` as suggested by @shinich1 . This actually implies that `scipy`has to be added to the requirements. @shinich1 wasn't a huge fan of adding requirements, so what do we decide?

graphix/channels.py
```diff
@@ -0,0 +1,209 @@
...
+class KrausChannel:
+    """quantum channel class in the Kraus representation.
+    Defined by Kraus operators $K_i$ with scalar prefactors ('coef') $c_i$,
+    where the channel act on density matrix as
+    .. math::
+        \rho' = \sum K_i^\dagger \rho K_i
+
+    the data should satisfy
+    .. math::
+        \sum K_i^\dagger K_i = I
```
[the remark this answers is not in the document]
nabe98: I think it is appropriate to classify our implementation as trace-preserving.
In my understanding, a scenario becomes non-trace-preserving (NTP) when projection measurements are conducted, although there might exist other NTP cases. In our implementation, measurement operations are not represented by Kraus operators; instead, they are handled using a different method.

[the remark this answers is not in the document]
mgarnier59: CPTP maps represent what happens when  a **unitary** evolution is applied on the larger system system tensor ancilla (or environment) and you trace out the ancilla part. If measurements are introduced, then you recover the theory of POVM which still must sum to 1 (completeness). As far as I can tell, we don't need non-trace preserving maps.

graphix/sim/density_matrix.py
```diff
@@ -0,0 +1,439 @@
...
+
+    def fidelity(self, statevec):
+        """calculate the fidelity against reference statevector.
+
+        Parameters
+        ----------
+            statevec : numpy array
+                statevector (flattened numpy array) to compare with
+        """
+        return np.abs(statevec.transpose().conj() @ self.rho @ statevec)
```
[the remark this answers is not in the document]
shinich1: I agree with Maxine, let us leave it like this for now.

graphix/random_objects.py
shinich1: I missed that point that I asked! 
I think, at this stage, having random_objects toolkit far exceeds the dependency being a bit heavier. shall we add `scipy` then?

graphix/channels.py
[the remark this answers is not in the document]
mgarnier59: I agree, let's postpone this discussion if we realize we need to implement non TP maps (with I don't believe is necessary so far). The photon-loss channel is a CPT map but this requires modelling that is not currently done at the graphix level.

graphix/random_objects.py
mgarnier59: I also think so. Random objects might be interesting for benchmarking algorithms outside of tests.

graphix/channels.py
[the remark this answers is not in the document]
masa10-f: @mgarnier59 @shinich1 @nabe98 Thanks! I agree to address it in another issue

graphix/random_objects.py
masa10-f: I'd like to note that `quimb` , tensor network library, already depends on `scipy`, so there will be no actual dependency change :)

mgarnier59: Great!

graphix/random_objects.py
masa10-f: I've read that Haar-random Unitary can be created using QR decomposition. Original paper is [1] and it is also cited in `scipy.stats.unitary_group` module. So, I believe it would be better to use `unitary_group` module!

[1] Francesco Mezzadri [How to generate random matrices from the classical compact groups](https://arxiv.org/abs/math-ph/0609050)
[2] [Understanding the Haar measure](https://pennylane.ai/qml/demos/tutorial_haar_measure/#mezzadri2006)

mgarnier59: As you want. I'm not sure it matters that much anyways.

shinich1 (commented):
All the code look good! Docs rendering and example(sphinx gallery) fixed.

Once test passes, we can squash and merge to master. I can't seem to approve  because I created this PR. Can you approve this PR now, @masa10-f ?

once approved, could you squash and merge, @mgarnier59 , so that you'll become the committer?

masa10-f (approved):

Issue #45, opened 2023-04-28
Parameterized measurement patterns


shinich1:
The current implementation only allows the creation of measurement patterns with a fixed set of parameters (rotation angles). However, for quantum-classical algorithms such as VQE or QAOA, it is desirable to run many identical quantum circuits (patterns) with differences in the rotation parameters only. For this, it would be ideal to have the option to create parameterized measurement patterns whose rotation angles can be changed after transpilation and optimizations (standardization, signal shifting, Pauli measurement preprocessing, and more).

For this, I propose to:
- write `Parameter` class in a new file `graphix/parameter.py` to instantiate parameter objects (c.f. [qiskit parameterized circuit](https://qiskit.org/documentation/tutorials/circuits_advanced/01_advanced_circuits.html)), which has parameter name and bounds (we assume them to be real values). 
- Modify `graphix.transpiler.Circuit` class to accept parameterized rotation angles
- modify methods of `graphix.pattern.Pattern` and  `graphix.simulator.Simulator` classes to ask for parameter value substitution when running the pattern as keyword argument, if there are parameterized commands.
- modify `graphix.pattern.pauli_nodes()` to consider parameterized measurement commands as non-Pauli measurements
- add methods to `graphix.pattern.Pattern` to inspect parameterized patterns

I may be missing other fixes and additions that need to be done to make this feature work, which I would be keen to discuss on this issues page.

Please see our [contribution guide](https://github.com/TeamGraphix/graphix/blob/master/CONTRIBUTING.md) before starting, and do not forget to comment on this issue to show your interest. We are always happy to answer questions, so do not hesitate to ask on this page.

Thank you for your interest, and good luck!

pafloxy:
Hello! I am quite interested in this project.

pafloxy:
Hey! 
I was thinking about what we were trying to do here. (** Btw the link you have provided in the comment appears to be void ; ) Anyway I understand that we need to have placeholder for a paramter as they have in the qiskit't parameter class (`Parameter`, `ParameterExpression`, etc. [ link ](https://qiskit.org/documentation/_modules/qiskit/circuit/parameterexpression.html#ParameterExpression)). As for it, were you aiming to have a new class alike it here or maybe we can simply use the availiable implementation from qiskit ? In my experience the class is pretty much adaptable to the things we might need here, or so it seems to me. (** I have looked through the details of the `transpiler` and the `simulator` module , and goinf through the `pattern` module at the moment.)

[...]

PR #68, opened 2023-07-06
Completed Parameter Implementation 

pafloxy:
 
Then, please fill in below:

**Context (if applicable):**
 

Introduced `Parameter` class for providing symbolic placeholder for parameterised arguments in circuit (rotation-angles) / patterns (measurement-angles). 

It is allowed to make numerical operation on the instances of the `Parameter` class which upadates the symbolic expression inplace, however interaction between multiple instances of the class are not allowed . 

Numerical value can be assigned using `Parameter.bind()` method 

```
th = Parameter('th');
np.sin(th/3)
------------------
>Out: expr = sin(th[/3])
```  

```
th.bind({th:0.3})
-------------------
>Out: expr = sin(th[/3]), assignment = [ th : 0.3 ]
```

```
th.value
------------------
>Out: 0.09983341664682814
```
**Related issue:**
#45 


shinich1 (approved):

mgarnier59:
Do we validate the behaviour modification of `output_nodes`in `pattern.py`around line 100? I don't yet understand how it works if the output nodes are specified by the user but if it is handled by the pattern, then I fear this modification will break the code since output_nodes is built iteratively.

mgarnier59:
> > Do we validate the behaviour modification of `output_nodes`in `pattern.py`around line 100? I don't yet understand how it works if the output nodes are specified by the user but if it is handled by the pattern, then I fear this modification will break the code since output_nodes is built iteratively.
> 
> Thanks for pointing this out. I agreed to this change in the discussion on the issues page, but I suppose you are right, breaking change like this should be avoided. More so because we don’t need this change for the parametrised patterns to work.

Maybe the error pointed out by @pafloxy could be resolved by calling the `add` method directly in `__init__`.

shinich1:
@pafloxy I accidentally deleted the target branch of this PR, sorry! 
what should we do with this feature? I think many of the changes I asked back when we were working on this are now likely unnecessary (sorry..) and your implementation may fit very nicely in the current version of graphix?

pafloxy:
Hello @shinich1 , 
Its alright, I think it would be more convenient to rewrite it with the current version of graphix which would also be more convenient. Also we were reconsidering how we use the parameterised class for yourselves among oursleves so maybe this time we will write it in a different fashion, the `Parameter` class on overall might not be changed but the way it interacts with rest of graphix should be require some re-thinking based on what we decide :)

graphix/parameter.py
```diff
@@ -38,70 +37,158 @@ def __init__(self, name):
...
+
+        if len(val.free_symbols) == 0:
+
+            if overwrite:
+                self._value = float(val)
+                self._assignment = value
+            else:
+                if not self.assigned:
+                    self._value = float(val)
+                    self._assignment = value
```
benjvmin93, line 95: In my sense, this part is redudant and should be replaced by something more concise. In any case, you are setting `self._value` and `self._assignment`. For me the conditions should only be `if not self.assigned: ... else: raise ValueError("...")`

graphix/parameter.py
```diff
@@ -38,70 +37,158 @@ def __init__(self, name):
...
+                else:
+                    raise ValueError("Symbols are already assigned, set overwrite = True to overwrite value")
+
+        if len(val.free_symbols) != 0:
```
benjvmin93, line 99: This could be replaced by an `else` statement.

graphix/pattern.py
```diff
@@ -88,11 +93,134 @@ def add(self, cmd):
...
+            self.inspect_parameterized_commands()
+
+        else:
+            new_pattern = deepcopy(self)
+            cmd_indx = -1
+            for cmd in new_pattern.seq:
+                cmd_indx += 1
+                if cmd[0] == "M" and isinstance(cmd[3], Parameter):
+                    cmd[3] = cmd[3].bind(parameter_assignment, allow_unknown_parameters=True).value
+                    new_pattern.seq[cmd_indx] = cmd
```
benjvmin93, line 210: Here, there is some code duplication which could be avoided.

graphix/transpiler.py
```diff
@@ -33,6 +36,104 @@ def __init__(self, width):
...
+            new_pattern = deepcopy(self)
+
+            for gate_indx in parameterized_gate_indx:
+
+                gate_instr = new_pattern.instruction[gate_indx]
+                if isinstance(gate_instr[-1], (Parameter)):
+                    gate_instr[-1] = gate_instr[-1].bind(parameter_assignment, allow_unknown_parameters=True).value
+                    new_pattern.instruction[gate_indx] = gate_instr
+
+            new_pattern.inspect_parameterized_gates()
```
benjvmin93, line 119: Code duplication that could be avoided by just assigning `self.instruction[gate_indx]` or `new_pattern.instruction[gate_indx]` depending on the `inplace` condition.

graphix/transpiler.py
```diff
@@ -33,6 +36,104 @@ def __init__(self, width):
...
+        return_data : bool
+                    if True then returns
+                    set - of all the parameters used in the circuit
+
+        """
+
+        if self.num_parameterized_gates == 0:
+            raise ValueError(" no paramaters to assign ")
+
+        elif self.num_parameterized_gates > 0:
```
benjvmin93, line 59: No need to make this part as a condition statement since we know the `num_parameterized_gates` attribute is setted to 0 and we already test it in the if statement above and that it raises an error if so.

Issue #69, opened 2023-07-06
Probability calculation


mgarnier59:
**Is your feature request related to a problem? Please describe.**
So far, all backends rely on the fact that the probability of every measurement outcome is 1/2. That holds so far but it won't be the case for eg noisy simulations and maybe in other cases depending on the future developments.

**Describe the feature you'd like**
Allow the user to specify if he/she wants to explicitly compute the probability. Indeed since it is not completely clear to me when this simplification is valid, maybe it's good to leave the choice to the user. For noisy simulation, I guess there won't be any choice.
This would have to be done for all backends.
This might require some backend updates to avoid useless computations (probability is an expectation value that itself relies on the `evolve_single` method)

**Additional context**
NA

mgarnier59:
Working on it.

shinich1:
@mgarnier59 Shall we close this issue? I believe this is done.

mgarnier59:
Yes indeed!

PR #71, opened 2023-07-07
#69 Implemented calculation of the probability in StateVector backend

mgarnier59:
- Done
- Done
 
Then, please fill in below:

**Context (if applicable):**

**Related issue:**

### DIFF
--- a/graphix/sim/statevec.py
+++ b/graphix/sim/statevec.py
@@ -109,17 +109,19 @@ class StatevectorBackend:
             m_op = meas_op(angle, vop=vop, plane=cmd[2], choice=0)
             result = 0
 
-        m_op = meas_op(angle, vop=vop, plane=cmd[2], choice=0)
-        result = 0
+            # probability to measure in the |+_angle> state.
+            # NOT EFFICIENT AT ALL since expectation_single calls evolve_single...
+            prob_0 = self.state.expectation_single(m_op, loc)
 
-        # probability to measure in the |+_angle> state.
-        # NOT EFFICIENT AT ALL since expectation_single calls evolve_single...
-        prob_0 = self.state.expectation_single(m_op, loc)
+            # choose the measurement result randomly according to the computed probability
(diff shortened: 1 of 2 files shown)


graphix/sim/statevec.py
```diff
@@ -8,9 +8,9 @@
 class StatevectorBackend:
     """MBQC simulator with statevector method."""
 
-    def __init__(self, pattern, max_qubit_num=20):
+    def __init__(self, pattern, max_qubit_num=20, pr_calc=False):
```
shinich1, line 11: Is there a reason to set this at the instantiation of the backend? 
how about adding `pr_calc` as kwarg for `measure()`?

shinich1:
If the intension is to merge this to master independent from the densitymatrix PR, I believe it is best to start a new branch and copy and paste the changes to statevec only and make a commit or two there (I tried to come up with good place to rebase but this branch seems to be convolved with DM branch).
I would personally suggest we do so, to keep issues independent and trackable :)

graphix/sim/statevec.py
shinich1: I probably see the intension now, is it so that we can pass the option directly from `pattern.simulate_pattern` kwarg? If so I think it makes perfect sense!

mgarnier59: Sorry for the late reply. Yes that was the intention :).

mgarnier59:
What is your opinion concerning the `prob_0 = self.state.expectation_single(m_op, loc)` bit?
In the current version, `evolve_single`is called one too many times since it is called by `expectation_single` but since it is only `numpy`tensor operations, it might not be very expensive. 

I've written an improvement that saves this one call if you believe it is better.

shinich1:
@mgarnier59 Thanks for the idea, it's worth a try! Perhaps good to do a quick benchmarking (measure time it takes for 100 measurements with and without the proposed change) before deciding?

mgarnier59:
Ok, I'll try that!

Issue #73, opened 2023-07-11
Remove ptrace mention in the docstring (Partial trace of state vector backend)


masa10-f:
**Update [@thierry-martinez]: The ptrace method has been removed from Statevec in https://github.com/TeamGraphix/graphix/pull/312. Note that there is still a reference to this method in the docstring of remove_qubit so I am leaving this issue open.**

**Describe the bug**

The partial trace of a state vector is generally expected to return a density matrix, but `Statevec.ptrace` currently returns a state vector. This is an incorrect operation for non-separable quantum states, such as a Bell state.

**To Reproduce**

You can check the above behavior with the following code.

```
# %%
import numpy as np
from graphix.sim.statevec import Statevec

# %%
# prepare H and CNOT gates.
H = np.array([[1, 1], [1, -1]]) / np.sqrt(2)
CNOT = np.array([[1, 0, 0, 0], [0, 1, 0, 0],
                 [0, 0, 0, 1], [0, 0, 1, 0]])

# %%
# make a bell state
sv = Statevec(plus_states=False, nqubit=2)
sv.evolve(H, [0])
sv.evolve(CNOT, [0, 1])

print(sv.flatten())
# %%
# trace out 2nd qubit
sv.ptrace([1])
print(sv.flatten())
# the return should be |0><0| + |1><1| but, it returns |0> or |1>
```

**Expected behavior**

The `Statevec.ptrace` should return a density matrix. A code converting a reduced density matrix into a state vector should check its purity before conversion.

**Environment (please complete the following information):**
 - OS: Ubuntu 20.04
 - Python version: 3.8.10
 - Related module versions if applicable: graphix=0.2.0

**Additional context**

N/A

shinich1:
Just to clarify: this indeed is something to be taken care of, but `pattern.simulate_pattern(backend='statevector')` is working perfectly fine as it is, because the measured qubits are separable from the rest.

masa10-f:
I've written a draft improvement plan below.

```
 # %%
import numpy as np
from graphix.sim.statevec import Statevec, meas_op

import time


def truncate_one_qubit(k, sv):
    n = len(sv.dims())
    taken = np.zeros(2**(n - 1), dtype=np.complex128)
    state = sv.flatten()
    for i in range(2**k):
        for j in range(2**(n - k - 1)):
            taken[i * 2**(n - k - 1) + j] = state[2**(n - k) * i + j]

    norm = taken.dot(taken.flatten().conjugate())
[...]

PR #74, opened 2023-07-24
remove_qubit method for faster statevec simulation

nabe98:
 
Then, please fill in below:

**Context (if applicable):**
`ptrace` method was implemented by using density matrix. But for separable states (states that measured), using DM is unnecessary. 

 - Add `truncate_one_qubit` method to `Statevec`. 
 
directly truncate the specified qubit.

**Related issue:**
#73 

### DIFF
--- a/graphix/sim/statevec.py
+++ b/graphix/sim/statevec.py
@@ -259,10 +259,9 @@ class Statevec:
             qarg (int): qubit index
         """
         # extract |***0_{qarg}***> components if not zero else |***1_{qarg}***>
+        assert not np.isclose(norm(self.psi), 0)
         psi = self.psi.take(indices=0, axis=qarg)
-        if np.isclose(norm(psi), 0):
-            psi = self.psi.take(indices=1, axis=qarg)
-        self.psi = psi
+        self.psi = psi if not np.isclose(norm(psi), 0) else self.psi.take(indices=1, axis=qarg)
         self.normalize()
 
     def entangle(self, edge):
--- a/graphix/sim/statevec.py
+++ b/graphix/sim/statevec.py
@@ -278,8 +278,8 @@ class Statevec:
         Note this does not exactly perform the partial trace: see :meth:`~graphix.sim.statevec.Statevec.ptrace`
         and warning therein.
 
-        For a statevector :math:`\ket{\psi} = \sum c_i \ket{i}` with :math:`\ket{i} = \ket{0 \dots 0}, \dots,
-        \ket{1 \dots 1}`, this method returns
+        For a statevector :math:`\ket{\psi} = \sum c_i \ket{i}` with sum taken over :math:`i \in 0 \dots 00,\ 0\dots 01,\ \dots,\
+        1 \dots 11`, this method returns
 
         .. math::
             \begin{align}
@@ -291,7 +291,7 @@ class Statevec:
            \end{align}
 
         after normalization, for :math:`k =` qarg. If the :math:`k` th qubit is in :math:`1` state,
-        above will be zero, so in such a case the returned state will be the one above with :math:`0_{\mathrm{k}}`
+        above will not work, and in such a case the returned state will be the one above with :math:`0_{\mathrm{k}}`
         replaced with :math:`1_{\mathrm{k}}` .
 
         .. warning::
--- a/graphix/sim/statevec.py
+++ b/graphix/sim/statevec.py
@@ -297,9 +297,9 @@ class Statevec:
             qubit index
         """
         # extract |***0_{qarg}***> components if not zero else |***1_{qarg}***>
-        assert not np.isclose(get_norm(self.psi), 0)
+        assert not np.isclose(_get_statevec_norm(self.psi), 0)
         psi = self.psi.take(indices=0, axis=qarg)
-        self.psi = psi if not np.isclose(get_norm(psi), 0) else self.psi.take(indices=1, axis=qarg)
+        self.psi = psi if not np.isclose(_get_statevec_norm(psi), 0) else self.psi.take(indices=1, axis=qarg)
         self.normalize()
 
     def entangle(self, edge):
@@ -357,7 +357,7 @@ class Statevec:
 
     def normalize(self):
         """normalize the state"""
-        norm = get_norm(self.psi)
+        norm = _get_statevec_norm(self.psi)
         self.psi = self.psi / norm
 
     def flatten(self):
@@ -405,6 +405,6 @@ class Statevec:
         return np.dot(st2.psi.flatten().conjugate(), st1.psi.flatten())
 
 
-def get_norm(psi):
+def _get_statevec_norm(psi):
     """returns norm of the state"""
     return np.sqrt(np.sum(psi.flatten().conj() * psi.flatten()))
--- a/graphix/sim/statevec.py
+++ b/graphix/sim/statevec.py
@@ -260,7 +260,9 @@ class Statevec:
         """
         # extract |***0_{qarg}***> components if not zero else |***1_{qarg}***>
         psi = self.psi.take(indices=0, axis=qarg)
-        self.psi = psi if not np.isclose(psi.flat[0], 0) else self.psi.take(indices=1, axis=qarg)
+        if np.isclose(norm(psi), 0):
+            psi = self.psi.take(indices=1, axis=qarg)
+        self.psi = psi
         self.normalize()
 
     def entangle(self, edge):
--- a/graphix/sim/statevec.py
+++ b/graphix/sim/statevec.py
@@ -252,13 +252,13 @@ class Statevec:
         return self.psi.shape
 
     def ptrace(self, qargs):
-        """partial trace selected qubits
+        """Perform partial trace of the selected qubits.
 
         .. warning::
             This method currently assumes qubits in qargs to be separable from the rest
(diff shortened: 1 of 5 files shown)


masa10-f (commented):
Could you add more test cases?
At least include the following measurement commands
1. measure into |0>
2. measure into |1>
3. non-Pauli measurement

masa10-f:
Could you change the `measure` method in the `StatevectorBackend` to trace out measured qubits individually?

masa10-f:
Thanks for your commitment!
Your code looks great, but I would like to mention one minor point: the state vector backend should not depend on the pattern or circuit simulator. Therefore, it's preferable to avoid them in the unit tests. So, I have removed the random circuit check.

masa10-f (approved):

masa10-f:
@nabe98  Before merging, can you write changes to CHANGELOG.md?

graphix/sim/statevec.py
```diff
@@ -1,8 +1,9 @@
+from copy import deepcopy
+
```
shinich1, line 2: please remove extra line here.

graphix/sim/statevec.py
```diff
@@ -254,8 +252,14 @@ def dims(self):
     def ptrace(self, qargs):
         """partial trace
 
-        Args:
-            qargs (list of ints): qubit inices to trace over
+        CAUTION: This method is currently valid only for separable states.
```
shinich1, line 255: let me fix up this text later, reads a bit strange?

graphix/sim/statevec.py
```diff
@@ -335,15 +367,24 @@ def expectation_single(self, op, loc):
...
+        complex: expectation value
         """
         st1 = deepcopy(self)
         st1.normalize()
         st2 = deepcopy(st1)
         st1.evolve(op, qargs)
         return np.dot(st2.psi.flatten().conjugate(), st1.psi.flatten())
+
+
+def get_norm(psi):
```
shinich1, line 388: why not add as a method (eg `_get_norm` if you don't want to explicitly expose it

graphix/sim/statevec.py
```diff
@@ -264,11 +268,27 @@ def ptrace(self, qargs):
         evals, evecs = np.linalg.eig(rho)  # back to statevector
         self.psi = np.reshape(evecs[:, np.argmax(evals)], (2,) * nqubit_after)
 
+    def truncate_one_qubit(self, qarg):
```
shinich1, line 271: is `truncate` a good word here? let me have a think for a bit

shinich1: truncate sounds like removing the last qubit?

tests/test_statevec_backend.py
```diff
@@ -0,0 +1,52 @@
+import time
+import unittest
+from copy import deepcopy
+
+import numpy as np
+from graphix.ops import States
+from graphix.sim.statevec import Statevec, meas_op
+
+
+class TestStatevec(unittest.TestCase):
```
shinich1, line 10: is this supposed to be complete testing of the backend, or is this only for `truncate` method?

shinich1:
@nabe98 please let me know once you updated CHANGELOG.md. Please do not merge yet. 
Also remember we always `prefer squash and merge` to keep the commit history concise on master branch.

graphix/sim/statevec.py
masa10-f: I followed `isort` package here. Why don't we obey PEP8?

graphix/sim/statevec.py
```diff
@@ -205,15 +194,18 @@ def meas_op(angle, vop=0, plane="XY", choice=0):
 class Statevec:
     """Simple statevector simulator"""
 
-    def __init__(self, plus_states=True, nqubit=1):
+    def __init__(self, nqubit=1, plus_states=True):
```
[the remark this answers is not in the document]
masa10-f: Because we can generate Statevec object simply with `Statevec(n)`. I think this is a standard format.

graphix/sim/statevec.py
masa10-f: Hmm... `discard` or `remove` ?

graphix/sim/statevec.py
shinich1: sure if there's good reason!

tests/test_statevec_backend.py
masa10-f: supposed to be complete testing of the statevector backend. In the future, when we add another method into statevec, we will make another tests on this file.

graphix/sim/statevec.py
```diff
@@ -264,11 +269,41 @@ def ptrace(self, qargs):
         evals, evecs = np.linalg.eig(rho)  # back to statevector
         self.psi = np.reshape(evecs[:, np.argmax(evals)], (2,) * nqubit_after)
 
+    def remove_qubit(self, qarg):
+        """ Remove a separable qubit from the system and assemble a statevector for remaining qubits.
+        Note this does not exactly perform the partial trace: see :meth:`~graphix.sim.statevec.Statevec.ptrace` and warning therein.
+
+        For a statevector $\ket{\psi} = \sum c_i \ket{i}$ with $i = 00\dots 0 \dots 11\dots 1$, this method returns
```
shinich1, line 276: @nabe98 does this sound correct? could you check? 
also could you make sure that docs for this function generates as intended (you'll need to `make html` in `docs/` and check for simulator API doc). I have not checked on my side, I would appreciate if you could debug and check that docs gen work!

graphix/sim/statevec.py
nabe98: `_get_statevec_norm` sounds preferable to me!

Wildcard import is not recommended (https://peps.python.org/pep-0008/#imports) so we should avoid that.

graphix/sim/statevec.py
nabe98: @shinich1 
I have updated the docs slightliy and see that the intended HTML is generated correctly.

nabe98:
@shinich1 also updated `CHANGELOG.md`. Pleas check them!

graphix/sim/statevec.py
shinich1: thank you!

masa10-f:
I've conducted benchmarks for the new method.

[Environment]
CPU: intel i7-12700K
RAM: 64GB (DDR4-3200)
OS: Ubuntu 20.04 (WSL)

The following graph shows a performance comparison between `reduce_qubit` and `ptrace` when simulating n (width) x n (depth) random square circuits.
svbackend_comparison_square_circuit

I also benchmarked shallow (depth=1) random circuits, only changing the number of qubits.
svbackend_comparison_shallow

masa10-f:
I used the following random circuit generator

```
def random_circuit(nqubit, depth, seed=None):
    r"""Generate a random circuit.

    This function generates a random circuit with nqubit qubits and depth layers.
    The circuit is generated by CNOT gates and RZ gates.

    Parameters
    ----------
    nqubit : int
        number of qubits
    depth : int
        number of layers
    seed : int
        random seed

    Returns
    -------
    circuit : graphix.transpiler.Circuit object
        generated circuit
    """
    if seed is not None:
        random.seed(seed)
    qubit_index = [i for i in range(nqubit)]
    circuit = Circuit(nqubit)
    for _ in range(depth):
        random.shuffle(qubit_index)
        for j in range(len(qubit_index) // 2):
            circuit.cnot(qubit_index[2 * j], qubit_index[2 * j + 1])
        for j in range(len(qubit_index)):
            circuit.rz(qubit_index[j], 2 * np.pi * random.random())
    return circuit
```

masa10-f:
I pasted the benchmark code below. I didn't perform Pauli measurements because `Pattern.minimize_space()` for a Pauli-measured pattern sometimes doesn't work well.

```
# %%
from graphix.generator import random_circuit
import matplotlib.pyplot as plt

from time import perf_counter

# %%
test_cases = [(i, 1) for i in range(2, 30, 2)]

score_time = []
circuit_time = []

# %%
for case in test_cases:
    nqubit, depth = case
    circuit = random_circuit(nqubit, depth)
    pattern = circuit.transpile()
    pattern.standardize()
    pattern.minimize_space()
    print(f"max space for nqubit={nqubit} circuit is ", pattern.max_space())
    start = perf_counter()
    pattern.simulate_pattern()
    end = perf_counter()
    print(f"nqubit: {nqubit}, depth: {depth}, time: {end - start}")
    score_time.append(end - start)
    start = perf_counter()
    circuit.simulate_statevector()
    end = perf_counter()
    circuit_time.append(end - start)

# %%
# plot the score_time
plt.scatter([case[0] for case in test_cases], score_time, label="score_time")
plt.scatter([case[0] for case in test_cases],
            circuit_time, label="circuit_time")
plt.xlabel("nqubit")
plt.ylabel("time (s)")
# plt.xscale("log")
plt.yscale("log")
plt.title("Time to simulate square random circuits")
plt.legend()
plt.show()

# %%
# write to file
with open("score_time.txt", "w") as f:
    for item in score_time:
        f.write(f"{item}\n")

# %%
# write circuit_time to file
with open("circuit_time.txt", "w") as f:
    for item in circuit_time:
        f.write(f"{item}\n")
```

shinich1 (approved):

shinich1:
@nabe98 can you squash and merge?

PR #76, opened 2023-08-11
DensityMatrix backend fixes

mgarnier59:
Passing

Done

Then, please fill in below:

**Context (if applicable):**
Density matrix backend.

- Modified the normalization of density matrices from Frobenius (2-norm) to trace norm
- Added checks in the initialization of DensityMatrix to check for unit trace and hermiticity of the input data
- Modified the tests accordingly (normalization, hermiticity). For evolution tests, used unitary operators from `scipy.stats.unitary_group`. Can be modified in the future.

**Related issue:**
None.

### DIFF
--- a/graphix/sim/density_matrix.py
+++ b/graphix/sim/density_matrix.py
@@ -4,9 +4,12 @@ Simulate MBQC with density matrix representation.
 """
 
 from copy import deepcopy
+
 import numpy as np
+
+import graphix.checks as checks
 from graphix.ops import Ops
-from graphix.sim.statevec import meas_op, CNOT_TENSOR, SWAP_TENSOR, CZ_TENSOR
+from graphix.sim.statevec import CNOT_TENSOR, CZ_TENSOR, SWAP_TENSOR, meas_op
 
 
 class DensityMatrix:
@@ -38,11 +41,14 @@ class DensityMatrix:
                 pass
             else:
                 raise TypeError("data must be DensityMatrix, list, tuple, or np.ndarray.")
-            nqubit = np.log2(len(data.flatten())) / 2
-            assert np.isclose(nqubit, int(nqubit))
-            nqubit = int(nqubit)
-            self.Nqubit = nqubit
-            self.rho = np.asarray(data, dtype=complex).reshape((2**nqubit, 2**nqubit))
+
+            assert checks.check_square(data)
+            self.Nqubit = int(np.log2(len(data)))
+
+            self.rho = data
+
+        assert checks.check_hermitian(self.rho)
+        assert checks.check_unit_trace(self.rho)
 
     def __repr__(self):
         return f"DensityMatrix, data={self.rho}, shape={self.dims()}"
@@ -64,6 +70,55 @@ class DensityMatrix:
         rho_tensor = self.rho.reshape((2,) * self.Nqubit * 2)
         rho_tensor = np.tensordot(np.tensordot(op, rho_tensor, axes=[1, i]), op.conj().T, axes=[i + self.Nqubit, 0])
         rho_tensor = np.moveaxis(rho_tensor, (0, -1), (i, i + self.Nqubit))
+        self.rho = rho_tensor.reshape((2**self.Nqubit, 2**self.Nqubit))
+
+    def evolve(self, op, qargs):
+        """Multi-qubit operation
+
+        Args:
+            op (np.array): 2^n*2^n matrix
+            qargs (list of ints): target qubits' indexes
+        """
+
+        d = op.shape
+        # check it is a matrix.
+        if len(d) == 2:
+            # check it is square
+            if d[0] == d[1]:
+                pass
+            else:
+                raise ValueError(f"The provided operator has shape {op.shape} and is not a square matrix.")
+        else:
+            raise ValueError(f"The provided data has incorrect shape {op.shape}.")
+
+        op_dim = np.log2(len(op))
+        if not np.isclose(op_dim, int(op_dim)):
+            raise ValueError("Incorrect operator dimension: not consistent with qubits.")
+        op_dim = int(op_dim)
+
+        if op_dim != len(qargs):
+            raise ValueError("The dimension of the operator doesn't match the number of targets.")
+
+        for i in qargs:
+            if i < 0 or i >= self.Nqubit:
+                raise ValueError("Incorrect target indices.")
+        if len(set(qargs)) != op_dim:
+            raise ValueError("A repeated target qubit index is not possible.")
+
+        op_tensor = op.reshape((2,) * 2 * op_dim)
+
+        rho_tensor = self.rho.reshape((2,) * self.Nqubit * 2)
+
+        rho_tensor = np.tensordot(
+            np.tensordot(op_tensor, rho_tensor, axes=[tuple(op_dim + i for i in range(len(qargs))), tuple(qargs)]),
+            op.conj().T.reshape((2,) * 2 * op_dim),
+            axes=[tuple(i + self.Nqubit for i in qargs), tuple(i for i in range(len(qargs)))],
+        )
+        rho_tensor = np.moveaxis(
+            rho_tensor,
+            [i for i in range(len(qargs))] + [-i for i in range(1, len(qargs) + 1)],
+            [i for i in qargs] + [i + self.Nqubit for i in reversed(list(qargs))],
+        )
         self.rho = rho_tensor.reshape((2**self.Nqubit, 2**self.Nqubit))
 
     def dims(self):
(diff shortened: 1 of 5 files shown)


graphix/sim/density_matrix.py
```diff
@@ -42,7 +43,14 @@ def __init__(self, data=None, plus_state=True, nqubit=1):
...
             nqubit = int(nqubit)
             self.Nqubit = nqubit
-            self.rho = np.asarray(data, dtype=complex).reshape((2**nqubit, 2**nqubit))
+
+            # conversion done above
+            self.rho = data.reshape((2**nqubit, 2**nqubit))
+            np.testing.assert_almost_equal(self.rho.trace(), 1, err_msg="The provided matrix does not have unit trace!")
+            # also add test for hermiticity!!! QuTip test? and modify the DM tests that use not hermitian DMs
+            np.testing.assert_equal(
+                scipy.linalg.ishermitian(self.rho), True, err_msg="The provided matrix is not Hermitian!"
```
shinich1, line 52: how about using numpy, for example,
`np.testing.assert_allclose(self.rho, self.rho.getH())`?
scipy is not yet on requirements.txt.

shinich1: I see scipy is uesd for testing too; where necessary, we can install scipy specifically for testing, by adding to https://github.com/TeamGraphix/graphix/blob/master/.github/workflows/ci.yml

I've tried so far to keep requirements.txt as minimum as possible, for faster installation and less chance of version/compatibility issues in the future.

mgarnier59: > how about using numpy, for example, `np.testing.assert_allclose(self.rho, self.rho.getH())`? scipy is not yet on requirements.txt.

I didn't bother thinking about it first since I believed `scipy`was more efficient but I guess 'numpy' is a good way to go. 
Something that might be interesting for performance upgrade: Qutip mentions that it is possible to do a hermiticity check without compuing the conjugate transpose https://qutip.org/features.html. Might be worth having a look.



> I see scipy is uesd for testing too; where necessary, we can install scipy specifically for testing, by adding to https://github.com/TeamGraphix/graphix/blob/master/.github/workflows/ci.yml
> 
> I've tried so far to keep requirements.txt as minimum as possible, for faster installation and less chance of version/compatibility issues in the future.

Of course. It is just used to generate unitaries but it can be easily replace. Indeed from any random matrix building $U U^\dagger$ will yield a random unitary.

shinich1: The random unitary gen is for testing right? Perhaps the best way might be to require scipy for testing (as above), by adding to ci.yml - what do you think?

Of course, when it becomes necessary, I am always up to the idea of adding to requirements.txt too :) just trying to resist at first to keep the dependency as simple as possible.

mgarnier59: > The random unitary gen is for testing right? Perhaps the best way might be to require scipy for testing (as above), by adding to ci.yml - what do you think?

Yes, only for testing! I've added it to ci.yml, I hope I did it correctly. But does `pip install -e .[test]` install the additional dependencies specific to the tests?

shinich1: I was imagining adding `pip install scipy` to ci.yml but you had a good point here, it's actually here where we should add scipy:
https://github.com/TeamGraphix/graphix/blob/25e59bc5dfa0b3042348233d7b02626c84d2bf92/setup.py#L39C1-L39C77

mgarnier59: You sure it's better in the setup.py and not the ci? I guess tests are ran at install. 
Then what does `pip install -e .[test]`do if not installing the dependencies in the tests?

mgarnier59:
New commit solving the issues we mentionned as well as a few updates:

- refactored 2-qubit gates application with `evolve` method (actually works for arbitrary number of qubits, tested for 1-, 2- and 3- qubits)
- rewrote the tensor contraction in a clearer way to avoid confusion (as I experienced) but kept the previous implementation as comment if need be.
- removed `scipy` from `density_matrix.py`
- added data validation with Exceptions instead of `assert`s
- added `scipy` and `random` to `setup.py`
- added tests according to new features

For the feature, maybe generic data validation like checking operators are square matrices of size $2^n \times 2^N$ can be mutualised with the density matrix that just adds further conditions on top of that (hermiticity, unit trace). Maybe put those methods in a specific file?

shinich1:
Thanks! Will take a look through later. 

> Maybe put those methods in a specific file?
good idea, sounds like a portable feature outside this simulator. maybe a new file in in `graphix/sim/`?

mgarnier59:
> good idea, sounds like a portable feature outside this simulator. maybe a new file in in `graphix/sim/`?

Actually I was thinking of it while working on the `kraus.py`file that requires the same types of checks. So either we move `kraus` (we might rename it) in `graphix/sim/` or we include the checks in `graphix/` but that doesn't really matter much.

shinich1:
@mgarnier59 maybe graphix/sim/quantum_channel.py, but as you say it does not matter much at this stage. 
let's continue with whatever that works for you, we then deal with file arrangement after all the implementation part is ready :)

graphix/sim/density_matrix.py
```diff
@@ -66,6 +72,73 @@ def evolve_single(self, op, i):
...
+        #     np.tensordot(op_tensor, rho_tensor, axes=[tuple(op_dim + i for i in range(len(qargs))), tuple(qargs)]),
+        #     op_tensor.conj().T,
+        #     axes=[tuple(i + self.Nqubit for i in reversed(qargs)), tuple(i for i in range(len(qargs)))],
+        # )
+        # rho_tensor = np.moveaxis(
+        #     rho_tensor,
+        #     [i for i in range(len(list(qargs)))] + [-i for i in range(1, len(list(qargs)) + 1)],
+        #     [i for i in qargs] + [i + self.Nqubit for i in list(qargs)],
+        # )
+        # self.rho = rho_tensor.reshape((2**self.Nqubit, 2**self.Nqubit))
```
shinich1, line 140: let's get rid of this!

graphix/sim/density_matrix.py
```diff
@@ -91,20 +164,24 @@ def cnot(self, edge):
...
+        # assert i != j
+
+        # rho_tensor = self.rho.reshape((2,) * n * 2)
+        # rho_tensor = np.tensordot(
+        #     np.tensordot(CNOT_TENSOR, rho_tensor, axes=[(2, 3), edge]),
+        #     CNOT_TENSOR.conj().T,
+        #     axes=[(edge[1] + n, edge[0] + n), [0, 1]],
+        # )
+        # rho_tensor = np.moveaxis(rho_tensor, (0, 1, -1, -2), (edge[0], edge[1], edge[0] + n, edge[1] + n))
+        # self.rho = rho_tensor.reshape((2**n, 2**n))
```
shinich1, line 182: this too

graphix/sim/density_matrix.py
```diff
@@ -218,6 +272,7 @@ def __init__(self, pattern, max_qubit_num=12):
         if pattern.max_space() > max_qubit_num:
             raise ValueError("Pattern.max_space is larger than max_qubit_num. Increase max_qubit_num and try again.")
 
+    # TODO to be removed
     def dephase(self, p=0):
         """Apply dephasing channel to all nodes. Phase is flipped with probability p.
```
shinich1, line 278: let's remove now?

tests/test_density_matrix.py
```diff
@@ -455,6 +628,44 @@ def qft_circ():
...
+        np.testing.assert_allclose(tmp @ tmp.conj().T @ psi, psi)
+        np.testing.assert_allclose(tmp.conj().T @ tmp @ psi, psi)
+
+        # direct assert equal identity doesn't seem to work. Precision issues?
+        # np.testing.assert_allclose(tmp @ tmp.conj().T, np.eye(d))
+        # np.testing.assert_allclose(tmp.conj().T @ tmp, np.eye(d))
+
+
 if __name__ == "__main__":
     np.random.seed(2)
```
shinich1, line 670: perhaps better to remove seeding?

.gitignore
```diff
@@ -6,3 +6,4 @@
 **/build
 dist/*
 docs/source/gallery/**
+.pre-commit-config.yaml
```
shinich1, line 9: let's get rid of this?

shinich1 (approved):
There are a few parts that needs tidying up, but looks good otherwise!
Let's get rid of pre-commit-config from gitignore too. 
After that, please squash and merge.

graphix/sim/density_matrix.py
mgarnier59: Done.

graphix/sim/density_matrix.py
mgarnier59: Done.

graphix/sim/density_matrix.py
mgarnier59: If that's ok with you, I plan to remove it in the next contribution that incorporates the noise handling

tests/test_density_matrix.py
mgarnier59: I think it was there before I jumped in. Maybe it's better for reproducibility?

.gitignore
mgarnier59: Done.

shinich1:
@mgarnier59 I just assigned you to the PR (sorry I haven't done this yet). Do you see the button for squash and merge now?

Issue #47, opened 2023-04-28
Resource state visualization


shinich1:
We need a cool visualization of the graph state from the measurement pattern. It would be ideal to be able to read off the following information from the plot:

1. Underlying graph of the graph state
2. whether qubits are measured in Pauli or non-Pauli bases
3. measurement order where applicable
4. (optional) local Clifford decorations (M[6] of measurement commands M - see [here](https://graphix.readthedocs.io/en/latest/lc-mbqc.html#mbqc-on-local-clifford-decorated-graph-lc-mbqc)) 
5. (optional) flow or gflow structure of the graph

Please implement this as a new method `graphix.pattern.Pattern.view_graph()`. Feel free to move the implementation elsewhere (separate file) if code is too long. Do try to tune the colors/etc for a better look for a range of graph size/shapes!

see [QAOA example](https://graphix.readthedocs.io/en/latest/gallery/qaoa.html#sphx-glr-gallery-qaoa-py) for a flow-based arrangement of graph nodes which gives a linear arrangement of qubits following the measurement order.

pafloxy:
Hi! I was interested about this project though I still have some questions about it. As much as I understand currently we are still using networkx Graph objects for the visualisation part ? However, I was wondering whether it is possible to have a new class of it s own for this purpose, I have seen a similar implementation for the graph class of pyzx. Do you think it would have any advantages this way ?

shinich1:
@pafloxy Yes I think that would work nicely! let's create a new python file, e.g. `visualization.py`, to place the new class.

PR #78, opened 2023-08-21
 Resource state visualization

d1ssk:
 
Then, please fill in below:

**Context (if applicable):**

Implement a `GraphVisualizer` class within visualization.py and add a `view_graph()` method to the `Pattern` class, which invokes methods from the `GraphVisualizer` class.
- The graph states are visualized with flow structure (or gflow structure, to be implemented). 
- Input and output nodes are distinguished by edge color and the color of the node faces respectively.
- (Optional) Nodes measured in the Pauli bases are colored blue.
- (Optional) Indexes of the local Clifford operator are displayed adjacent to the nodes.

**Related issue:**
#47 


pafloxy (commented):
Thanks for the PR. I have tried out the plotting subroutines you have introduced and they work really smooth and the plots are visually appleasing.

I had one suggestion about it though, in `visualize_w_flow` for building the intial layout perhaps it might be helpful to check the `nx.multipartite_layout()` function which can generate a layered structure as well given the layer information is passed on to the graph as a `node_attribrute`. 

These method would in particular be helpful for the case of plotting resource states with gflow as the method used here sets the y-axis coordinates based on the sequence of flow edges connecting an input node to its corresponding output node, which might not be possible for the case where the gflow maps a node to multiplle nodes. 

Let me know what you think :)

graphix/vizualization.py
```diff
@@ -0,0 +1,221 @@
...
+            Filename of the saved plot.
+        """
+        if figsize is None:
+            width = (max(l_k.values()) + 1) * 0.8
+            height = len(self.v_in)
+            figsize = (width, height)
+        plt.figure(figsize=figsize)
+        pos = nx.spring_layout(self.g)  # Initial layout for the nodes
+
+        n = len(self.v_in)
```
pafloxy: The part I commented about in my [review](https://github.com/TeamGraphix/graphix/pull/78#pullrequestreview-1588217396)

d1ssk:
I was just wondering how to set the position of nodes in the case of gflow and your suggestion looks great!
I will use `nx.multipartite_layout()` for visualization of both flow and gflow.

pafloxy:
Sure! I'll be happy to discuss if you want.  

However in my experience, time to time the `nx.multipartite_layout` does behave weirdly in placing the graph nodes which I think we will need to take care of it somehow in the post-processing. 

For eg, see 
image
the nodes 7, 28, 21 are unnecessarily pushed out of line. 

###### The figure is from a code I wrote for the visualisations but ofcourse it isn't even closely sophisticated compared to your implementation :sweat_smile:

d1ssk:
Thank you for the valuable information.  I would like to reproduce this example to consider a countermeasure, could you give me the code that generates this graph?

pafloxy:
Sure, you will find a fucntion called `pretty_plot_pattern()` defined in this [notebook](https://github.com/pafloxy/graphix_pafloxy/blob/47-resource-state-visualization/scratchnb_0.ipynb) It has the required implementation. Though it isn't really a good peice of programming.

graphix/pattern.py
```diff
@@ -888,6 +889,36 @@ def get_meas_plane(self):
...
+        local_clifford : dict
+            local clifford index of the each node.
+        """
+        local_clifford = {}
+        for cmd in self.seq:
+            if cmd[0] == "C":
+                local_clifford[cmd[1]] = cmd[2]
+            elif cmd[0] == "M" and len(cmd) == 7:
+                local_clifford[cmd[1]] = cmd[6]
+        return local_clifford
```
shinich1, line 920: isn't this same as get_vops?

graphix/pattern.py
```diff
@@ -1212,6 +1243,40 @@ def perform_pauli_measurements(self, leave_input=False):
...
+        if pauli_indicator:
+            angles = self.get_angles()
+        else:
+            angles = None
+        if local_clifford_indicator:
+            local_clifford = self.get_local_clifford()
+        else:
+            local_clifford = None
+        vis.visualize(figsize=figsize, angles=angles, local_clifford=local_clifford, save=save, filename=filename)
+
```
shinich1, line 1279: this should be named `visualize_flow` instead, perhaps?

shinich1 (approved):
once you worked on the comments above and merged in updates of master branch, please squash and merge.

graphix/pattern.py
d1ssk: This method also visualize gflow graph or no flow/gflow graph, so a name with a broader meaning is preferred.
I set the name `draw_graph()`

Issue #80, opened 2023-08-23
[Bug]: Gflow finding algorithm


masa10-f:
**Describe the bug**

As per the proposed (expanded) gflow finding algorithm outlined in [M. Backens et al.](https://arxiv.org/abs/2003.01664), there is an issue with the current implementation within the `graphix.gflow.gflow` module.

**To Reproduce**

For instance, consider the underlying graph provided below. While this graph possesses a gflow, the current method fails to identify it.

```
Nodes = {0, 1, 2, 3, 4, 5}
Edegs = {(0, 1), (0, 2), (0, 4), (1, 2), (1, 4), (2, 4), (4, 3), (4, 5)}
Measurement planes = {0: "XY", 1: "XZ", 3: "XY", 4: "XY}
Input = {0, 3}
Output = {2, 5}
```

The root cause of this issue is that the current implementation's handling of conditional branches in measurement planes does not align with the gflow definition.

**Expected behavior**

One of the solutions is like,
```
0: {1},
1: {1, 2},
3: {1, 4},
4: {5}
```

**Environment (please complete the following information):**
 - OS: Ubuntu 20.04
 - Python version: 3.8.10
 - Related module versions if applicable: graphix==0.2.0

**Additional context**

Currently, graphix depends on z3-solver for gflow identification. However, this dependency could be substituted with a solver tailored for boolean system of equations. Such a solver can be efficiently implemented using only numpy.

pafloxy:
Hello @masa10-f ,

>>The root cause of this issue is that the current implementation's handling of conditional branches in measurement planes does not align with the gflow definition.

Could you maybe elaborate a bit more on this, specifically which part of the implementation you are reffering to ?

>>Currently, graphix depends on z3-solver for gflow identification. However, this dependency could be substituted with a solver tailored for boolean system of equations. Such a solver can be efficiently implemented using only numpy.

Actually I did look for efficient means of solving linear equations on GF2, but the methods I found were rather sophisticated moreover I was not convinced if they could find all possible solutions to the set of equations either. (for the current implementatio `solbebool` finds just one). I would be interested in knowing how you plan to approach this problem. :)

masa10-f:
Hi @pafloxy ,

> Could you maybe elaborate a bit more on this, specifically which part of the implementation you are reffering to ?

I  referred to the following part(graphix.gflow l.192-203).

```
for u in iter(v_rem):
    if meas_plane[u] == "Z":
        c_set = c_set | {u}
[...]

PR #81, opened 2023-08-29
Gflow finding algorithm

masa10-f:
 
Then, please fill in below:

**Context (if applicable):**

- Fix a conditional branch in gflow finding algorithm (See #80 for details)
- Add a GF(2) solver tailored for gflow finding algorithm
- Add an option whether to write down all possible gflow or not
- Require meas_plane argument every time to find flow or gflow 
- Remove z3-solver from dependency

**Related issue:**

#80 


shinich1:
thanks, I will take a look at the code!
- let's write in changelog.md, specifically highlighting the breaking change of interface (meas_plane is now required).
- please paste the benchmark plot here, so people can see the benefit of new code.
- perhaps @pafloxy can also take a look and provide comments where necessary?

graphix/generator.py
```diff
@@ -9,7 +9,7 @@
 from graphix.gflow import flow, gflow, get_layers, find_odd_neighbor
 
 
-def generate_from_graph(graph, angles, inputs, outputs, timeout=100):
+def generate_from_graph(graph, angles, inputs, outputs, meas_planes):
```
shinich1, line 12: I'm still wondering whether it is good to require meas_planes for things like generate_from_graph. Most literatures on MBQC (say papers and reviews by Raussendorf) deal only with XY-plane measurement and most people may not go further than that (gflow, local-Clifford decoration, etc). I guess for those using only XY-plane it's just one-liner to create the meas_plane dict, so perhaps not very cumbersome. BTW, do we need meas_plane argument for `flow`? I see it's to raise error but I feel it's a bit strange to add it just to throw errors?
what's your thoughts @pafloxy ?

pafloxy: >BTW, do we need meas_plane argument for flow? I see it's to raise error but I feel it's a bit strange to add it just to throw errors?
 
The causal flow algorithm has not been adapted to handle different measurement planes, and most probably it is not possible either given measuring in some planes would require more than one element in their correction set. So it seems we may not really need it, but maybe @masato-fuk had something on his mind ?

  >Most literatures on MBQC (say papers and reviews by Raussendorf) deal only with XY-plane measurement and most people may not go further than that (gflow, local-Clifford decoration, etc).

I agree, actually measurement patterns with other a generalised set of planes is something that has not really been considered widely before. However note that  [1] shows, even if we start with multiple measurement patterns we can alway use certain graph operations such that we are left with only measurement planes in {XY, YZ}, along side restrictions to the geometry of nodes with YZ measurements. 
Nonetheless I think it would be by no means superflous to allow all three measuremnt planes in the patterns, perhaps it might have some application we have not seen yet. 

Sorry I was bit late yesterday, I still have to check rest of the commit. But on the outset it seems pretty neat and precise. Nice work @masa10-f  ;)

masa10-f: Thank you for your review!

> I'm still wondering whether it is good to require meas_planes for things like generate_from_graph. Most literatures on MBQC (say papers and reviews by Raussendorf) deal only with XY-plane measurement and most people may not go further than that (gflow, local-Clifford decoration, etc). I guess for those using only XY-plane it's just one-liner to create the meas_plane dict, so perhaps not very cumbersome. 

I agree with your opinion. I will eliminate the need for the `meas_planes` argument of some functions that use flow/gflow inside.

> BTW, do we need meas_plane argument for flow? I see it's to raise error but I feel it's a bit strange to add it just to throw errors?

In my opinion, yes. I think causal flow is a property of an underlying graph (graph with measurement planes), not a graph itself. Requiring the `meas_planes` is not meant to raise an error. An underlying graph with either XZ or YZ measurement planes actually does not have any flow from the definition. I mean that 'no flow' is also a correct answer.

How do you think about it? @pafloxy

graphix/gflow.py
```diff
@@ -461,3 +392,382 @@ def get_layers(l_k):
     for i in l_k.keys():
         layers[l_k[i]].append(i)
     return d, layers
+
+
+class GF2Solver:
+    """Solver for GF(2) linear equations"""
+
+    def __init__(self, graph, input, output):
```
pafloxy, line 391: The solver implementation looks great, I was just wondering since the `GF2Solver()` is meant ot solve non-square systems in $GF(2)^{\otimes(n)}$, should we always restrict the input parameter to be an instance of `nx.Graph()` [^1] ?

[^1]: I see you have not typecasted the input, but I guess you still are expecting a graph.

graphix/generator.py
masa10-f: Thank you for your review! @shinich1 

> I'm still wondering whether it is good to require meas_planes for things like generate_from_graph. Most literatures on MBQC (say papers and reviews by Raussendorf) deal only with XY-plane measurement and most people may not go further than that (gflow, local-Clifford decoration, etc). I guess for those using only XY-plane it's just one-liner to create the meas_plane dict, so perhaps not very cumbersome. 

I agree with your opinion. I will eliminate the need for the meas_planes argument in some functions that uses flow/gflow inside.

> BTW, do we need meas_plane argument for flow? I see it's to raise error but I feel it's a bit strange to add it just to throw errors?

In my opinion, yes. I think causal flow is a property of an underlying graph (graph with measurement planes), not a graph itself. Requiring meas_planes is not meant to raise an error. An underlying graph with either XZ or YZ measurement planes actually does not have a flow. I mean that "no flow" is also a correct answer. However, as @shinich1 mentioned, in some literature flow is defined without measurement planes because they restrict all measurement planes to "XY". So, it might be good idea not to require meas_planes in `flow` for those not familiar with multiple measurement planes

masa10-f: So, I will remain a conditional branch in `flow` but eliminate the need to set the `meas_planes` argument

graphix/gflow.py
masa10-f: Thank you! @pafloxy 

As you point out, the solver doesn't need to restrict its input to `nx.Graph()`. Furthermore, it doesn't need graph information to solve a system of equations. I'll remove arguments for `__init__` and separate the `get_adjacency_matrix` method out of the class.

graphix/gflow.py
```diff
@@ -461,3 +388,356 @@ def get_layers(l_k):
...
+    """
+    substituted_solution = set()
+    uncertainty_map = dict(zip(uncertainty, [0] * len(uncertainty)))
+    for solution_node in solution.values():
+        if solution_node.substitute(uncertainty_map):
+            substituted_solution |= {solution_node.node_index}
+    return substituted_solution
+
+
+def find_all_solutions(solutions, uncertainty):
```
shinich1, line 576: are you planning to add a method to find N possible solutions rather than all? There might be a case where just a few gflows are useful enough.

graphix/generator.py
```diff
@@ -55,8 +55,13 @@ def generate_from_graph(graph, angles, inputs, outputs, timeout=100):
     assert len(inputs) == len(outputs)
     measuring_nodes = list(set(graph.nodes) - set(outputs) - set(inputs))
 
+    if meas_planes is None:
+        meas_planes = {}
+        for i in measuring_nodes:
+            meas_planes[i] = "XY"
```
shinich1, line 61: maybe make it one-liner

graphix/gflow.py
```diff
@@ -461,3 +388,356 @@ def get_layers(l_k):
     for i in l_k.keys():
         layers[l_k[i]].append(i)
     return d, layers
+
+
+class GF2Solver:
+    """Solver for GF(2) linear equations"""
```
shinich1, line 394: any reference or description of the algorithm?

shinich1:
> gflow_finding_time
> 
> This is a benchmark compared with the previous implementation. Performance has been improved for graphs with moderate size.

really nice! 
maybe worth comparing with pyzx implementation of gflow-finding algorithm? up to you!

graphix/gflow.py
pafloxy: As much as I understand, here we are adding all possible elements from the  kernel space of the LHS to the solution of our linear equation. Usually a m,n system of rank r has 2^(n-r) solutions.
It might be the case that some of the solutions lead to patterns that are close to each other in terms of the unitaries they might implement (once we designate a measurement plane ), but I think it is still beneficial to have an indea of all possible gflows for a graph atleast for the sake of generality.

Did I understand it correctly @masa10-f ?

shinich1: @pafloxy thanks for the comment! I agree it's useful to find all indeed. 
I was just wondering whether it's useful to add optional method to _not_ find all but a few, because Masato mentioned it takes quite long to find all solutions.

pafloxy: I see. I think I would agree that to make it *optional*, given all solutions here refers to exponentially many in the size of the nullspace of the matrix it would slow down the algorithm considerably.

masa10-f: Sorry for the late reply. I was busy last week.
Thank you for your discussions. I'm afraid that we don't have good criteria for which N gflow to select. So, I implemented another option called 'abstract' that returns abstract nodes and uncertainty nodes of each layer. Though users will need to manually substitute uncertainty map as in `find_single_solution` or `find_all_solutions`, this approach offers flexibility in finding suitable solutions.

graphix/gflow.py
masa10-f: I didn't refer to anything in particular.
I generalize well-known Gauss-Jordan elimination onto GF2 and non-square matrix.

Step 1. Conduct row elimination and column pivoting when necessary(`GF2Solver.forward_elimination`). We get a standard form of n x m matrix and kernel of matrix here.
Step 2. We get solutions for each node(`GF2Solver.backward_substitution`).
Step 3. If mode is 'abstract', the solver returns a set of `SolutionNode` and uncertainty nodes(kernel of matrix). If mode is 'single' or 'all', the solver substitutes uncertainty map and returns concrete solutions.

shinich1: good, I think what you wrote above seems to be a good docstring for this class?

masa10-f: I updated the docstring just now

masa10-f:
> maybe worth comparing with pyzx implementation of gflow-finding algorithm? up to you!

In this PR, I originally intended to fix the conditional branch of gflow algorithm, not to improve performance, actually, I didn't improve the core of the linear equations solver. I don't really want to compare it with another package now.

pafloxy:
Sure. Nice work 👏

graphix/gflow.py
```diff
@@ -372,34 +303,36 @@ def search_neighbor(node, edges):
     return N
 
 
-def find_flow(g, v_in, v_out, meas_plane=None, timeout=100):
+def find_flow(graph, input, output, meas_planes=None, mode="single"):
```
shinich1, line 306: what's the mode argument?

graphix/gflow.py
```diff
@@ -87,145 +29,142 @@ def gflow(g, v_in, v_out, meas_plane=None, timeout=1000):
...
     else:
-        finished = False
-        exist = True
+        return gflowaux(graph, input, output | corrected, meas_planes, k + 1, l_k, g, mode=mode)
 
-    return v_out | c_set, g, l_k, finished, exist
 
-
-def flow(g, v_in, v_out, meas_plane=None, timeout=10000):
+def flow(graph, input, output, meas_planes=None):
```
shinich1, line 164: dont we need timeout anymore?

graphix/gflow.py
```diff
@@ -235,71 +174,65 @@ def flow(g, v_in, v_out, meas_plane=None, timeout=10000):
...
-    v_in: set
+    input: set
         set of node labels for input
-    v_out: set
+    output: set
         set of node labels for output
-    timeout: int
-        number of iterations allowed before timeout
+    meas_planes: int(optional)
+        measurement planes for each qubits. meas_planes[i] is the measurement plane for qubit i. Note that an underlying graph has a causal flow only if all measurement planes are "XY".
```
shinich1, line 184: mention that None will be interpreted as all being XY plane

graphix/gflow.py
```diff
@@ -461,3 +394,365 @@ def get_layers(l_k):
...
+class GF2Solver:
+    """Solver for GF(2) linear equations
+
+    Generalization of well-known Gauss-Jordan elimination onto GF2 and non-square matrix.
+
+    Step 1. GF2Solver.forward_elimination
+        Conduct row elimination and column pivoting when necessary.
+        We get a standard form of n x m matrix.
+    Step 2. GF2Solver.backward_substitution
+        We get solutions for each node and kernel of the matrix.
```
shinich1, line 408: maybe add code example in the docstring (optional)?

graphix/gflow.py
```diff
@@ -461,3 +394,365 @@ def get_layers(l_k):
...
+    """
+
+    def __init__(self):
+        """Constructor for GF2Solver"""
+        pass
+
+    def solve(self, adjacency_matrix, RHS_map, node_order_col):
```
shinich1, line 415: don't we want to generalize this class for things other than adjacency matrix? Just change argument names to be more general?

CHANGELOG.md
```diff
@@ -9,10 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- A new option whether to write down all maximally delayed gflow or not in `gflow.gflow` (#80)
+
 ### Changed
 
+- Implemented GF(2) solver independent from `z3-solver` (#80)
+
```
shinich1, line 17: Let's mention some major interface changes here (eg added or removed arguments of methods). I can do this later too

graphix/gflow.py
masa10-f: I missed the explanation, but it is the same option of `gflow` function.

graphix/gflow.py
masa10-f: I don't understand why `timeout` argument is necessary. A maximum iteration time is a number of layers, which is not so large for moderate graphs.

graphix/gflow.py
masa10-f: I'm afraid that a code example doesn't provide any more information than the current docstring because it's obvious.

graphix/gflow.py
shinich1: please explain what it is in the docstring.

graphix/gflow.py
shinich1: I mean, I do get why you removed it, flow finding algorithm is fast enough to be operationally fine without timeout. 
be sure to mention API change in changeling and why it's fine

graphix/gflow.py
shinich1: for example, something like this is useful
https://numpy.org/doc/stable/reference/generated/numpy.linalg.solve.html
this being obvious is quite subjective and it's always good to be overly clear how to use it.

graphix/gflow.py
masa10-f: Like below?
"adjacency_matrix" -> "LHS_matrix"
"RHS_map" -> "RHS_matrix"
"node_order_col" -> "variable_names"
"NODE" -> "variable"
"SolutionNode" -> "Solution"

graphix/gflow.py
shinich1: just use the example I had on my z3-based solver

graphix/gflow.py
shinich1: yes! looks good

masa10-f: I find that type of some variables should be changed for general GF2 solver, so it will take a little longer.

graphix/gflow.py
masa10-f: I initially implemented this class as an internal function of glow because this is quite a rare problem and contains a little complicated process, especially the part of `SolutionNode`. I wonder if it is worthwhile to explain the usage of this method in detail. What do you think about it?

shinich1: ok understood! I'll make this resolved.

graphix/gflow.py
masa10-f: I just decided to change the overall structure of the GF2 Solver for general inputs because a sophisticated linalg solver would be beneficial for graph problems that might be implemented in the future.
By the way, I wonder whether it is good to implement matrix class in a separate file like `graphix/linalg.py` as [pyzx](https://github.com/Quantomatic/pyzx/tree/master/pyzx) does. Some graph procedures, that graphix implements using networkx now, can be replaced with matrix procedures (which can be developed in C++ if needed). It would speed up the functions
What do you think about it? @shinich1

shinich1: @masa10-f Nice! `graphix.linalg.py` sounds good to me. 

Re graph manipulations, I am trying out rustworkx backend for graph-state simulation at the moment, but as you say matrix procedures might prove to be faster in the future. let's see how these go.

masa10-f: @shinich1 Thanks for your comment!
I will make a `graphix/linalg.py`

I think local complementation can be easily implemented with a NOT operation on a sub-adjacency matrix as following picture. So matrix backend could be used for graph-state simulator

File

shinich1: very nice idea!

masa10-f: Hi! @shinich1 
Are you planning to support [SymPy](https://docs.sympy.org/latest/index.html) in Graphix? I might think that Sympy or something like that could be useful for a parameterized Circuit or Pattern.

Anyway, I began to consider that symbolic variables should be used for the GF2 solver. In the current implementation, I'm using the `SolutionNode` class which is tailored for the gflow finding algorithm. However, it needs some modification to offer a more general user interface.

If you plan to use Sympy, I will use Sympy's symbolic class. Otherwise, I will make a symbolic variable class in `graphix/linalg.py`. I want to hear your opinion.

shinich1: @masa10-f Yes we do plan to add sympy, see #68 for @pafloxy 's implementation of parameterized pattern (which I believe is WIP). Let's add it, since we'll be adding it for #68 anyway.

masa10-f: Thanks! I'll use it!

shinich1 (approved):
looks good - please squash and merge. Use `fix and improve gflow-finding algorithm (#80)`

masa10-f:
@shinich1 Thanks!

PR #85, opened 2023-09-07
Add tox support for running tests and linting

king-p3nguin:
 
Then, please fill in below:

**Context (if applicable):**

- Added tox support for running tests and linting
- `MANIFEST.in` needs `include requirements.txt` statement for tox to recognize the file
- Modified CI to run tox

**Related issue:**


.github/workflows/ci.yml
```diff
@@ -25,21 +25,12 @@ jobs:
     - uses: actions/checkout@v3
```
shinich1, line 25: do we want to remove line 19 for the python versions? seems redundant?

king-p3nguin: As far as I know, GitHub Actions can only use one python version for each session, so we still need to use matrix to check for multiple python versions. I think it should not perform same test multiple times because we can specify what test to run on each python version at `[gh-actions]` in `tox.ini` using `tox-gh-actions`.

shinich1 (approved):

Issue #86, opened 2023-09-20
Graph state generation compiler


shinich1:
**Is your feature request related to a problem? Please describe.**
In discrete-variable photonic QPUs, graph states are often generated by performing fusion on small entangled states (linear clusters, GHZ states, etc). To include this in the simulation and and potentially optimize it, we would want to be able to compile the fusion scheduling for desired graph state.

**Describe the feature you'd like**
a class or function that decomposes a graph state into smaller entangled states that can be fused together to generate the desired graph.

**Additional context**

PR #87, opened 2023-09-20
Support extraction of clusters from graph state

king-p3nguin:

Then, please fill in below:

**Context (if applicable):**
#86 

- Add `Cluster` class to define extracted cluster.
- Add function `extract_clusters_from_graph` which will extract GHZ and linear clusters from `graphix.GraphState`.

**Related issue:**
Related to #86 

### DIFF
--- a/docs/source/device_interface.rst
+++ b/docs/source/device_interface.rst
@@ -2,7 +2,7 @@ Device Interface
 ==================
 
 :mod:`graphix.device_interface` module
-+++++++++++++++++++++++++++++++++++
+++++++++++++++++++++++++++++++++++++++
 
 .. currentmodule:: graphix.device_interface
 
@@ -14,4 +14,3 @@ Device Interface
 
     .. automethod:: retrieve_result
 
-    .. automethod:: format_result
--- a/docs/source/extraction.rst
+++ b/docs/source/extraction.rst
@@ -1,0 +1,18 @@
+Graph state generation
+======================
+
+:mod:`graphix.extraction` module
+++++++++++++++++++++++++++++++++
+
+This provides functions to extract clusters from a given the graph state.
+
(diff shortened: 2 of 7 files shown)


shinich1:
@king-p3nguin could you add API docs?

shinich1:
Perhaps an example too?

shinich1:
do we want to mention fusion-based QC ref or something similar in the docstring?

king-p3nguin:
I made the following changes:
- Previous extraction algorithm did not consider isolated nodes in the graph state as a cluster. Now it does.
- Added API docs
- Added an example
- Mentioned [1] for the definition of fusion in the example.
[1] Daniel E. Browne and Terry Rudolph. Resource-efficient linear optical quantum computation. Physical Review Letters, 95(1):010501, 2005.

king-p3nguin:
I also modified some docs to resolve sphinx errors.

shinich1 (approved):

shinich1:
@king-p3nguin looks good to me, please squash and merge! let's use commit message `fusion network extraction for graph state generation (#87)`.
after this is merged, I will change some names e.g. `extract_clusters_from_graph` because cluster state sounds like it's larger and extract_fusion_network may make more sense. what do you think?

king-p3nguin:
@shinich1 Thank you for reviewing! I will merge this now.
Come to think of it, as you mentioned, the word "cluster" may create confusion by making people think of cluster states rather than mini clusters in the graph state. Maybe `get_fusion_network` or `get_fusion_network_from_graph_state` will be nicer, since the size of the fusion network is similar to the graph state.

shinich1:
@king-p3nguin your suggestions look good! I will make these changes and get new release out soon, so that we can put the new version in the graphix-Perceval's requirements.

PR #88, opened 2023-09-20
Add input_nodes attribute

d1ssk:

Then, please fill in below:

**Context (if applicable):**
In the current implementation, the `Pattern` class does not explicitly possess the attribute of input nodes. Instead, it implicitly assumes that nodes with indices ranging from 0 to (width-1) are the inputs. It is considered preferable to have `input_nodes` as an attribute to accommodate more general situations.

- Add `input_nodes` attribute to the `Pattern` class and the `LocalPattern` class
- Add `input` attributes to the `CommandNode` class
- Add implementations to set the `input_nodes` during various pattern generation processes (e.g., transpiling from circuits, generating from graphs).
- Add a `leave_inputs` option to `perform_pauli_measurements()` (default: False). When set to True, input nodes will not be removed even if their measurement basis is Pauli. When set to False, indices of removed input nodes will be deleted from the `input_nodes` attribute.

**Related issue:**


graphix/pattern.py
```diff
@@ -1653,7 +1680,11 @@ def measure_pauli(pattern, copy=False):
     vop_init = pattern.get_vops(conj=False)
     graph_state = GraphState(nodes=nodes, edges=edges, vops=vop_init)
     results = {}
-    to_measure, non_pauli_meas = pauli_nodes(pattern)
+    to_measure, non_pauli_meas = pauli_nodes(pattern, leave_inputs)
+    if leave_inputs:
+        new_inputs = pattern.input_nodes
+    else:
+        new_inputs = list(set(pattern.input_nodes) - set([i[0][1] for i in to_measure]))
```
shinich1, line 1687: It might be clearer to set entire input_nodes to None, or replace the measured input index with None. This is because leaving only part of indices here makes it harder to have mapping between inputs and outputs. Also having non-equal number of inputs and outputs is potentially confusing.

I would suggest setting input_nodes to None if at least one of the input nodes are Pauli-measured. What's your thoughts @masa10-f

graphix/pattern.py
```diff
@@ -43,14 +43,15 @@ class Pattern:
         total number of nodes in the resource state
     """
 
-    def __init__(self, width=0, output_nodes=[]):
+    def __init__(self, width=0, input_nodes=[], output_nodes=[]):
```
shinich1, line 46: I think we can get rid of width argument here. maybe we can get rid of Nnode attribute too??

shinich1:
Is the intention to merge this before the resource state visualisation?

d1ssk:
Yes, my intention is to merge this change first, then make a minor update to the resource state visualisation and after that, merge the visualization.

graphix/pattern.py
d1ssk: As you say, the __init__ procedure of the Pattern class seems somewhat redundant. I think it can be more concise along with changes to the pattern generation code. I will try that.

d1ssk: The dependency relation of attributes is (seq→output_nodes, width, Nnode), (output_nodes→width), input_nodes. (The right side of the arrow is determined if the left side is given.) 
We can use this to change the implementation so that the attributes assignment would be more consistent and concise for various types of pattern generation, but I think it is better to do this in other pull requests since there will be many changes.

graphix/pattern.py
d1ssk: That seems reasonable. I changed this part.

graphix/pattern.py
shinich1: ok sounds good.

masa10-f:
@d1ssk 
The code looks good!
Could you add a test for the 'leave_input' option of 'perform_pauli_measurements' into 'test_pattern.py' to automatically verify the new option functions correctly?

masato-fuk (approved):
Code looks good.
Thank you for adding test! @d1ssk

shinich1:
@d1ssk please squash and merge, with commit message `add input_nodes attribute to pattern class (#88)`

Issue #84, opened 2023-09-06
Benchmarking page on the documentation


shinich1:
**Describe the feature you'd like**
Add a page for benchmarking (e.g. ones on #74, #43, memory usage of TN simulation, #81, etc) to the docs.

**Additional context**
Do leave benchmarking script somewhere: a dedicated page on documentation (maybe only reference from the benchmarking page), or make a subfolder in `example`?

PR #43, opened 2023-04-17
42 fast pattern operation with distributed command sequence

masato-fuk:
Before submitting, please check the following:
- Make sure you have tests for the new code and that test passes (run `pytest`)
- format added code and tests by `black -l 120 <filename>`
- If applicable, add a line to the [unreleased] part of CHANGELOG.md, following [keep-a-changelog](https://keepachangelog.com/en/1.0.0/).
 
Then, please fill in below:

**Context (if applicable):**

na

**Description of the change:**

Implemented `LocalPattern` and `CommandNode`. With these new classes, `standardize` and `shift_signals` have been significantly accelerated.
Default backend for pattern operations is changed to `LocalPattern` class.

**Related issue:**

#42 

also see that checks (github actions) pass.
If lint check keeps failing, try installing black==22.8.0 as behavior seems to vary across versions.

PR #89, opened 2023-09-26
84 benchmarking page

nabe98:

Then, please fill in below:

**Context (if applicable):**
In the current implementation, there is no benchmarking page although there were some benchmarking examples (e.g. ones on https://github.com/TeamGraphix/graphix/pull/74, https://github.com/TeamGraphix/graphix/pull/43, memory usage of TN simulation, https://github.com/TeamGraphix/graphix/pull/81, etc)

- Add benchmarks directory to store sample codes for benchmarking
- Add benchmarking code (statevec.py) for `remove_qubit`
  - `graphix` vs `paddle_quantum` vs `mentpy` with reference circuit simulator on `graphix`

**Related issue:**

#84 


shinich1:
@nabe98 I see the readthedocs build is now working, great! Do you want me to review this PR?

benchmarks/statevec.py
```diff
@@ -0,0 +1,208 @@
...
+    grid_cluster = mp.templates.grid_cluster(n=width, m=depth)
+    random_state = mp.utils.generate_haar_random_states(n_qubits=width)
+    simulator = mp.PatternSimulator(grid_cluster, backend="numpy-sv", window_size=width)
+    simulator.reset(input_state=random_state)
+    angles = np.zeros(width)
+    start = perf_counter()
+    simulator.run(angles=angles)
+    end = perf_counter()
+    mentpy_time.append(end - start)
+
```
shinich1, line 183: is this a fair comparison??

benchmarks/statevec.py
```diff
@@ -0,0 +1,208 @@
...
+        generated circuit
+    """
+    qubit_index = [i for i in range(width)]
+    circuit = PaddleCircuit(width)
+    for _ in range(depth):
+        np.random.shuffle(qubit_index)
+        for j in range(len(qubit_index) // 2):
+            circuit.cnot(which_qubits=[qubit_index[2 * j], qubit_index[2 * j + 1]])
+        for j in range(len(qubit_index)):
+            circuit.rz(which_qubit=qubit_index[j], theta=to_tensor(2 * np.pi * np.random.random(), dtype="float64"))
```
shinich1, line 147: instead of generating randomly for each packages, we should simulate for exact same circuits. will update the code.

benchmarks/statevec.py
```diff
@@ -0,0 +1,208 @@
...
+        elif cmd[0] == "X":
+            backend.correct_byproduct(cmd)
+        elif cmd[0] == "Z":
+            backend.correct_byproduct(cmd)
+        elif cmd[0] == "C":
+            backend.apply_clifford(cmd)
+        else:
+            raise ValueError("invalid commands")
+        if backend.pattern.seq[-1] == cmd:
+            backend.finalize()
```
shinich1, line 53: why not use `pattern.simulate_pattern()` as stated above? perhaps this is for 'old ptrace method' which you had below (I removed this, because there's no point benchmarking our old, inefficient code)

requirements.txt
```diff
@@ -1,5 +1,5 @@
 numpy>=1.22
-networkx>3.0
+networkx>=2.8,<3.0
```
shinich1, line 2: what's the reason for this?

nabe98: It is due to mentpy which requires networkx >=2.5,<3.0

shinich1: why do it in main requirement and not in docs/requirements? Doesn't make sense to change graphix's dependency for mentpy requirements

nabe98: Sorry, it was my mistake. I will be careful to keep dependencies separable from the main implementation from the next time.

shinich1 (approved):
all look good - except that the fig size in each gallery page is fixed and distorting the plots, for example visualisation plot of https://graphix.readthedocs.io/en/84-benchmarking-page/gallery/deutsch-jozsa.html#sphx-glr-gallery-deutsch-jozsa-py. Please fix this before squash-and-merge. use `add benchmarking page (#89)` as commit message.

Issue #90, opened 2023-10-03
Add support for python 3.11?


king-p3nguin:
Python 3.12.0 was just released today (Oct. 2, 2023), and it might be a good time to add support for Python 3.11.

shinich1:
yes perhaps a good idea! we can consider stopping support for 3.8 because newer versions of numpy stopped supporting it (throws error when trying to install all dependency by `python setup.py develop` with py3.8 because of that)

king-p3nguin:
OK, I will make a PR on this.

PR #91, opened 2023-10-05
Add support for Python 3.11

king-p3nguin:
**Context (if applicable):**
#90 

- Using the current code on Python 3.11 does not seem to have a problem, so I just modified the CI, `setup.py`, and `tox.ini`.
- Mocks were introduced to remove the dependency on `graphix_ibmq` during testing.

Currently, when running the `PatternRunner` test, `graphix_ibmq` is imported to see if `PatternRunner` works correctly. This strategy for testing checks the functionality of `PatternRunner` and `graphix_ibmq` at the same time, so when an error occurs, it is difficult to tell whether the problem derives from `PatternRunner` or `graphix_ibmq`. Also, because `graphix` imports `graphix_ibmq` and `graphix_ibmq` imports `graphix`, the test will not work correctly until `graphix_ibmq` supports Python 3.11. It is not ideal to encounter this issue every time `graphix` adds support for a newer Python version, so I avoided it by running the test assuming that `graphix_ibmq` is working correctly during testing. 

`unittest.mock.MagicMock` provides the ability to mock dependent modules.
https://github.com/TeamGraphix/graphix/blob/2eb09789e4005139820215ec709ba0b89ae88fdf/tests/test_runner.py#L44-L47

**Related issue:**
#90


shinich1 (approved):
Looks good to me - thanks! please squash and merge.

Issue #79, opened 2023-08-23
Faster graph state simulator


shinich1:
**Is your feature request related to a problem? Please describe.**
The graph state simulator is slow for large-scale simulations, for example the pauli measurement preprocessing part of https://github.com/TeamGraphix/graphix/blob/master/examples/qft_with_tn.py 

**Describe the feature you'd like**
add rustworkx as backend for the graph state simulator. Perhaps best to keep it optional since it is not supported for all platforms.
Also consider adding 'stabilizer backend' for MBQC simulations (simulate with graph state simulator), which may be useful for e.g. QEC simulation.

king-p3nguin:
Hi @shinich1, if no one is working on it, I would like to implement a graph state simulator with `rustworkx`.  I am thinking of creating another class for graph state simulator with `rustworkx` and abstract graph state class to store common methods.

shinich1:
@king-p3nguin sounds great! it would be good to benchmark full rustworkx implementation of graphstate simulator.
I was thinking about switching the backend between networkx and rustworkx internally of the graph state class based on availability of rustworkx (try-catch type syntax for `import rustworkx`; with option to specify by hand explicitly which to use), what do you think about that? also see #95 for @masa10-f 's idea on speeding up local complementation by matrix backend.

king-p3nguin:
I understand. In that case, how about something like
```python
class GraphState:  # a factory class for graph states
    def __init__(self, use_rustworkx=True):
        if use_rustworkx:
            # check if rustworkx is installed
            # ...
            if rustworkx_installed:
                return RustworkxGraphState()
            else:
                throw_warning()
                return NetworkxGraphState()
        else:
            return NetworkxGraphState()

class AbstractGraphState:
    @abstractmethod
    def ...

class NetworkxGraphState(AbstractGraphState):
    def ...

class RustworkxGraphState(AbstractGraphState):
    def ...
```
?

king-p3nguin:
@shinich1

Issue #65, opened 2023-06-26
Avoid using random numbers for tests


king-p3nguin:
**Is your feature request related to a problem? Please describe.**
The use of random numbers in testing can be cited as a reason why transpiler bugs were not found until recently. A simple way to avoid this is to fix the seed value, but the best approach is to increase the number of deterministic test cases that cover various scenarios without relying on random numbers.

**Describe the feature you'd like**
Create as many test cases as possible to cover various scenarios in `tests/test_pattern.py` and `tests/test_transpiler.py`.

**Additional context**
- #54 
- #62

shinich1:
@king-p3nguin thank you for the issue report! 
Actually we've added a new class of circuits in the tests: see [here](https://github.com/TeamGraphix/graphix/pull/63/files), random_circuit.py, and tests now run for 10 different realisations of this. Do you think it can be generalised/improved? Please let me know what you think.

king-p3nguin:
@shinich1 
In that case, I think it would be better to test all the circuits that `get_rand_circuit` can create instead of picking up certain circuits. It will definitely increase the time for testing, but since the transpiler is one of the most important features in graphix, it should not be a bad price to pay.

mgarnier59:
Also, we have starting using the `hypothesis`package that is good at generating strategies and finding pathological cases. Sometimes it's a bit too much (like testing for NaN parameter values) but it can be very useful.
We'll take a closer look at the testing strategy and give you feedback!

PR #98, opened 2023-11-15
Add `rustworkx` as a backend for the graph state simulator

king-p3nguin:
# TODO:
- [x] benchmarking page

# Details
- Use `rustworkx` for `GraphState` when it is installed

I added five classes as follows:
- `GraphState` class
  - A factory class that returns an instance of either `NXGraphState` or `RXGraphState`.
  - If `rustworkx` is installed and `use_rustworkx=True`, it returns `RXGraphState`.
  - If `rustworkx` is not installed and `use_rustworkx=True`, it returns `NXGraphState` and throws a warning.
  - If `rustworkx` is installed and `use_rustworkx=False`, it returns `NXGraphState`.
  - `use_rustworkx=True` is the default.
- `BaseGraphState` class
  - It implements methods that do not depend on the backend and defines abstract methods for maintainability.
- `NXGraphState` class
  - A graph state simulator based on `networkx.Graph`.
- `RXGraphState` class
  - A graph state simulator based on `rustworkx.PyGraph`
- `NodeList` and `EdgeList` class
  - It provides functionality similar to `NodeView` and `EdgeView` in `networkx`.
  - `rustworkx.PyGraph` stores the information of the node by its index, not by the name like in `networkx.Graph` which is a major difference. This class will make it possible to specify the node by its name.
  ```python
  from rustworkx import PyGraph
  nodes = [1, 2, 3]
  graph = PyGraph()
  graph.add_nodes_from(nodes)
  graph[0] # returns 1
  ```
  - Note that when you add nodes to `NodeList`, you need to call the node data `{"loop": False, ...}` by reference otherwise, modification of the node data like `g.node[0]["loop"] = True` will not affect the node data in graph object (i.e. it will still be `g.graph[0][1]["loop"] = False`)

I also:
- Added tests to test the graph state simulator for both backends.
- Changed `setup.py` to install `rustworkx` while testing

**Related issue:**
#79


king-p3nguin:
It seems like there is an issue in inheriting `rustworkx` classes. Running the following code will throw `TypeError`:

```python
from rustworkx import PyGraph


class A(PyGraph):
    def __init__(self, nodes=None) -> None:
        super().__init__()
        if nodes is not None:
            self.add_nodes_from(nodes)


x = A(nodes=[0, 1, 2])  # TypeError: PyGraph.__new__() got an unexpected keyword argument 'nodes'
```

Instead of inheriting a graph object, I will put a graph object in `self.graph` to avoid this.

king-p3nguin:
I think it is ready for review. I updated the description.

masa10-f:
@king-p3nguin 

Thanks for your great contributions!!
I will review the details of the code tomorrow.

Just a quick review, the current implementation does not work without rustworkx installation although it is not required in the requirements.txt.
This is because `rx.PyGraph` is called at `graphix.graphsim.basegraphstate.py` l.25. Could you check the whole code without rustworkx installation?

masa10-f:
Also, would you take a performance comparison between the both graph backends?

king-p3nguin:
@masa10-f 

> @king-p3nguin
> 
> Thanks for your great contributions!! I will review the details of the code tomorrow.
> 
> Just a quick review, the current implementation does not work without rustworkx installation although it is not required in the requirements.txt. This is because `rx.PyGraph` is called at `graphix.graphsim.basegraphstate.py` l.25. Could you check the whole code without rustworkx installation?

Oops, I forgot to check that. I fixed it and changed the `tox.ini` to make the test work both with and without `rustworkx`.

graphix/graphsim/rxgraphviews.py
```diff
@@ -0,0 +1,123 @@
...
+        self.nodes.add(nnum)
+        self.num_to_data[nnum] = ndata
+        self.num_to_idx[nnum] = nidx
+
+    def add_nodes_from(self, node_nums: list[int], node_datas: list[dict], node_indices: list[int]):
+        if not (len(node_nums) == len(node_datas) and len(node_nums) == len(node_indices)):
+            raise ValueError("node_nums, node_datas and node_indices must have the same length")
+        for nnum in node_nums:
+            if nnum in self.num_to_data:
+                raise ValueError(f"Node {nnum} already exists")
```
masa10-f, line 45: This part overlaps with `add_node`

graphix/graphsim/basegraphstate.py
```diff
@@ -235,9 +415,15 @@ def equivalent_graph_E2(self, node1, node2):
...
             connected graph nodes to apply rule E2
+
+        Returns
+        ----------
+        None
         """
-        assert (node1, node2) in list(self.edges) or (node2, node1) in list(self.edges)
-        assert not self.nodes[node1]["loop"] and not self.nodes[node2]["loop"]
+        if (node1, node2) not in list(self.edges) and (node2, node1) not in list(self.edges):
+            raise ValueError("nodes must be connected by an edge")
```
masa10-f, line 424: From my profiling result, this assertion is the bottleneck of the Pauli measurements procedures. So could you delete this part, or convert it to a more efficient one if possible?

graphix/graphsim/rxgraphviews.py
masa10-f, line 45: This part overlaps with `add_node`

graphix/graphsim/rxgraphstate.py
```diff
@@ -0,0 +1,144 @@
...
+            # adding edges may add new nodes
+            if u not in self.nodes:
+                nidx = self._graph.add_node((u, {"loop": False, "sign": False, "hollow": False}))
+                self.nodes.add_node(self._graph[nidx][0], self._graph[nidx][1], nidx)
+            if v not in self.nodes:
+                nidx = self._graph.add_node((v, {"loop": False, "sign": False, "hollow": False}))
+                self.nodes.add_node(self._graph[nidx][0], self._graph[nidx][1], nidx)
+            uidx = self.nodes.get_node_index(u)
+            vidx = self.nodes.get_node_index(v)
+            eidx = self._graph.add_edge(uidx, vidx, None)
```
masa10-f, line 127: This part seems to be the bottleneck of the rustworkx-based simulator. I'm sorry for not having any suggestions, but could you consider if it can be improved?

graphix/graphsim/graphstate.py
```diff
@@ -0,0 +1,26 @@
...
+class GraphState:
+    """Factory class for graph state simulator."""
+
+    def __new__(self, nodes=None, edges=None, vops=None, use_rustworkx: bool = True) -> BaseGraphState:
+        if use_rustworkx:
+            if RUSTWORKX_INSTALLED:
+                return RXGraphState(nodes=nodes, edges=edges, vops=vops)
+            else:
+                warnings.warn("rustworkx is not installed. Using networkx instead.")
+        return NXGraphState(nodes=nodes, edges=edges, vops=vops)
```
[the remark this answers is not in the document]
shinich1: I would agree with @masa10-f, let's make rust one optional even though many may end up using the option.

graphix/graphsim/rxgraphstate.py
king-p3nguin: I think I figured out the problem.
It seems like I should have defined `__contains__` method for `NodeList` and `EdgeList`, otherwise the `in` uses `__iter__` method to make `in` work and the conversion of `set` to `iter` affected the performance.

```python
class TestIn1:
    def __init__(self, x):
        self.x = set(x)

    def __iter__(self):
        return iter(self.x)

ins1 = TestIn1(list(range(10000000)))
```

```python
%%timeit
59999 in ins1
```
out: `406 µs ± 4.04 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)`

```python
class TestIn2:
    def __init__(self, x):
        self.x = set(x)

    def __iter__(self):
        return iter(self.x)

    def __contains__(self, item):
        return item in self.x

ins2 = TestIn2(list(range(10000000)))
```

```python
%%timeit
59999 in ins2
```
out:`45.8 ns ± 0.414 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)`

masa10-f:
I benchmarked the performance of both graphsim backends for QFT and random quantum circuits.
For graphs smaller than $10^3$ nodes, rustworkx performs better. However, for larger graphs, its performance was the same as or worse than the networkx backend. I wonder why rustworkx's performance is not as good as networkx's. If we can not resolve this problem soon, we should address it in a separate issue.

rc_graphsim_benchmark
qft_graphsim_benchamark

masa10-f:
It looks like there is an error occurs in `tests/test_pattern.py::TestPattern::test_minimize_space_with_gflow`. The issue seems to be that the quantum circuit for this test is generated randomly, so the output of `pattern.minimize_space` varies each time we run the test. Also, the maximum space of a pattern with measured Pauli nodes doesn't have a non-trivial upper limit, unlike regular patterns. We could fix this error by reducing the circuit width or fix the random seed. Could you update the test code to address this?

king-p3nguin:
> It looks like there is an error occurs in `tests/test_pattern.py::TestPattern::test_minimize_space_with_gflow`. The issue seems to be that the quantum circuit for this test is generated randomly, so the output of `pattern.minimize_space` varies each time we run the test. Also, the maximum space of a pattern with measured Pauli nodes doesn't have a non-trivial upper limit, unlike regular patterns. We could fix this error by reducing the circuit width or fix the random seed. Could you update the test code to address this?

I changed the implementation of `tests/random_circuit.py`. `rc.set_seed(seed_value)` will fix the seed throughout the testing process. Related to #65 .

king-p3nguin:
It seems like the use of `nx.isolates(G)` for RustworkX graph state was causing the problem. The following code will not throw assertion error when `use_rustworkx=False`.

```python
from tests import random_circuit as rc
import numpy as np
from copy import copy
nqubits = 5
depth = 5
pairs = [(i, np.mod(i + 1, nqubits)) for i in range(nqubits)]
circuit = rc.generate_gate(nqubits, depth, pairs, seed=42)
circuit_copy = copy(circuit)

pattern = circuit.transpile()
pattern.standardize(method="global")
pattern.shift_signals(method="global")
pattern.perform_pauli_measurements(use_rustworkx=True)

pattern2 = circuit_copy.transpile()
pattern2.standardize(method="global")
pattern2.shift_signals(method="global")
pattern2.perform_pauli_measurements(use_rustworkx=True)

assert pattern.seq == pattern2.seq  # AssertionError
```

masa10-f:
@shinich1 
I have a question about the graph state simulator. When it receives the same measurement pattern, will the processed graph state always be the same, or can they vary?

shinich1:
> @shinich1 I have a question about the graph state simulator. When it receives the same measurement pattern, will the processed graph state always be the same, or can they vary?

Pauli measurements in the graphs state simulator do have some degree of freedom (for example, depending on the order of node coming up in `neighbors` the resulting graph structure can change via [this code](https://github.com/TeamGraphix/graphix/blob/8313d3c773c71920836b5cd1ac2ce1a485e7a1ad/graphix/graphsim.py#L302)). so if you have different backend you might end up with LC-equivalent graph states with different underlying graph.

king-p3nguin:
As you say, it looks like `.neighbors()` in rustworkx returns a different result each time, making a difference in the pattern. The algorithm does not seem to have a problem, so I just reduced the circuit size in the tests.

king-p3nguin:
(I will add the benchmarking page later.)

shinich1:
@king-p3nguin how about comparing with Anders&Briegel graph sim? e.g. https://github.com/libtangle/graph-state

graphix/graphsim/rxgraphstate.py
```diff
@@ -89,10 +89,7 @@ def remove_node(self, node: int) -> None:
         nidx = self.nodes.get_node_index(node)
         self._graph.remove_node(nidx)
         self.nodes.remove_node(node)
-        edge_list = list(self.edges)
-        for e in edge_list:
-            if node in e:
-                self.edges.remove_edge(e)
+        self.edges.remove_edges_by_node(node)
```
king-p3nguin, line 93: I think this part worsen the performance of `RXGraphState`, since it searches all the edges in the graph every time removing the nodes. This is fixed, so the performance of `RXGraphState` will be at least similar to that of `NXGraphState`.

king-p3nguin:
@shinich1 
Thank you for the suggestion! I added their graph state simulator to the benchmark.
image

shinich1:
@king-p3nguin another point - I tried to implement from Anders&Briegel paper some time ago but never managed to make it return the correct result - and graph-state do not seem to have unit tests. 

it might be tricky due to LC and local complementation degree of freedom, but could you check that the graph-state is always returning the correct result (compare with our graph sim, by turning them into statevec)?

graphix/graphsim/rxgraphstate.py
masa10-f: Great! Thanks!

benchmarks/graphsim.py
```diff
@@ -0,0 +1,168 @@
...
+DEPTH = 3
+test_cases = [i for i in range(2, 300, 10)]
+graphix_patterns = {}
+
+for i in test_cases:
+    circuit = random_clifford_circuit(i, DEPTH)
+    pattern = circuit.transpile()
+    pattern.standardize()
+    nodes, edges = pattern.get_graph()
+    graphix_patterns[i] = (circuit, pattern, len(nodes))
```
masa10-f, line 75: We recommend executing `pattern.shift_signals` here to fully reduce Pauli nodes.

benchmarks/graphsim.py
```diff
@@ -0,0 +1,168 @@
...
+    rustworkx_node.append(num_nodes)
+    print(f"width: {width}, number of nodes: {num_nodes}, depth: {DEPTH}, time: {end - start}")
+    rustworkx_time.append(end - start)
+
+# %%
+# For reference, we also perform Anders and Briegel's stabilizer circuit simulator based on
+# graph state representation.
+# See https://arxiv.org/abs/quant-ph/0504117v2 or https://github.com/libtangle/graph-state for more details.
+
+from graph_state import GraphState as ABGraphState
```
masa10-f, line 113: Can we include it in the requirements.txt?
Ref. https://stackoverflow.com/questions/16584552/how-to-state-in-requirements-txt-a-direct-github-source

shinich1:
> @masa10-f Thank you for the review! @shinich1 Actually, I always build the document locally to check it (that's why I forgot to push the requirements.txt). Should we make ci to test building the document (only when it is modified)?

it's good to check on RTD because it always start with a fresh environment defined by requirements.py (for example, it could not have worked without addition of graph-state as pointed out above by @masa10-f). CI is not really necessary for this.

shinich1 (approved):
all looks good!

@king-p3nguin A final point, could we remove Anders-Briegel method from the benchmark for now? It's not worth comparing with package that may not actually work in practice (sorry I should have remembered about this issue earlier). I will also look at the code.
after this final change, please squash and merge.

Issue #97, opened 2023-11-12
[Bug]: visualising with gflow


shinich1:
**Describe the bug**
singular matrix error occurs with visualisation tool when drawing graph with gflow.

**To Reproduce**
https://github.com/TeamGraphix/graphix/blob/9661708be39903b7431dc49daff0270afd0f865a/examples/qnn.py
(note - I have done temporary docs fix by adding manual visualisation in the latest commit on master)

**Expected behavior**

**Environment (please complete the following information):**
 - OS: Mac OS
 - Python version: 3.8.6
 - Related module versions if applicable: [e.g. numpy=1.23.5]

**Additional context**

masa10-f:
Supplementary information

graph structure
![output](https://github.com/TeamGraphix/graphix/assets/63156491/5f89e35a-5414-4bb7-9bc2-e3331783862d)

and

gflow = {23: {23}, 25: {25}, 28: {27, 28}, 33: {33}, 22: {27}, 31: {27, 31}, 21: {28}, 30: {35}, 20: {20, 21}, 29: {30}, 5: {5}, 7: {7}, 13: {13}, 15: {15}, 4: {20, 28}, 12: {28}, 3: {4}, 11: {12}}.

The error occurred in visualization.py l.211
```
nx.draw_networkx_edges(self.G, pos, edgelist=[edge], edge_color="black", arrowstyle="->", arrows=True)
```

shinich1:
cc: @d1ssk

d1ssk:
I found that the issue is in the `get_figsize()` function, which returns a height of zero when the `input_nodes` attribute of the pattern is None. The plot below illustrates the resource state of the qnn after Pauli measurements, incorporating a quick bug fix. I plan to commit changes after addressing the following issues:
<img width="514" alt="スクリーンショット 2023-11-14 8 22 59" src="https://github.com/TeamGraphix/graphix/assets/52164693/623fd753-fa3f-49c9-bb19-cbbdb09c90d0">

- Self-arrows are hidden by the nodes.
- The tips of curved arrows are hidden by the nodes.
- The nodes and lines are crowded and difficult to discern. An option to set custom node distances needs to be added.

shinich1:
perhaps off topic for this issue, but another thing to consider: add two more node edge colors to express measurement planes?

PR #107, opened 2023-12-15
bugfix gflow visualization

d1ssk:

Then, please fill in below:

**Context (if applicable):**
Singular matrix error occured with visualisation tool when drawing graph with gflow. (#97 )

The direct cause of the error was in the `get_figsize()` method, which returned 0 height for pattern without input_nodes. Fixed that part and modified some detailed drawing methods for better gflow visualization.

**Related issue:**
#97 

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ this will install `graphix` and inteface for [IBMQ](https://github.com/TeamGraph
 
 ## Using graphix
 
-### Generating pattern from a circuit and inspecting the resource graph with information flow (arrows)
+### generating pattern from a circuit
 ```python
 from graphix import Circuit
 circuit = Circuit(4)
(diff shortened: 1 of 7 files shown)


d1ssk:
As a example, these are visualizations of QNN pattern after Pauli measurements. The first one was generated by `pattern.draw_graph()` and the second one by `pattern.draw_graph(node_distance=(1.2, 1.6))` to broaden node distances for better visualization. 
スクリーンショット 2023-12-15 21 18 12
スクリーンショット 2023-12-15 21 18 03

shinich1 (approved):
@d1ssk looks good! please see my changes and if you are happy about them, please squash and merge.

shinich1:
@d1ssk if you have time - can you add type hints to visualization.py? I think it's time to start adding them part by part..

README.md
```diff
@@ -29,30 +28,54 @@ $ pip install graphix[extra]
...
+from graphix import Circuit
+circuit = Circuit(4)
+circuit.h(0)
+...
+pattern = circuit.transpile()
+pattern.draw_graph()
+```
+<img src="https://github.com/TeamGraphix/graphix/assets/33350509/de17c663-f607-44e2-945b-835f4082a940" alt="logo" width="750">
+
+<small>note: this graph is generated from QAOA circuit, see [our example code](examples/qaoa.py). Arrows indicate the [*information flow*](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.74.052310) of MBQC and dashed lines are the other edges of the graph. the vertical dashed partitions and the labels 'l:n' below indicate the execution *layers* or the order in the graph (measurements should happen from left to right, and nodes in the same layer can be measured simultaneously), based on the partial order associated with the (maximally-delayed) flow. </small>
```
masa10-f, line 45: Isn't it a "causal flow" rather than an "information flow"? Flow does not always transfer information especially in generalized flow.

shinich1: good point, I thought information flow is more accessible to non-specialist (here it's only referring to flow) but let's be precise. updated this part.

masa10-f: Thanks for the update!

>  the vertical dashed partitions and the labels 'l:n' below indicate the execution layers or the order in the graph (measurements should happen from left to right, and nodes in the same layer can be measured simultaneously), based on the partial order associated with the (maximally-delayed) flow.

This part explains well what flow is in MBQC

d1ssk:
It seems that python<=3.8 and python>=3.9 have different ways of adding type hints. (Either include the typing module or use the built-in type hints class). Currently, graphix requirements are >=3.8,<3.12, so we need to decide which one we should use and make it consistent across the library. This time, I'm merging this PR without adding them.

pafloxy:
I am not sure if this is the right place to ask this, but sometimes(or maybe always) there are edges with double arrows are showinf up in the plots which I am not sure how to interpret :/. Like if there are the elements from the correction-set then the arrows should have single heads no ? 

---
For example 
image


here is a pettern where the correction sets are as follows : 
 [{2: {1}, 3: {0, 1}, 4: {1, 3}}]

all vertices measured in XY, {0,1} are the outputs and {4,2} the inputs

---

Can you point out wht am I doing wrong ? 

I instantiated `graphix.vizualisation.GraphVisualizer` on the adjacenecy matrix 
[[0, 0, 1, 1, 0],
     [0, 0, 1, 0, 0],
     [1, 1, 0, 0, 1],
     [1, 0, 0, 0, 1],
     [0, 0, 1, 1, 0]]
i,e 
image

and the information about the input an the outputs. and used  classmethod `visualize_w_gflow()` to generate the plot.

d1ssk:
I have investigated this issue and found a basic logic error in the visualization method that had not been revealed until now. I will commit the change to fix this in PR #113. Thank you for pointing it out!

pafloxy:
Ahn okay. glad to be of help ;). But can you maybe hint me at the problem ? So that I can bypass it somehow for my own tasks for a while before you fix it in the mentioned PR.

d1ssk:
Sure. The problem had lain in the section where the edges to be drawn are collected, lines 547 and 548 of `visualization.py`. You can bypass this by modifying this part according to the latest commit in PR #113.

PR #108, opened 2023-12-15
Implement and expose noisy DM backend + utilities

mgarnier59:
Tests pass.
Formatted.

Then, please fill in below:

**Context (if applicable):**
Developed and exposed the noisy density matrix backend. 

- Adds classes for channel creation and operation as Kraus operators
- Changes to the NoiseModel definition (only returns channels)
- Exposes the  DM backend to the user, with or without noise

Other small features: 
- adds routine to create some random objects (random density matrices, random channels, ...)
- added tests and common checks. 
- Modified the randomness management of tests using a `setUp` method (as in https://docs.python.org/3/library/unittest.html) since tests are never ran as `__main__`with `pytest`

Everything is up for discussion.

**Related issue:**
https://github.com/TeamGraphix/graphix/issues/31


shinich1 (approved):
@mgarnier59 made quite a bit of tidy-ups and class/function renames for clarity. please check them and if you are happy, please merge them in to the feature branch.

shinich1:
@mgarnier59 please select `rebase and merge`. usually we do `squash and merge` but as this is a merge into feature branch (not master), I chose rebase.

Issue #99, opened 2023-11-15
Flow and Gflow Checker


masa10-f:
**Describe the feature you'd like**

A function that determines whether a given (g)flow is a correct (g)flow or not based on the definition.

**Additional context**

It can be useful for a graph-based generator and ZX-calculus-based backend, which will be introduced in the future.

pafloxy:
Hello! it seems a good idea :)  how do you plan to proceed about it ? 
I think it can be most conveniently checked by building two matrices one for the Gflow (say $\mathtt{F}$ ) another for the odd neighbourhood of Glfow (say $\mathtt{O} \ : \ (\mathtt{A}_G \  \mathtt{F}^T)^T$ , where $\mathtt{A}_G$ is the adjacency matrix )  and then simply ensuring that those matrices are lower-triangular. Depending on the measurement plane we might need to check the diagonal elements as well but that won't add up to any further complicacies. I can put in more details about the calculation involved if you want, let me know what you think ;) 
*PS:* The linear algebra must be in $GF2$                                                                             

Also what do you mean by a 'graph-based generator'  ?

pafloxy:
Hello! it seems a good idea :)  how do you plan to proceed about it ? 
I think it can be most conveniently checked by building two matrices one for the Gflow (say $\mathtt{F}$ ) and another for the odd neighbourhood of Glfow (say $\mathtt{O} \ : \ (\mathtt{A}_G \  \mathtt{F}^T)^T$ , where $\mathtt{A}_G$ is the adjacency matrix )  and then simply ensuring that those matrices are lower-triangular. Depending on the measurement plane we might need to check the diagonal elements as well but that won't add up to any further complicacies. I can put in more details about the calculation involved if you want, let me know what you think ;) 
*PS:* The linear algebra must be in $GF2$                                                                             

Also what do you mean by a 'graph-based generator'  ?

masa10-f:
Thanks for your comment and sorry for the late reply.
I understand your idea. I didn't think much about the implementation 
when I made this issue. Your proposal seems good, so I plan to follow it for implementation. Once I've pushed the changes, would you review my implementation?

> Also what do you mean by a 'graph-based generator' ?
The `graph-based generator` here indicates `graphix.generator.generate_from_graph` that generates a pattern from a graph with measurement planes.

pafloxy:
Sure ! I will be glad to help

PR #112, opened 2024-01-09
Flow and GFlow verifier

masa10-f:

Then, please fill in below:

**Context (if applicable):**

Create flow and gflow checker.

TODO in this PR
- [x] Create flow and gflow checker
- [x] Make unittest for them
- [x] Add type hint to the other methods in `gflow.py`

**Related issue:**

#99 


shinich1:
(optional) we could also test with patterns (graphs) transpiled from random circuits or after perform_pauli_measurments.

'opt' transpile with Rzz gate (phase gadget) shouldn't have a flow so might be a good 'gflow-but-not-flow' graph example to add?

masa10-f:
Thanks for your comment!

> (optional) we could also test with patterns (graphs) transpiled from random circuits or after perform_pauli_measurments.

I will add it to the flow check test. But we cannot know pauli-node measured graph always has no flow(Pauli measured graph sometimes has a flow)

> 'opt' transpile with Rzz gate (phase gadget) shouldn't have a flow so might be a good 'gflow-but-not-flow' graph example to add?

Yes, it's the case. But this is covered with the current 'gflow-but-not-flow' case. The extended Rzz gate pattern has only YZ plane measurement but the current test has all the measurement planes. But, unit graphs with each single measurement plane could be a good unit test, so I will consider it.

masa10-f:
It seems that the node order of matrix representation needs to be considered more carefully since the current `get_layers_from_flow` does not work well to make flow matrix or oddneighbor matrix upper triangular. Maybe dynamic reordering is necessary?

masa10-f:
Or it could be feasible to reconstruct an effective order from both (g)flow and Odd(f/g). I plan to explore this approach soon

masa10-f:
I've included odd neighbor into `get_layers_from_flow`. As a result, the flow checker test with random circuit now works correctly. However, the gflow checker test still fails. Investigating the test graph, I discovered that `gflow` returns an incorrect gflow, where gflow is not a partial order. This could be attributed to either `gflow` or `get_meas_planes`.
Therefore, I plan to investigate the bug in the gflow finding function tomorrow.

masa10-f:
gflow

In this graph, M(49) = YZ, M(61) = XZ and

```math
\begin{align*}
g(49) = (49, 28)\\
g(61) = (49)\\
Odd(g(49)) = (8, 61)\\
Odd(g(61)) = (61)
\end{align*}
```
So, it violates gflow condition.

pafloxy:
I see :/. In the picture does the index of the node respect the partial order $\prec$ imposed by the gflow ?

masa10-f:
> I see :/. In the picture does the index of the node respect the partial order imposed by the gflow ?

I think the answer would be no. This graph was generated from a random circuit and `perform_pauli_measurements`, and it inherited the node index from the original graph, which did not include Pauli measurements. In such cases, gflow is not generally consistent with the original index order.

pafloxy:
>I think the answer would be no. ...
Ahn okay I see. 

As much I understand the gflow is incorrect beacuse $g(v_{61})$ does not contain $v_{49}$ which should have been the case as per the rule for $\lamda = XZ$. Right ? 
Now adding a vertex to its own correction set is not dependent with the rest of the graph-structure I guess, no ? (I haven't delved deeper into the code yet , sorry :/ ). I mean,  my guess is the problem did not arise from the non-solvability of the the linear-equations involved, what do you say :?

masa10-f:
image

@pafloxy Sorry, I made a mistake. The measurement plane of node 61 is XY. Therefore, g and odd(g) are correct in the sense of measurement plane. However, according to the definition of gflow (g1) and (g2), we get 61 < 49 from $g(61) = (49)$, but also 49 < 61 from $Odd(g(49) = (8, 61)$. This is not a valid partial order.

pafloxy:
Ahn okay, I didn't have access to the $\prec$ in the diagram you provided so couldn't have guessed that. But if this is indeed the case then perhaps it has t do with the solver I believe. No ?

PS. : Btw I usually have theorems from this paper right on top of my head so for convenience in future discussions just put the theorem index and I'll take it from there :)

masa10-f:
> But if this is indeed the case then perhaps it has t do with the solver I believe. No ?

Yes, I guess that the cause lies in the solver. 
Sorry, I missed the generation code for generating the above graph. 

By changing depth and width of `get_rand_circuit`, we can probably find a same kind of bug.
```python
seed = 42
circ = get_rand_circuit(5, 5, seed=seed)
pattern = circ.transpile()
pattern.standardize()
pattern.shift_signals()
pattern.perform_pauli_measurements()
nodes, edges = pattern.get_graph()
graph = nx.Graph()
graph.add_nodes_from(nodes)
graph.add_edges_from(edges)
input = set()
output = set(pattern.output_nodes)
meas_planes = pattern.get_meas_plane()
g, l_k = gflow(graph, input, output, meas_planes)
```

pafloxy:
Sorry for replying late but I'll try to reproduce the same errors on my end.

I was going to ask you to add more algebra tools to the `linalg.py` module but I realised you added a few in this PR wrt `80-bug-gflow-finding ...` I was  wondering if it would be convenient to inherit methods from `galois.GF2` directly since it has lot os useful ones. 

For the incorrect gflow you mentioned above, was the `gflow.check_gflow()` able to recognise the error ? I'm asking to make sure that atleast there is no issue with the checker. I was going through the code of it and it seemed alright from up-front, tried it on some examples too and they seemed to work as well.

pafloxy:
@masa10-f So could you manage to find the errors in the implementation of `gflow.gflow()`

pafloxy:
Thanks a lot!

shinich1:
> Thanks a lot!
> 
> > I found the cause and just fixed. The error stems from the permutation inconsistency in MatGF2.forward_eliminate and related methods.
> 
> Where can I find the updated version? Did you commit it somewhere?

perhaps this commit? https://github.com/TeamGraphix/graphix/pull/112/commits/655c99f404d3ec3b559953a82e801c8c049815da

I will complete the review soon so this can be merged in, please add comments if you have any @pafloxy .

graphix/gflow.py
```diff
@@ -432,17 +456,109 @@ def get_layers(l_k):
...
+    flow: dict[int, set], odd_flow: dict[int, set], inputs: set[int], outputs: set[int]
+) -> tuple[dict[int, set], int]:
+    """Get layers from flow (incl. gflow).
+
+    Parameters
+    ----------
+    flow: dict[int, set]
+        flow function. flow[i] is the set of qubits to be corrected for the measurement of qubit i.
+    odd_flow: dict[int, set]
+        odd flow neighbors of each node.
```
shinich1, line 514: so it's different from odd neighbors right? I would recommend noting the definition (using math environment) to docstring.

graphix/gflow.py
```diff
@@ -458,3 +574,194 @@ def get_adjacency_matrix(graph):
     adjacency_matrix = nx.to_numpy_array(graph, nodelist=node_list)
     adjacency_matrix = MatGF2(adjacency_matrix.astype(int))
     return adjacency_matrix, node_list
+
+
+def check_flow(
```
shinich1, line 579: `verify_flow` might be clearer?

graphix/gflow.py
```diff
@@ -458,3 +574,194 @@ def get_adjacency_matrix(graph):
...
+    # Neighbor_f = MatGF2(Neighbor_f.data)
+    # check whether v < N(f(v))
+    triu = np.triu(Neighbor_f.data)
+    triu = MatGF2(triu)
+    dif = triu - Neighbor_f
+    valid_flow &= np.count_nonzero(dif.data) == 0
+    return valid_flow
+
+
+def check_gflow(
```
shinich1, line 652: `verify_gflow`?

graphix/gflow.py
```diff
@@ -458,3 +574,194 @@ def get_adjacency_matrix(graph):
...
+    Parameters
+    ----------
+    graph: nx.Graph
+        graph (incl. in and out)
+    input: set
+        set of node labels for input
+    output: set
+        set of node labels for output
+    gflow: dict[int, set]
+        gflow function. gflow[i] is the set of qubits to be corrected for the measurement of qubit i.
```
shinich1, line 670: more explicit definition of `gflow` here, based on original def might be useful?

masa10-f:
@pafloxy 
@shinich1 Thanks! I fixed in this commit
> perhaps this commit? https://github.com/TeamGraphix/graphix/commit/655c99f404d3ec3b559953a82e801c8c049815da

graphix/gflow.py
masa10-f: Sorry, it's just odd neighbors of (g)flow. Odd(f(node)) or Odd(g(node))

graphix/gflow.py
masa10-f: I agree

graphix/gflow.py
masa10-f: Do you mean [this one](https://github.com/TeamGraphix/graphix/pull/112#issuecomment-1899427717)? We can also place it inside gflow.gflow. Which do you think would be more useful

graphix/gflow.py
pafloxy: I agree with this convention, also what do you think about naming the main algorithm `gflow_finder()` istead of just `gflow` ? Since in essence it is lookinf for a gflow in the presented graph.

shinich1: thanks for the comment! Along your line of argument, I would suggest 'find_gflow'.

graphix/gflow.py
shinich1: maybe just add 
```python
.. seealso:: :func:`gflow.gflow`
```
 to the docstring, then it's minimally clear what needs to be supplied to the function.

shinich1 (approved):
Looks good to me! 

@mgarnier59 do you have any final comments?

shinich1:
@masa10-f talked with @mgarnier59 re this PR, please squash and merge!

Issue #109, opened 2023-12-19
Visualization from pattern


shinich1:
**Is your feature request related to a problem? Please describe.**
Current visualization tool is based on the flow or gflow obtained from the `graphix.gflow.flow` or `graphix.gflow.gflow` functions, which may have different feedforward structure from the pattern itself.

**Describe the feature you'd like**
Add an option to visualize the feedforward structure according to pattern, not from flow or gflow. This can be done by obtaining layers from `Pattern.get_layers` and then constructing the arrows according to the feedforward part of measurement commands. Perhaps only consider either s or t feedforward for arrows as there may be too many of them.

**Additional context**
- also, it might be useful to change arrow type when nodes are not connected, as mentioned elsewhere by @d1ssk

masa10-f:
> Perhaps only consider either s or t feedforward for arrows as there may be too many of them.

Comment on this part, s(X correction) is equal to g and t(Z correcion) is equal to Oddneighbor(g), so, in this context, arrows should be s signals(see thm. 2 in [Generalized Flow and Determinism in Measurement-based Quantum Computation](https://arxiv.org/pdf/quant-ph/0702212.pdf))

We can construct a function like `get_gflow_from_pattern` that collects all s signals or X corrections from pattern.

PR #113, opened 2024-01-15
Gflow from pattern

d1ssk:

Then, please fill in below:

**Context (if applicable):**
Current gflow visualization method is based on gflow obtained from `graphix.gflow.gflow`, which is not always consistent with the pattern. #109

Add a method to obtain gflow which is consistent with the feedforward structure of the patterns. Integrate this to the visualization method which shows gflow structure based on the pattern with the option `pattern.draw_graph(gflow_from_pattern=True)`
Also, add an option `meas_plane_indicator` to the visualization. If True, measurement plane for each node is showed adjacent to the node.

**Related issue:**
#109

### DIFF
--- a/graphix/gflow.py
+++ b/graphix/gflow.py
@@ -19,7 +19,7 @@ import sympy as sp
 from graphix.linalg import MatGF2
 
 
-def gflow(graph, input, output, meas_planes, mode="single"):
+def gflow(graph, input, output, meas_planes, mode="single", pattern=None):
     """Maximally delayed gflow finding algorithm
 
     For open graph g with input, output, and measurement planes, this returns maximally delayed gflow.
@@ -47,11 +47,14 @@ def gflow(graph, input, output, meas_planes, mode="single"):
     mode: str(optional)
         The gflow finding algorithm can yield multiple equivalent solutions. So there are three options
             - "single": Returrns a single solution
+            - "pattern": Returns a single solution consistent with a given pattern
             - "all": Returns all possible solutions
             - "abstract": Returns an abstract solution. Uncertainty is represented with sympy.Symbol objects,
               requiring user substitution to get a concrete answer.
 
         Default is "single".
+    pattern: graphix.Pattern object(optional)
+        pattern to be based on. This is used only when mode is "pattern".
 
     Returns
     -------
@@ -60,6 +63,10 @@ def gflow(graph, input, output, meas_planes, mode="single"):
     l_k: dict
         layers obtained by gflow algorithm. l_k[d] is a node set of depth d.
     """
+    if mode == "pattern":
+        if pattern is None:
+            raise ValueError("pattern must be specified when mode is pattern")
+        g, l_k = gflow_from_pattern(pattern)
     l_k = dict()
     g = dict()
     for node in graph.nodes:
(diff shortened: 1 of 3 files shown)


shinich1:
Nice, would you be happy to add type hints to `visualization.py`?

graphix/gflow.py
```diff
@@ -19,7 +19,7 @@
 from graphix.linalg import MatGF2
 
 
-def gflow(graph, input, output, meas_planes, mode="single"):
+def gflow(graph, input, output, meas_planes, mode="single", pattern=None):
```
shinich1, line 22: I don't think it's a good idea to have pattern as optional argument in `gflow`. having `gflow_from_pattern` is enough. can we get rid of added codes to `gflow`?

graphix/gflow.py
```diff
@@ -207,6 +214,55 @@ def gflowaux(
...
+                    g[g_out] = {g_in}
+        if cmd[0] == "X":
+            g_in = cmd[1]
+            g_outs = set(cmd[2]) & nodes
+            for g_out in g_outs:
+                if g_out in g.keys():
+                    g[g_out] = g[g_out].union({g_in})
+                else:
+                    g[g_out] = {g_in}
+
```
shinich1, line 262: Could you add gflow check (based on #112) at the end of this function?

graphix/pattern.py
```diff
@@ -1234,9 +1234,11 @@ def draw_graph(
         self,
         node_distance=(1, 1),
         figsize=None,
+        gflow_from_pattern=False,
```
shinich1, line 1237: perhaps `gflow_from_pattern` should be defaulted to True

graphix/gflow.py
d1ssk: This was difinitely unnecessary. I deleted it. Thank you.

graphix/pattern.py
d1ssk: I set the default to True.

shinich1:
@d1ssk let's merge this once the additional error reported in #107 is fixed, as #112 may take some time to complete. we'll add the gflow check later.

d1ssk:
Understood. I have already corrected the additional error and I'm ready for merging this after your final review.

graphix/pattern.py
```diff
@@ -1234,9 +1234,11 @@ def draw_graph(
         self,
         node_distance=(1, 1),
         figsize=None,
+        gflow_from_pattern=True,
         pauli_indicator=True,
         show_loop=True,
         local_clifford_indicator=False,
+        meas_plane_indicator=False,
```
shinich1, line 1241: change to show_measurement_planes

graphix/visualization.py
```diff
@@ -344,7 +409,15 @@ def visualize_w_gflow(
             plt.savefig(filename)
         plt.show()
 
-    def visualize_wo_structure(self, angles=None, local_clifford=None, node_distance=(1, 1), save=False, filename=None):
+    def visualize_wo_structure(
```
shinich1, line 412: Is this currently showing entirely by random node placements? Is it possible to at least arrange nodes by their preparation or measurement order, perhaps (no need to be very tidy as there's no structure in this case

shinich1 (approved):
Good work! Let's merge once the additional comments are resolved.

graphix/visualization.py
d1ssk: I will address this in a separate issue.

shinich1:
@d1ssk lets squash and merge.

Issue #92, opened 2023-11-10
Contraction-Wise State Vector Simulator


masa10-f:
**Is your feature request related to a problem? Please describe.**

The state vector simulator for MBQC inherently exhibits slower performance compared to that for quantum circuits, primarily attributed to two key factors:

1. MBQC necessitates the utilization of at least one ancillary qubit to proceed computation. This leads to at least a twofold increase in memory consumption, consequently extending the computational time.
2. In contrast to a gate in a quantum circuit, a single step in MBQC cannot encompass all single-qubit Unitary operations without changing the graph structure. This characteristic augments the size of the computational graph, demanding a higher computational load.

In order to address the aforementioned challenges, Graphix incorporates a `minimize_space` method designed to mitigate the impact of ancillary qubits on computational efficiency. This method aims to minimize the number of ancillary qubits required simultaneously, enabling Graphix to simulate large-scale MBQC with a comparatively lower computational burden.

However, it is important to note that this approach is constrained to simulating measurement patterns in standardized or NEMC forms, which are prevalent in MBQC. Additionally, it's crucial to acknowledge that the current implementation of the `minimize_space` method is heuristic. Consequently, it does not guarantee optimal outcomes in all cases, as it may not consistently yield the most efficient solutions.

**Describe the feature you'd like**

I propose the implementation of a new simulator backend, the Contraction-wise State Vector Simulator(CW-SVS), into Graphix. The CW-SVS optimizes contraction order by delaying the calculation of outer products until the latest possible stage, leading to a substantial reduction in FLOPs. Importantly, this simulator maintains the same computational capabilities as conventional state vector simulators, as both calculate the same tensor network. The difference is just in contraction order.

The CW-SVS's intelligent contraction order incorporates the `minimize_space` feature of the `Pattern` class. This eliminates the need to explicitly consider the order of measurement patterns, enabling the seamless simulation of standardized measurement patterns. We can implement this backend with `quimb` and slightly modify the current `tensornetwork` backend.

Furthermore, applying this contraction-wise strategy could extend beyond the state vector simulator. It can be leveraged to enhance density matrix backends and circuit simulators, contributing to a more efficient and versatile computational framework.

**Additional context**

na

shinich1:
Nice!

A minor point but what's the reasoning for the name 'contraction-wise simulation'? this '-wise' sounds a bit like doing simulation for every contraction?

Perhaps something like:
- tensor-based state vector simulation 
- path-flexible state vector simulation
[...]

PR #114, opened 2024-01-16
92 efficient contraction order statevector simulator

nabe98:

Then, please fill in below:

**Context (if applicable):**

add TN contract based state vector simulator

**Related issue:**

#92 


shinich1:
@nabe98 do we have benchmark plot for this backend?

shinich1:
@nabe98 nice! you can just paste on a comment here. where did you plan to add it in the next commit?

graphix/sim/tensornet.py
```diff
@@ -13,7 +18,7 @@ class TensorNetworkBackend:
     Executes the measurement pattern using TN expression of graph states.
     """
 
-    def __init__(self, pattern, graph_prep="auto", **kwargs):
+    def __init__(self, pattern, backend="tensornetwork", graph_prep="auto", **kwargs):
```
masa10-f, line 21: Why did you add the `backend` argument? it seems not called anywhere

masa10-f:
I will remove `graph_prep` argument and `mps` keyword later becuase the former is not critical point in tensor network calculation and the latter is not properly implemneted in Graphix.

graphix/sim/tensornet.py
```diff
@@ -488,28 +522,32 @@ def get_basis_amplitude(self, basis, **kwagrs):
...
+            tn = self.copy()
+            output_inds = [self._dangling[str(index)] for index in self.default_output_nodes]
+            psi = tn.contract(output_inds=output_inds, **kwagrs).data.reshape(-1)
+            return Statevec(psi=psi)
+
+        tn = self.copy()
+        tn_simplified = tn.full_simplify("ADCR")
+        output_inds = [self._dangling[str(index)] for index in self.default_output_nodes]
+        psi = tn_simplified.contract(output_inds=output_inds, **kwagrs)
+        return psi if path_info else psi.data.reshape(-1)
```
masa10-f, line 550: `path_info` is the flag to get [opt_einsum.contract.PathInfo](https://optimized-einsum.readthedocs.io/en/stable/autosummary/opt_einsum.contract.PathInfo.html)-like object without contraction. Altough I originally implemented it, in my reference code, to evaluate the contraction cost for the case it is impossible to explicitly calculate, this argument may be confusing for users. So, you can remove it

graphix/sim/tensornet.py
nabe98: I forgot to remove this and deleted it.

Issue #66, opened 2023-06-26
PyZX as a pattern optimization backend?


shinich1:
**Is your feature request related to a problem? Please describe.**
na

**Describe the feature you'd like**
Implement pyzx interface to use ZX-calculus as a pattern optimization backend. This means we will need to construct graph-like diagram, optimize with pyZX, and then get pattern out of it to be simulated or executed on device.

**Additional context**
https://quantum-journal.org/papers/q-2021-03-25-421/ 
https://arxiv.org/pdf/2304.08166.pdf
etc

king-p3nguin:
Note that PyZX is now updating the function `to_graph_like`. The current version's `to_graph_like` function does not work, so we may need to wait for the next release (or implement it ourselves). 
https://github.com/Quantomatic/pyzx/pull/189

Getting a pattern out of the PyZX diagram can be done by `graphix.generator.generate_from_graph()`.

PR #116, opened 2024-01-19
ZX based MBQC

masa10-f:

Then, please fill in below:

- Create ZX Calculus based MBQC representation class(mgraph)
  - [ ] Support ZX rewrite rules
  - [ ] Interface to pyzx
  - [ ] interface with simulator
  - [ ] transpiler to measurement pattern
- [x] Create transpiler from circuit to mgraph(now limited to "XY" planes)
- [ ] support transpiler with extended measurement planes(maybe same as the 'opt' option of the circuit-pattern transpiler)

**Related issue:**

#66 


masa10-f:
We will implement a new graph state class that can be converted between ZX diagram. So this PR is no longer necessary

PR #121, opened 2024-02-28
Safer pattern nodes

thierry-martinez:
This pull request ensures that `Pattern` class preserves this internal invariant: `Pattern.output_nodes` is now ensured to be equal to `Pattern.input_nodes` plus nodes introduced by `N` commands minus nodes removed by `M` commands.

`Pattern.seq` is renamed into a private field `Pattern.__seq` and `Pattern.Nnode` is now a read-only property.  `Pattern` constructor now only takes an optional list of `input_nodes`, and can only be updated via `add` and `extend`. `Pattern` are now iterable and `len` is now defined for patterns: we should write `for command in pattern:` instead of `for command in pattern.seq:` and `len(pattern)` instead of `len(pattern.seq)`.  `N` commands are no longer added by `Pattern` constructor and should be added explicitly after the instantiation.

### DIFF
--- a/graphix/pattern.py
+++ b/graphix/pattern.py
@@ -1084,7 +1084,7 @@ class Pattern:
         Returns
         -------
         node_list : list
-            list of nodes that are entangled with speicifed node
+            list of nodes that are entangled with specified node
         """
         if not self.is_standard():
             self.standardize()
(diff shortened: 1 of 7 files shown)


graphix/pattern.py
```diff
@@ -93,34 +90,98 @@ def add(self, cmd):
         assert type(cmd) == list
         assert cmd[0] in ["N", "E", "M", "X", "Z", "S", "C"]
         if cmd[0] == "N":
-            self.Nnode += 1
-            self.output_nodes.append(cmd[1])
+            if cmd[1] not in self.__output_nodes:
```
thierry-martinez, line 93: For the time being, we accept to reprepare with an `N` command a node that is already prepared (either by a previous `N` command or by being listed in `input_nodes`): it seems preferable to fail in this case, but this breaks other tests. We should discuss whether we address this question here or for a future pull request. Ideally, we would prefer that `input_nodes` are prepared by the context (in our case, the simulator) and not by `N` commands.

thierry-martinez: Fixed in b7985a4

shinich1 (approved):
@thierry-martinez could you squash and merge (you may need to merge graphix:master to this branch first as there's new push)? I have a few minor changes I want to make (e.g. removing commented out codes, updating the examples where necessary) but these will be looked at after this is merged.

thierry-martinez:
Merged. Thanks!

PR #122, opened 2024-03-05
Refactor measure operator in a new pauli module

thierry-martinez:
This pull-request introduces a `pauli` module for symbolic Pauli calculations, which are used in `statevec` instead of direct matrix computation. The `test_pauli` module checks that the computations are the same as before, modulo a bug fix: ` CLIFFORD[CLIFFORD_CONJ[vop]]` was used instead of `CLIFFORD[vop].conj().T`, despite the two are only equal modulo phase in the general case (see the comment in `pauli.py`). The new implementation allows the angle and the plane to be computed separately from the state vector back-end, which will be useful for future client/server computations and to share the code with other back-ends.

### DIFF
--- a/graphix/pauli.py
+++ b/graphix/pauli.py
@@ -254,3 +254,32 @@ def parse(name: str) -> Pauli:
     Return the Pauli gate with the given name (limited to "I", "X", "Y" and "Z").
     """
     return get(IXYZ[name], UNIT)
+
+
+class MeasureUpdate(pydantic.BaseModel):
+    new_plane: Plane
+    coeff: int
+    add_term: float
+
+    @staticmethod
+    def compute(plane: Plane, s: bool, t: bool, clifford: "graphix.clifford.Clifford") -> "MeasureUpdate":
+        gates = list(map(Pauli.from_axis, plane.axes))
+        if s:
+            clifford = graphix.clifford.X @ clifford
+        if t:
+            clifford = graphix.clifford.Z @ clifford
+        gates = list(map(clifford.measure, gates))
+        new_plane = Plane.from_axes(*(gate.axis for gate in gates))
+        cos_pauli = clifford.measure(Pauli.from_axis(plane.cos))
+        sin_pauli = clifford.measure(Pauli.from_axis(plane.sin))
+        exchange = cos_pauli.axis != new_plane.cos
+        if exchange == (cos_pauli.unit.sign == sin_pauli.unit.sign):
+            coeff = -1
+        else:
+            coeff = 1
+        add_term = 0
+        if cos_pauli.unit.sign:
+            add_term += np.pi
+        if exchange:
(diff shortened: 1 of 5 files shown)


graphix/pauli.py
```diff
@@ -0,0 +1,285 @@
...
+        match a, b:
+            case Axis.X, Axis.Y:
+                return Plane.XY
+            case Axis.Y, Axis.Z:
+                return Plane.YZ
+            case Axis.X, Axis.Z:
+                return Plane.XZ
+        assert a == b
+        raise ValueError(f"Cannot make a plane giving the same axis {a} twice.")
+
```
shinich1, line 162: comment from @mgarnier59 : this convention is consistent with Backens et al., 
to @masa10-f : I think this is much clearer than what we had before- any comments?

thierry-martinez:
Do we still want to support Python <3.10, or may we bump version constraint to allow `match` construct?

graphix/pauli.py
masa10-f: It seems nice. No comment so far :)

thierry-martinez:
> Are there significant differences? if we can work around with a few lines of additional code, my personal opinion is it's good to keep supporting 3.9. What's your thoughts?

Python 3.10 introduced `match`-blocks, allowing branching code to be written more elegantly by pattern-matching instead of `if-elif-else` chains. I rewrote in d7f7683 the few `match`-blocks I introduced into `if-elif-else` chains to be compatible with Python <3.10.

Python 3.9 allowed list and tuple types to be written as `list[...]` and `tuple[...]` instead of `typing.List[...]` and `typing.Tuple[...]`. I rewrote the types in 386f718 to be compatible with Python <3.9.

I think that when we will be ready to move to Python 3.10, `match`-blocks will allow many code places to be written in a more readable way, but there is no urgency to do so.

shinich1:
> > Are there significant differences? if we can work around with a few lines of additional code, my personal opinion is it's good to keep supporting 3.9. What's your thoughts?
> 
> 
> 
> Python 3.10 introduced `match`-blocks, allowing branching code to be written more elegantly by pattern-matching instead of `if-elif-else` chains. I rewrote in d7f7683 the few `match`-blocks I introduced into `if-elif-else` chains to be compatible with Python <3.10.
> 
> 
> 
> Python 3.9 allowed list and tuple types to be written as `list[...]` and `tuple[...]` instead of `typing.List[...]` and `typing.Tuple[...]`. I rewrote the types in 386f718 to be compatible with Python <3.9.
> 
> 
> 
> I think that when we will be ready to move to Python 3.10, `match`-blocks will allow many code places to be written in a more readable way, but there is no urgency to do so.

Thanks, that seems pretty good, let me think a bit.  @masa10-f: let's discuss this next week.

shinich1 (approved):
Looks good to me! I will add docstring and docs page, once this PR is merged.

PR #125, opened 2024-03-27
Allow empty output nodes and measures at circuit level

thierry-martinez:
This pull request enables measures to be expressed at circuit level, in order to be able to consider circuits with classical outputs. Patterns are now allowed to measure all their nodes, and have an empty output node set, to enable to consider patterns where only classical outputs are considered.

Transpiled circuits can now have "measure" gates, introduced with the `circ.m(qubit, plane, angle)` method. The measured qubit cannot be used in any subsequent gate. The transpiler now returns a `TranspileResult` dataclass: the pattern is available in the `pattern` field, and the field `classical_outputs` contains the index where the classical measures can be found in the `results` array of the simulator.

### DIFF
--- a/tests/test_tnsim.py
+++ b/tests/test_tnsim.py
@@ -52,7 +52,7 @@ class TestTN(unittest.TestCase):
     def test_entangle_nodes(self):
         random_vec = np.array([1.0, 1.0, 1.0, 1.0]).reshape(2, 2)
         circuit = Circuit(2)
-        pattern = circuit.transpile()
+        pattern = circuit.transpile().pattern
         pattern.add(["E", (0, 1)])
         tn = pattern.simulate_pattern(backend="tensornetwork", graph_prep="sequential")
         dummy_index = [gen_str() for _ in range(2)]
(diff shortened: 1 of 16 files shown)


tests/test_tnsim.py
shinich1, line 1: should we also test the `measure` feature for TN and DM backends?

shinich1 (commented):
Could you also add appropriate lines to docs (e.g. `generator.rst`) so they show up in module references? also it might be good to illustrate the use of this feature in examples - perhaps simple example like `rotation.py`?

shinich1:
example and docs bits can be done on my side as well, I will approve as soon as there's test for TN backend.

tests/test_tnsim.py
thierry-martinez: Done in aee982d. Thanks!

thierry-martinez:
> Could you also add appropriate lines to docs (e.g. `generator.rst`) so they show up in module references? also it might be good to illustrate the use of this feature in examples - perhaps simple example like `rotation.py`?

I tried to improve the documentation in 1e97cdc. Thanks!

shinich1 (approved):
@thierry-martinez Great! please squash and merge.

thierry-martinez:
Thanks! Merged.

PR #135, opened 2024-05-03
Allowing arbitrary initialization of the input nodes

mgarnier59:
Allowing arbitrary intialization of the input nodes for statevector and density matrix backends + tests.

Several format of initialization are supported (Statevec or DensityMatrix objects, list of newly-defined State objects, numerical data).
Main change is that basic states (|0>, |+>, ...) are now in states.BasicStates and no longer in ops.py.

PR #141, opened 2024-05-07
Setup Ruff

### LINKED ISSUE #140: Setup Ruff

Use Ruff https://docs.astral.sh/ruff/ to detect errors.

EarlMilktea:

Setup Ruff linter CI.

**Related issue:**

Resolves #140


shinich1:
@EarlMilktea let us merge this after #135

graphix/visualization.py
```diff
@@ -1014,8 +1014,8 @@ def get_pos_from_flow(self, f: dict[int, int], l_k: dict[int, int]) -> dict[int,
         for i, k in enumerate(start_nodes):
             pos[k][1] = i
             while k in f.keys():
-                k = list(f[k])[0]
-                pos[k][1] = i
+                kk = next(iter(f[k]))
+                pos[kk][1] = i
```
thierry-martinez, line 1018: The purpose of this `while` loop is to make `k` follow the path given by `f`: it is essential for `k` to be updated during the loop, or else the loop never terminates (because `k` will never change). If the purpose of this change is to prevent updating a `for`-loop variable, a solution could be:
```python
            node = k
            while node in f.keys():
                node = next(iter(f[node]))
                pos[node][1] = i
```

thierry-martinez: (Note: we are missing `test_visualization.py` to catch that!)

tests/test_linalg.py
```diff
@@ -176,5 +176,5 @@ def test_backward_substitute(self, test_case: LinalgTestCase) -> None:
         mat_eliminated, rhs_eliminated, _, _ = mat.forward_eliminate(rhs_input)
         x, kernel = mat_eliminated.backward_substitute(rhs_eliminated)
         if x is not None:
-            assert np.all(x == x)
+            assert np.all(x == x)  # noqa: PLR0124 (necessary)
```
thierry-martinez, line 179: Could we replace that with: `assert not np.any(np.isnan(x))`, which looks clearer?

thierry-martinez (changes requested):
Good work, thanks! I noticed only one bug introduced in `graphix/visualization.py` that should be addressed. Other comments are minor remarks.

EarlMilktea:
> ~@EarlMilktea let us merge this after #135~

@thierry-martinez @shinich1 is not planning to merge this PR immediately.
Could you wait a while until everything is ready?

graphix/visualization.py
EarlMilktea: > the purpose of this change is to prevent updating a for-loop variable

This is exactly what I meant.
Anyway we need to cover these lines...

tests/test_linalg.py
EarlMilktea: I'm afraid not.
As far as I tested, it leads to `TypeError` because `x` can sometimes be `object`-valued and thus we cannot cast the elements into `float`.

graphix/graphsim/basegraphstate.py
```diff
@@ -501,11 +501,10 @@ def equivalent_fill_node(self, node: int) -> int:
                 self.equivalent_graph_E1(i)  # this gives loop to node.
                 self.equivalent_graph_E1(node)
                 return 0
+        elif len(list(self.neighbors(node))) == 0:
```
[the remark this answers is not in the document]
thierry-martinez: `self.neighbors(node)` returns an iterator: converting it into a list for computing its length costs *O*(n) just to check whether the iterator is empty, whereas `all` returns `False` as soon as the iterator returns one element, therefore it is *O*(1).

graphix/visualization.py
thierry-martinez: I propose a fix and a test here: https://github.com/thierry-martinez/graphix/commit/d761465b71995674193920dcbd0332fb0b4ee57d

tests/test_linalg.py
shinich1: @thierry-martinez yes, perhaps good to leave this part untouched, and think about this when `sympy` will be removed from dependency and `gflow` module refactored.

EarlMilktea:
Ruff config. is partly merged by #135 .

@shinich1 

How about migrating from black + isort to Ruff format?

Ruff provides almost the same features and is faster. Additionally we can simplify `pyproject.toml` .

MEMO: Not planning to merge this PR. I'm planning to abandon it.

Issue #130, opened 2024-04-17
[Bug]: Some tests are broken if seed is changed


thierry-martinez:
**Bug description**
The following tests are broken when tested with a different seed:
* `test_pattern.py::test_pauli_measurment_opt_gate`
* `test_pattern.py::test_pauli_measurment_opt_gate_transpiler`
* `test_pattern.py::test_pauli_measurment_opt_gate_transpiler_without_signalshift`
* `test_gflow.py::test_rand_circ_gflow`

**To Reproduce**
The failures in `test_pattern.py` can be observed by adding
```diff
+        rc.set_seed(25)
```
in the beginning of the method `test_pauli_measurment_opt_gate` (changing the `SEED` value in the beginning of the file will not change the behavior, since the seed value is overridden by other tests).

The failure in `test_gflow.py` can be observed by changing the value of `seed` in the beginning of the file.
```diff
-seed = 30
+seed = 25
```

**Expected behavior**
Changing the seed should not break the tests.

**Environment (please complete the following information):**
 - OS:  macOS 14.4.1 (23E224)
 - Python version: 3.11.8
 - Related module versions if applicable: numpy=1.25.2

**Additional context**
I noticed these failures in an effort to address https://github.com/TeamGraphix/graphix/issues/65

EarlMilktea:
For `test_pauli_measurment_opt_gate*` tests, I also encountered it.

masa10-f:
@shinich1 
Have you checked the `test_gflow.py::test_rand_circ_gflow` ?

shinich1:
> @shinich1 Have you checked the `test_gflow.py::test_rand_circ_gflow` ?

I didn't have any problem on my side (I removed the skip command when trying). Somehow I haven't pushed it but we need to remove this line.
https://github.com/TeamGraphix/graphix/blob/19eb94cabccf0ca9f9ad2b765a2f2bfbf65bba93/tests/test_gflow.py#L510

PR #145, opened 2024-05-08
patch: preserve output node order in measure_pauli

shinich1:
- Format added code by `black` and `isort`
  - See `pyproject.toml` for configurations

Then, please fill in below:
**Context (if applicable):**
#130

It seems like the order of `output_nodes` is not preserved in `pat.replace` in `measure_pauli`, modified in #121.
https://github.com/TeamGraphix/graphix/blob/e9ccc1590be328c81e4afb216536906bb85e403f/graphix/pattern.py#L1948

commit: https://github.com/TeamGraphix/graphix/commit/981eff8f5b310b53b7497bdc4ff5704f7ec69c92

Now the test passes with SEED=25.

**Related issue:**

### DIFF
--- a/graphix/pattern.py
+++ b/graphix/pattern.py
@@ -1945,7 +1945,9 @@ def measure_pauli(pattern, leave_input, copy=False, use_rustworkx=False):
     else:
         pat = pattern
 
+    output_nodes = deepcopy(pattern.output_nodes)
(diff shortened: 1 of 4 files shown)


thierry-martinez:
Good catch! Thank you, @shinich1 !

PR #151, opened 2024-05-17
Use codecov just for information

EarlMilktea:
Now codecov action passes regardless of code coverage changes.


EarlMilktea:
MEMO: https://docs.codecov.com/docs/quick-start#tips-and-tricks

PR #158, opened 2024-05-30
Fix #45: Parameterized measurement patterns

### LINKED ISSUE #45: Parameterized measurement patterns

thierry-martinez:
This commit is yet another tentative to implement parameterized measurement patterns, to fulfill issue #45 (previous tentative: #68).

This commit adds two methods to the class `Pattern`:

- `is_parameterized()` returns True if there is at least one measurement angle that is not just an instance of numbers.Number: indeed, a parameterized pattern is a pattern where at least one measurement angle is an expression that is not a number, typically an instance of `sympy.Expr` (but we don't force to choose sympy here).

- `subs(variable, substitute)` returns a copy of the pattern where occurrences of the given variable in measurement angles are substituted by the given value.  Substitution is performed by calling the method `subs` on measurement angles, if the method exists, which is the case in particular for `sympy.Expr`. If the substitution returns a number, this number is coerced to `float`, to get numbers that implement the full number protocol (in particular, sympy numbers don't implement `cos`).


shinich1:
@thierry-martinez great, is the intention to merge @pafloxy's #68-equivalent code first (when it's ready), before merging this?

thierry-martinez:
> @thierry-martinez great, is the intention to merge @pafloxy's #68-equivalent code first (when it's ready), before merging this?

I adapted #68 code for `Parameter` class in the commit 81bbeb7 I just pushed. #68 was incorrect because arithmetic operators changed parameters in place (computing `alpha + 1` did change the value of `alpha` to `alpha + 1`!). In this PR, we distinguish between `Parameter` (`alpha`) and `ParameterExpression` (`alpha + 1`), and a `Parameter` is a particular case of `ParameterExpression`.

The test `test_parameter_simulation` is illustrative: we can either substitute a parameter by a value in a pattern then simulate, or simulate symbolically (given that the pattern is deterministic!) then substitute in the (symbolic) state vector or the density matrix, and we get the same result.

pafloxy (approved):
I think as for now this is all that I was looking for, If @shinich1 thinks its alright maybe we should go for the merge. :)

shinich1:
@thierry-martinez @pafloxy great! give us a few days to look through. A few quick comments:

1. is TN backend working with parameters, or should we throw error in TN backend if pattern is parametrised?
2. let us add sympy to `requirements.txt`.
3. will `circuit.simulate_statevector`, `pattern.perform_pauli_measurements`, as well as visualization tool, work with parameters?

EarlMilktea:
Let me review it later.

Anyway, is it absolutely necessary to use sympy?
I'm concerned with...

- performance: Personally I feel sympy is far from performant.
- type stability: sympy has no type annotations, and it seems that not even planned.
- developer experience: Sometimes sympy makes linter (ex. flake8, Ruff) extremely slow.

Additionally, let me point out that @masa10-f is planning to completely eliminate sympy from the source.

EarlMilktea:
If sympy is absolutely necessary, it's OK, but we may need to verify that we're using it wisely.

graphix/parameter.py
```diff
@@ -0,0 +1,213 @@
...
+
+class ParameterExpression:
+    """Expression with parameters.
+
+    Implements arithmetic operations. This is essentially a wrapper over
+    sp.Expr, exposing methods like cos, conjugate, etc., that are
+    expected by the simulator back-ends.
+    """
+
+    def __init__(self, expression: sp.Expr):
```
EarlMilktea, line 25: Please do not omit `-> None`.

graphix/parameter.py
```diff
@@ -0,0 +1,213 @@
...
+        return ParameterExpression(sp.sqrt(self._expression))
+
+    @property
+    def expression(self) -> sp.Expr:
+        return self._expression
+
+    def __repr__(self) -> str:
+        return str(self._expression)
+
+    def subs(self, variable, value) -> ParameterExpression | numbers.Number:
```
EarlMilktea, line 141: Please use `complex` instead of abstract `numbers.Number` .

graphix/parameter.py
```diff
@@ -0,0 +1,213 @@
...
+        # simulate without pattern assignment
+        # (the resulting state vector is symbolic)
+        # Note: pr_calc=False is mandatory since we cannot compute probabilities on
+        # symbolic states; we explore one arbitrary branch.
+        sv2 = pattern.simulate_pattern(pr_calc=False, rng=np.random.default_rng(seed))
+        # Substituting alpha in the resulting state vector should yield the same result
+        assert np.allclose(sv.psi, sv2.subs(alpha, 0.5).psi)
+    """
+
+    def __init__(self, name):
```
EarlMilktea, line 179: Add `-> None` .

graphix/parameter.py
```diff
@@ -0,0 +1,213 @@
...
+        """Create a new :class:`Parameter` object.
+
+        Parameters
+        ----------
+        name : str
+            name of the parameter, used for binding values.
+        """
+        assert isinstance(name, str)
```
EarlMilktea, line 187: Please do not use `assert` if you're planning to expose this class.

graphix/parameter.py
```diff
@@ -0,0 +1,213 @@
...
+        self._name = name
+        super().__init__(sp.Symbol(name=name))
+
+    @property
+    def name(self):
```
EarlMilktea, line 192: Please explicitly specify the return type.

graphix/sim/base_backend.py
```diff
@@ -21,19 +25,19 @@ def perform_measure(
...
-        result = np.random.choice([0, 1])
+        result = rng.choice([0, 1])
         op_mat = op_mat_from_result(vec, result)
     state.evolve_single(op_mat, qubit)
     return result
 
 
 class Backend:
-    def __init__(self, pr_calc: bool = True):
+    def __init__(self, pr_calc: bool = True, rng: np.random.Generator = None):
```
EarlMilktea, line 40: Please use `np.random.Generator | None`.

graphix/sim/density_matrix.py
```diff
@@ -279,11 +282,21 @@ def apply_channel(self, channel: KrausChannel, qargs):
...
+        result = DensityMatrix(nqubit=self.Nqubit)
+        result.rho = np.vectorize(lambda value: graphix.parameter.subs(value, variable, substitute))(self.rho)
+        return result
+
 
 class DensityMatrixBackend(graphix.sim.base_backend.Backend):
     """MBQC simulator with density matrix method."""
 
-    def __init__(self, pattern, max_qubit_num=12, pr_calc=True):
+    def __init__(self, pattern, max_qubit_num=12, pr_calc=True, rng: np.random.Generator = None):
```
EarlMilktea, line 299: Use `| None` .

EarlMilktea (changes requested):

graphix/parameter.py
thierry-martinez: Numbers from `numpy` (or `sympy`...) are not necessarily instances of `complex`, for instance:
```
isinstance(np.array([1])[0], complex) == False
isinstance(np.array([1])[0], numbers.Number) == True
isinstance(sp.parse_expr(1), complex) == False
isinstance(sp.parse_expr(1), numbers.Number) == True
```

EarlMilktea: So do you mean this block can sometimes return non-`complex` value?

```python
if isinstance(result, numbers.Number):
    return complex(result)
```

EarlMilktea: If you're sure that `complex(result)` is always `complex` type, please use `ParameterExpression | complex`.

I personally believe that return type annotation should be as detailed as possible, or we need extra type narrowings.

thierry-martinez: Oops, sorry, I answered without thinking enough: you are right, and after the coercion, the result is `complex` indeed.

pafloxy:
> @thierry-martinez @pafloxy great! give us a few days to look through. A few quick comments:
> 
> 1. is TN backend working with parameters, or should we throw error in TN backend if pattern is parametrised?
>    ~2. could you add sympy to `requirements.txt`?~
> 2. will `circuit.simulate_statevector`, `pattern.perform_pauli_measurements`, as well as visualization tool, work with parameters?

@shinich1 @thierry-martinez 
Maybe you guys figured it already but I think we need to make some small modification to the `visualization.py` module as well.  There are some checks to determine if a measurement angle is Pauli in some of the class methods for `GraphVisualizer` of the form

```
elif (
                show_pauli_measurement
                and self.meas_angles is not None
                and (
                    2 * self.meas_angles[node] == int(2 * self.meas_angles[node])
                )  # measurement angle is integer or half-integer
            ):
```
where we might need to add another line or so to exclude the case where the measurement angles is an instance of `Parameter`/`ParameterExpression`.

shinich1:
Hi @pafloxy @thierry-martinez 

I think the best way is to move parameter implementation with `sympy` out of main `graphix` repository, for example a separate repo (wrapper/module) dedicated to parameterized patterns and their executions (it seems possible, given the relatively small changes required to implement?). I am happy to initiate a creation of such repository in `teamgraphix` organization, if that makes sense to you? for example, that can be part of `extra` pip installation.

The reason mostly follows @EarlMilktea 's. Sustaining maintainability is going to help a lot in the long run for everyone contributing to this repo and will help a lot those maintaining.

1. `sympy` has quite a performance issue. Looking at their github repo, it seems like a recurring issue in `qiskit` (which has `sympy` incorporated) and quite a bit of effort can be saved by deciding to move it out at this point (nb as @EarlMilktea mentioned `sympy` will be removed from `requirements.txt` when `gflow` is refactored for performance soon.)
2. typing, linter compatibility, as in @EarlMilktea's comment

thierry-martinez:
I propose in my last commits a version which does not require `mypy`.

- The module `parameter.py` implements symbolic expressions so as to support parameters, but without all the symbolic. machinery of `mypy` (computations with known values are done numerically, not symbolically).

- The rest of the code does not depend on `parameter.py`: in particular, the code for simulators is generic. If we still need the `sympy` version of parameters, we can implement it separately, and simulators should work without any specific modification in the code of graphix.

- Tests for parameters are done in a specific `test_parameter.py` module, and the tests include circuit simulation.

shinich1:
> I propose in my last commits a version which does not require `mypy`.
> 
> * The module `parameter.py` implements symbolic expressions so as to support parameters, but without all the symbolic. machinery of `mypy` (computations with known values are done numerically, not symbolically).
> * The rest of the code does not depend on `parameter.py`: in particular, the code for simulators is generic. If we still need the `sympy` version of parameters, we can implement it separately, and simulators should work without any specific modification in the code of graphix.
> * Tests for parameters are done in a specific `test_parameter.py` module, and the tests include circuit simulation.

Thank you! I believe you mean `sympy` instead of `mypy` above? Let me take a detailed look soon.

shinich1:
@thierry-martinez the implementation looks good to me! we need to:

- [ ] fix visualization which seems to throw error in the presence of parameter in pattern
- [ ] test more situations, so we can catch more errors (currently `rz` only with no arithmetic of param covered?)
- [ ] do we intend to cover simulations without `subs`? if so we should throw error if parameterized and being simulated?
- [ ] update QAOA, VQE and QNN `examples`.
- [ ] add appropriate page to `docs` 
- [ ] device interface - may need care. easy solution is to ask `subs` before `PatternRunner` initialization.

graphix/parameter.py
```diff
@@ -0,0 +1,921 @@
...
+
+    def __rsub__(self, other) -> Expression | complex | type(NotImplemented):
+        if isinstance(other, numbers.Number):
+            return Sub(complex(other), self).simplify()
+        elif isinstance(other, Expression):
+            return Sub(other, self).simplify()
+        else:
+            return NotImplemented
+
+    def __rsub__(self, other) -> Expression | complex | type(NotImplemented):
```
EarlMilktea, line 70: Duplicated.

graphix/parameter.py
```diff
@@ -0,0 +1,921 @@
...
+import numbers
+from enum import Enum
+
+import numpy as np
+
+
+class Expression:
+    """Expression with parameters."""
+
+    def __mul__(self, other) -> Expression | complex | type(NotImplemented):
```
EarlMilktea, line 22: Please do not specify `NotImplemented` type, as this value itself is never returned from `__mul__`.

It's just a marker for internals.

EarlMilktea, line 22: Please annotate `other`.

graphix/parameter.py
```diff
@@ -0,0 +1,921 @@
...
+        return self.__name
+
+    @property
+    def type(self) -> type:
+        return self.__type
+
+    def subs(self, variable: Parameter, value: Expression | numbers.Number) -> Expression | complex:
+        if self == variable:
+            if isinstance(value, numbers.Number):
+                return complex(value)
```
EarlMilktea, line 231: Type checker complains that `Number` cannot be converted to `complex`. Could you check it?

graphix/parameter.py
```diff
@@ -0,0 +1,921 @@
...
+
+    @property
+    def name(self) -> str:
+        return self.__name
+
+    @property
+    def type(self) -> type:
+        return self.__type
+
+    def subs(self, variable: Parameter, value: Expression | numbers.Number) -> Expression | complex:
```
EarlMilktea, line 228: I once here that `numbers.XXX` does not work efficiently with modern typing systems (see Fluent Python 2nd ed.).

Do you come up with alternatives to that?

graphix/parameter.py
```diff
@@ -0,0 +1,921 @@
...
+
+
+def is_integer(expr: Expression | complex) -> bool:
+    if isinstance(expr, complex):
+        return expr.imag == 0 and float(expr).is_integer()
+    return expr.is_integer()
+
+
+class Compound(Expression):
+    def __init__(self, children: typing.Iterable[Expression | complex]) -> None:
```
EarlMilktea, line 289: - `typing` is not imported at all.
- Please import from `collections.abc`.

graphix/parameter.py
```diff
@@ -0,0 +1,921 @@
+"""Parameter class
```
EarlMilktea, line 1: I'm encountering a bunch of type warnings. Are you using linter/intellisense?

graphix/parameter.py
```diff
@@ -0,0 +1,921 @@
...
+    return rebuild_linear_combination(terms)
+
+
+class Product(Binary):
+    def simplify(self) -> Expression | complex:
+        simpl = super().simplify()
+        return simplify_product(simpl)
+
+    def is_float(self) -> bool:
+        return lhs.is_float() and rhs.is_float()
```
EarlMilktea, line 816: `lhs` undefined.

graphix/parameter.py
```diff
@@ -0,0 +1,921 @@
...
+            add_linear_combination_terms(terms, coef * mul_coef, others)
+    elif isinstance(expr, Minus):
+        add_linear_combination_terms(terms, -coef, expr.operand)
+    elif isinstance(expr, complex):
+        terms[None] = terms.get(None, 0) + coef * expr
+    else:
+        terms[expr] = terms.get(expr, 0) + coef
+
+
+def syntactic_order_key(key: Expression | complex | None) -> Any:
```
EarlMilktea, line 603: Please import `typing`. Additionally, please do not use `Any` if possible as it ruins type annotations.

EarlMilktea (changes requested):

EarlMilktea:
I'm a little bit concerned about...

- complexity: Is it possible to maintain `parameter.py` ? It's already really huge.
- performance: I feel that current impl. cannot be faster than sympy as it's doing almost the same things.

shinich1:
We suggest the following:

- implement 'placeholder' parameter implementation in `graphix`
- move symbolic stuff outside, in https://github.com/TeamGraphix/graphix-symbolic/tree/master.

thierry-martinez:
As @shinich1 suggested in the above message, this PR now only provides abstract base classes for parameters, and a minimal concrete class for placeholders (symbolic angles that can solely be substituted, and that cannot appear in computation). The sympy part is implemented as a plugin in the `graphix-symbolic` repo: https://github.com/TeamGraphix/graphix-symbolic/pull/1 .

graphix/parameter.py
```diff
@@ -0,0 +1,255 @@
...
+                raise ValueError("Expression expected")
+            return obj
+
+        return pydantic_core.core_schema.no_info_plain_validator_function(function=check_expression)
+
+
+class PlaceholderOperationError(ValueError):
+    def __init__(self):
+        super().__init__(
+            "Placeholder angles do not support any form of computation before substitution. Either use `subst` with an actual value before the computation, or use a symbolic parameter implementation, such that https://github.com/TeamGraphix/graphix-symbolic ."
```
shinich1, line 113: typo: `subst` -> `subs`?

tests/test_parameter.py
```diff
@@ -0,0 +1,43 @@
...
+    assert pattern.subs(alpha, 2).is_parameterized()
+    pattern23 = pattern.subs(alpha, 2).subs(beta, 3)
+    # A full instantiated pattern is no longer parameterized.
+    assert not pattern23.is_parameterized()
+    assert list(pattern23) == [
+        graphix.command.M(node=0),
+        graphix.command.M(node=1, angle=2),
+        graphix.command.N(node=2),
+        graphix.command.M(node=2, angle=3),
+    ]
```
shinich1, line 42: don't we want to break them up into different test cases, as done in other test files? I understand it's smaller number of lines but it's perhaps a bit difficult to debug in case of errors.

thierry-martinez: Fixed in 697ff2d.

graphix/parameter.py
thierry-martinez: Fixed in 383fe22.

EarlMilktea:
@thierry-martinez There are ruff errors remain unresolved. Could you make sure that you're using the CIs correctly?

thierry-martinez:
Sorry, @EarlMilktea, I didn't merge this branch with `master` yet: it is now done and CI seems to run correctly now.

thierry-martinez:
Updated! Thanks.

graphix/parameter.py
```diff
@@ -0,0 +1,258 @@
...
+                raise ValueError("Expression expected")
+            return obj
+
+        return pydantic_core.core_schema.no_info_plain_validator_function(function=check_expression)
+
+
+class PlaceholderOperationError(ValueError):
+    def __init__(self):
+        super().__init__(
+            "Placeholder angles do not support any form of computation before substitution. Either use `subs` with an actual value before the computation, or use a symbolic parameter implementation, such that https://github.com/TeamGraphix/graphix-symbolic ."
```
shinich1, line 116: typo: such **as** https://github.com/TeamGraphix/graphix-symbolic

graphix/sim/statevec.py
```diff
@@ -495,6 +503,16 @@ def expectation_value(self, op, qargs):
         st1.evolve(op, qargs)
         return np.dot(st2.psi.flatten().conjugate(), st1.psi.flatten())
 
+    def subs(self, variable, substitute) -> Statevec:
+        """Return a copy of the state vector where all occurrences of
+        the given variable in measurement angles are substituted by
+        the given value.
+
+        """
```
shinich1, line 511: is it correct to understand this can be done only if `sympy`-version of parameter (e.g. graphix-parameter module) is used? if so, could you mention that in the docstring?

tests/test_pattern.py
```diff
@@ -11,6 +11,7 @@
 
 import graphix.clifford
 import graphix.ops
+import graphix.parameter
```
shinich1, line 14: should we add option to get parameterized (say replace a small number of angles with placeholder) circuits in `test.random_circuit.get_rand_circuit`, so that we can add parameterized version of patterns in the existing test suite?

EarlMilktea:
@thierry-martinez 

Checked the updates.
As for the organization and quality of the code, I see (almost) no problems, but for the interface or ultimately the approach itself, I'm strongly worried about future maintainability and performance.

I suggest the following approach:

1. Use unique identifier to distinguish different angles (not by `sympy.Symbol` or `float`)
1. Define a class that keeps track of measurement angle/plane changes introduced on simplification
1. In a separate package, read out the info. and perform symbolic calculus

The point is that, we only need to take into account sign flip $\alpha \to -\alpha$ and constant shift $\alpha \to \alpha + \pi$.
Obviously it does not require full symbolic interfaces or ABCs, just a tiny class.
With this approach, we may even rely on bindings or `numba` in the future.

graphix/parameter.py
thierry-martinez: Fixed in 49e63df .

graphix/sim/statevec.py
thierry-martinez: Fixed in af74a37 .

tests/test_pattern.py
thierry-martinez: Fixed in commit 05c05a0. These tests highlight that `Placeholder` objects without any form of computation were not suitable for use in rotation gates, as angles in rotation gates are transformed by the transpiler into expressions like `-alpha / pi`. This observation is related to @EarlMilktea's remark about simple computations such as sign flips and constant shifts. The commit 05c05a0 adds support for affine computation, which is simple enough to maintain while allowing these necessary operations.

thierry-martinez:
@EarlMilktea:
>     1. Use unique identifier to distinguish different angles (not by `sympy.Symbol` or `float`)

I think we have better to keep things generic at this level: most patterns will only have `float` angles, and I don't think there is a need for unique identifiers in this case. `Placeholder` uses object identity (`is`) in `__eq__` and I don't think we can be more efficient than this (and it is a kind of unique identifier...). We don't use `sympy` anymore in this PR.
 
>     2. Define a class that keeps track of measurement angle/plane changes introduced on simplification
> [...]
> The point is that, we only need to take into account sign flip α → − α and constant shift α → α + π . 

I added support for affine expressions in 05c05a0: the code is simple to maintain and support sign flip, constant shift, and the operations performed in the transpiler.

That's enough to handle transpiling and pattern optimizations, but that's not enough for simulation.

>     3. In a separate package, read out the info. and perform symbolic calculus

Yes, the package `graphix-symbolic` uses `sympy` to perform symbolic simulations.

graphix/parameter.py
```diff
@@ -0,0 +1,311 @@
...
+    def conjugate(self) -> ExpressionOrNumber: ...
+
+    @abstractmethod
+    def sqrt(self) -> ExpressionOrNumber: ...
+
+    @abstractmethod
+    def subs(self, variable: Parameter, value: ExpressionOrNumber) -> ExpressionOrNumber: ...
+
+    @classmethod
+    def __get_pydantic_core_schema__(
```
EarlMilktea, line 97: Is it really necessary?

graphix/parameter.py
```diff
@@ -0,0 +1,311 @@
...
+
+if TYPE_CHECKING:
+    import pydantic
+
+
+class Expression(ABC):
+    """Expression with parameters."""
+
+    @abstractmethod
+    def __mul__(self, other) -> ExpressionOrNumber: ...
```
EarlMilktea, line 28: `Number` is not mypy friendly. How about using `typing_extensions.Self` or generics?

graphix/parameter.py
```diff
@@ -0,0 +1,311 @@
...
+    This function is used to apply substitution to collections where
+    some elements are Expression and other elements are just
+    plain numbers.
+
+    """
+    subs = getattr(value, "subs", None)
+    if subs:
+        new_value = subs(variable, substitute)
+        if isinstance(new_value, Number):
+            return complex(new_value)
```
EarlMilktea, line 309: Seeing many type errors originating from `Number`.

graphix/sim/density_matrix.py
```diff
@@ -313,11 +315,30 @@ def apply_channel(self, channel: KrausChannel, qargs):
...
 
-    def __init__(self, pattern, max_qubit_num=12, pr_calc=True, input_state: Data = graphix.states.BasicStates.PLUS):
+    def __init__(
+        self,
+        pattern,
+        max_qubit_num=12,
+        pr_calc=True,
+        input_state: Data = graphix.states.BasicStates.PLUS,
+        rng: np.random.Generator | None = None,
+    ):
```
EarlMilktea, line 341:
```suggestion
    ) -> None:
```

EarlMilktea (changes requested):
General comments:

- Please do not use `Number` unless you are absolutely sure that it will never cause type warnings.
- I'm not sure why pydantic is necessary.

EarlMilktea:
> Placeholder uses object identity (is) in __eq__ and I don't think we can be more efficient than this (and it is a kind of unique identifier...). 

Anyway I love this approach!

graphix/sim/density_matrix.py
thierry-martinez: Type annotations for the whole `density_matrix.py` module have been added in 173c70e.

graphix/parameter.py
thierry-martinez: Fixed in 492b772: indeed, [PEP 484](https://peps.python.org/pep-0484/#the-numeric-tower) recommends `int`, `float` and `complex` in type annotations instead of `Number`.  Dynamic tests `isinstance(x, Number)` are replaced by `isinstance(x, SupportsFloat)`, followed by explicit `float(x)` coercions (in particular, numpy numbers are not instances of `float` but `SupportsFloat`).

graphix/parameter.py
thierry-martinez: Fixed in https://github.com/TeamGraphix/graphix/commit/492b772ce7ec15e8d22566921a8fff6d87e99028.

shinich1:
> @thierry-martinez the implementation looks good to me! we need to:
> 
> * [ ]  fix visualization which seems to throw error in the presence of parameter in pattern
> * [x]  test more situations, so we can catch more errors (currently `rz` only with no arithmetic of param covered?)
> * [ ]  do we intend to cover simulations without `subs`? if so we should throw error if parameterized and being simulated?
> * [ ]  update QAOA, VQE and QNN `examples`.
> * [ ]  add appropriate page to `docs`  (just autodoc, not much to write - see existing pages)
> * [ ]  device interface - may need care. easy solution is to ask `subs` before `PatternRunner` initialization.

bump above list

thierry-martinez:
Thank you, @shinich1, for the list!

> > * [x]   fix visualization which seems to throw error in the presence of parameter in pattern

Fixed in 34c16ed, with a dedicated test.

> > * [x]   test more situations, so we can catch more errors (currently `rz` only with no arithmetic of param covered?)

Should have been fixed with test on random circuits (05c05a0).

> > * [x]   do we intend to cover simulations without `subs`? if so we should throw error if parameterized and being simulated?

Simulations without `subs` are handled with [graphix-symbolic](https://github.com/TeamGraphix/graphix-symbolic). In 3a6e38d, I added a test to check that there is an exception `PlaceholderOperationError` if a simulation is run on a circuit with a non-substituted `Placeholder`.

> > * [ ]   update QAOA, VQE and QNN `examples`.

I am not sure what to do with QAOA since there is only one pattern in `qaoa.py`.
I added a placeholder version in VQE (with performance comparison) in 4cb326b.
I added placeholders for QNN in b337abe.

> > * [x]   add appropriate page to `docs`  (just autodoc, not much to write - see existing pages)

Placeholders and affine expressions are documented in 149b089.

> > * [x]   device interface - may need care. easy solution is to ask `subs` before `PatternRunner` initialization.

I added a test to check that there is an exception `CircuitError` if a `PatternRunner` is initialized on non-substituted parameterized patterns in 1e2f3c1.

EarlMilktea:
@thierry-martinez 

Is it impossible to move symbolic-related ABCs/`subs` functions to `graphix-symbolic` ?
Additionally, please do not define unnecessary mathematical/arithmetic abstract methods in `AffineExpression`, as they're not implemented or even used in `graphix`: consider extending it in the separate package. 
I understand `graphix` should be updated so that non-`float` can pass through functions, but the impact should be minimal.

thierry-martinez:
@EarlMilktea I believe that all the functions (`subs`, `xreplace`, and the trigonometric functions) are used in graphix. Simple and parallel substitutions apply to both `Placeholder` and `AffineExpression`. All trigonometric functions are potentially called during simulation; in these cases, `AffineExpression` raises an `Exception`, as expected, since simulation is not supported. The purpose of the parameter module is to define a common interface, which is why an abstract base class is defined.

EarlMilktea:
@thierry-martinez 
OK, I trust you.
Let me do final checks...

examples/MBQCvqe.py
```diff
@@ -81,22 +103,32 @@ def simulate_mbqc(self, params, backend="tensornetwork"):
...
-    def compute_energy(self, params):
+    def compute_energy(self, params: Iterable[float]):
         # Simulate the MBQC circuit using tensor network backend
         tn = self.simulate_mbqc(params, backend="tensornetwork")
         # Compute the expectation value using MBQCTensornet.expectation_value
         energy = tn.expectation_value(self.hamiltonian, qubit_indices=range(self.n_qubits))
         return energy
 
 
+class MBQCVQE_WithPlaceholders(MBQCVQE):
```
EarlMilktea, line 114: Please use PEP8-compliant name (maybe without `_` ?)

EarlMilktea (changes requested):

examples/MBQCvqe.py
thierry-martinez: Fixed in 8e9fb19.

thierry-martinez:
I added missing `xreplace` in `Circuit`, `Statevec` and `DensityMatrix`, and I replace some uses of `numbers.Number` with `typing.SupportsComplex`.

EarlMilktea:
I re-reviewed you code and felt that the whole logic of `graphix` (inclding simulator) is affected.
Could you request reviews from @masa10-f , as I'm not sure your idea is compatible with ZX calculus?
Additionally, please wait until his changes are merged.
Meantime, let me carefully examine your code in terms of type stability, performance, and maintainability.

graphix/sim/base_backend.py
```diff
@@ -1,14 +1,21 @@
...
 import graphix.pauli
 
 
 def op_mat_from_result(vec: tuple[float, float, float], result: bool) -> np.ndarray:
-    op_mat = np.eye(2, dtype=np.complex128) / 2
+    if all(isinstance(value, numbers.Number) for value in vec):
+        op_mat = np.eye(2, dtype=np.complex128) / 2
+    else:
+        # At least one value is symbolic (i.e., parameterized).
+        op_mat = np.eye(2, dtype="O") / 2
```
EarlMilktea, line 18: `dtype="O"` needs extra care as it can impact the performance of simulators.

EarlMilktea (changes requested):
WIP: May be updated later.

graphix/parameter.py
```diff
@@ -0,0 +1,348 @@
...
+from __future__ import annotations
+
+import sys
+import typing
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Mapping, SupportsComplex, SupportsFloat, TypeVar
+
+
+class Expression(ABC):
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in 9f97fda. Indeed, `pow`, `rpow`, `tan`, `arcsin`, `arccos`, `arctan` and `log` are unused. The other methods are used either by the transpiler, the pattern optimizations or the back-ends.

graphix/parameter.py
```diff
@@ -0,0 +1,348 @@
...
+class Parameter(Expression):
+    """Abstract class for substituable parameter."""
+
+    ...
+
+
+class PlaceholderOperationError(ValueError):
+    def __init__(self):
+        super().__init__(
+            "Placeholder angles do not support any form of computation before substitution except affine operation. Either use `subs` with an actual value before the computation, or use a symbolic parameter implementation, such as https://github.com/TeamGraphix/graphix-symbolic ."
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in d9dcef5.

graphix/parameter.py
```diff
@@ -0,0 +1,348 @@
...
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Mapping, SupportsComplex, SupportsFloat, TypeVar
+
+
+class Expression(ABC):
+    """Expression with parameters."""
+
+    @abstractmethod
+    def __mul__(self, other) -> ExpressionOrFloat: ...
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in b720e48.

graphix/parameter.py
```diff
@@ -0,0 +1,348 @@
...
+    method `subs`, then return `value.subs(variable, substitute)`
+    (coerced into a complex if the result is a number).  If `value`
+    does not implement `subs`, `value` is returned unchanged.
+
+    This function is used to apply substitution to collections where
+    some elements are Expression and other elements are just
+    plain numbers.
+
+    """
+    subs = getattr(value, "subs", None)
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in d554a7c.

graphix/sim/base_backend.py
thierry-martinez: We know that symbolic simulations are extra-costly!

graphix/parameter.py
```diff
@@ -0,0 +1,348 @@
...
+        return self.a * float(value) + self.b
+
+    def subs(self, variable: Parameter, value: ExpressionOrSupportsFloat) -> ExpressionOrComplex:
+        if variable == self.x:
+            return self.evaluate(value)
+        return self
+
+    def xreplace(self, assignment: Mapping[Parameter, ExpressionOrSupportsFloat]) -> ExpressionOrComplex:
+        # `value` can be 0, so checking with `is not None` is mandatory here.
+        if (value := assignment.get(self.x)) is not None:
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in 7189dba.

graphix/instruction.py
```diff
@@ -59,7 +60,9 @@ class RotationInstruction(OneQubitInstruction):
     Rotation instruction base class model.
     """
 
-    angle: float
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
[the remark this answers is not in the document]
thierry-martinez: Setting `arbitrary_types_allowed=True` does not eliminate validation. For types that do not implement their own validation (such as `Expression` in this case), the validation defaults to using `isinstance`, which is appropriate. Previously, I implemented validation in `Expression` specifically using `isinstance` (as discussed here: https://github.com/TeamGraphix/graphix/pull/158#discussion_r1722512340). Both approaches are equivalent: we can either specify the use of `isinstance` for validation within the `Expression` implementation, or we can specify it when using `Expression` within the model.

graphix/sim/base_backend.py
EarlMilktea: Then I suggest completely isolating the logic to different functions: `numpy` is notorious for unpredictable automatic type conversion, so I don't want it close to numerical simulator that should be as lightweight as possible.
Current approach (e.g., add conditional branches to existing functions) is very vulnerable to accidental contamination.

graphix/instruction.py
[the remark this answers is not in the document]
EarlMilktea: I see. Sorry for my misunderstanding...
By the way, why `BaseModel` is introduced here in the first place?

graphix/parameter.py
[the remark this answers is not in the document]
EarlMilktea: Ah sorry, I made you confused.
My true implication here is that, `Any` with `@abstractmethod` is acceptable (as we cannot call them at all and the concrete argument types can be overridden), but as usual, not allowed for implementation.

graphix/parameter.py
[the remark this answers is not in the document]
EarlMilktea: Great approach!

graphix/parameter.py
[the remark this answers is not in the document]
EarlMilktea: Memo: `mypy` check result for this file:

```bash
graphix/parameter.py:185: error: Argument 1 to "float" has incompatible type "Expression | SupportsFloat"; expected "str | Buffer | SupportsFloat | SupportsIndex"  [arg-type]
graphix/parameter.py:282: error: Argument 2 to "subs" of "Expression" has incompatible type "Expression | SupportsFloat"; expected "Expression | float"  [arg-type]
graphix/parameter.py:285: error: Incompatible return value type (got "Expression", expected "T | complex")  [return-value]
graphix/parameter.py:302: error: Argument 1 to "xreplace" of "Expression" has incompatible type "Mapping[Parameter, Expression | SupportsFloat]"; expected "Mapping[Parameter, Expression | float]"  [arg-type]
graphix/parameter.py:305: error: Incompatible return value type (got "Expression", expected "T | complex")  [return-value]
```

graphix/sim/base_backend.py
thierry-martinez: There is now an explicit Boolean flag `symbolic` (introduced in 1ae5cca) to enable computation with `dtype="O"`, so that contamination cannot occur accidentally.

graphix/parameter.py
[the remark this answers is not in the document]
thierry-martinez: `parameter.py` and `test_parameter.py` are now type-checked by CI (2976c00).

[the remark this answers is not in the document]
thierry-martinez: `parameter.py` and `test_parameter.py` are now type-checked by CI (2976c00).

graphix/instruction.py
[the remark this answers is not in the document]
thierry-martinez: This PR is now rebased on top of #205, which does not use `model_config` anymore.

graphix/parameter.py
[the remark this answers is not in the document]
thierry-martinez: `parameter.py` and `test_parameter.py` now type-checks.

shinich1:
@EarlMilktea I believe this is the direction we agreed on at the meeting in Paris. could you review this PR?

EarlMilktea:
@thierry-martinez (CC: @shinich1 )

I'm basically agree with your idea, but I still feel that this PR should be postponed until other TODOs are resolved, because...

- this PR makes the Pythonic layer thicker

It would make it very difficult to improve/assess the performance of the simulator if this PR were to be merged before simulator optimization/refactoring, because simulator optimization is basically the process of reducing Pythonic tasks (ex. using `numpy` to delegate heavy tasks to C).

- this PR needs to use `Any`

I admit this PR somehow require `Any`, but as `Any` can ruin the type checker, I would prefer resolving annotation issues (`mypy` errors, lacking annotations, etc.) first.

thierry-martinez:
Postponing the PR forces it to bear the ongoing maintenance cost until it’s finally merged, rather than distributing that cost among future PRs, and this approach doesn’t scale well. I’ve already spent significant time on this PR, which covers a feature that has widespread support, is definitely needed at Inria, and was praised during the workshop.

I agree that symbolic computation will require special attention when optimizing the simulators. However, this makes it even more important to merge this PR before those optimizations proceed, so we can verify that any optimized simulators don’t break symbolic computation. By the way, optimizing the simulators is ongoing work in our group at Inria, and we are aware of these concerns. However, that effort should not block merging this PR. More generally, merging features early allows subsequent optimizations and improvements to be developed without risking regressions.

Regarding type-checking, this PR passes the CI linter and type-checker. If you believe additional checks are necessary, they should be introduced in a separate PR, as they fall outside the scope of this one.

shinich1:
I'd personally agree with going for merge now, we might be left with additional future work but the benefit is huge and there are papers to be written out of these features. 
Thanks @EarlMilktea for careful review, though. We appreciate your thoughts as always.

shinich1:
@thierry-martinez could you resolve conflict with master, and check that https://github.com/TeamGraphix/graphix-symbolic/pull/1 is ready? we should cross-test the two packages and we should then merge this. thanks!

thierry-martinez:
Done. It appears I cannot merge the pull request https://github.com/TeamGraphix/graphix-symbolic/pull/1 myself, so the CI is currently checking that branch.

shinich1:
@thierry-martinez thanks, my bad the branch protection rule was too strict on the other repo. It's now fixed and it should be possible to merge there.

thierry-martinez:
Thanks. CI now uses `graphix-symbolic` master in nox tests.

shinich1:
@thierry-martinez thanks!

- shall we set up pypi for graphix-symbolic? then things will be simpler in here.
- it'd be nice to have a simple, short example demonstrating simple usage of parameterized pattern.
- rendered docs show `SupportsFloat` not float along with `Expression`. is this correct?
image

thierry-martinez:
>     * shall we set up pypi for graphix-symbolic? then things will be simpler in here.

I propose to discuss about that point there: https://github.com/TeamGraphix/graphix-symbolic/issues/2
 
>     * it'd be nice to have a simple, short example demonstrating simple usage of parameterized pattern.

Do you mean an example shorter than the VQE one? The tests already include short, purely formal examples of parameterized patterns. Is this the kind of example you'd like to add to the examples section?

>     * rendered docs show `SupportsFloat` not float along with `Expression`. is this correct?

Any `float` is an instance of `SupportsFloat`, but `SupportsFloat` also includes objects that are not instance of `float`, such that `np.float32`.

EarlMilktea:
@thierry-martinez (CC: @shinich1 )

I really appreciate your efforts, but still, I cannot fully agree with your idea.

If I understand correctly, this PR is consisting of two parts:

1. make it possible to manipulate commands without specifying concrete angles
2. keeping track of mathematical operations

For the former part, I feel it's definitely necessary and would not hinder future refactoring.
But for the latter, I'm very very worried about code qualities and future impacts, as this part is inevitably the same as (or at least similar to) reinventing sympy, which I @shinich1 are trying to avoid as commented in https://github.com/TeamGraphix/graphix/pull/158#issuecomment-2161306427 _before_ the workshop.
If you _really absolutely_ want this feature, my request is either

- completely externalize the latter part,
- reduce features so that the code base will not be too complex/dynamic (at least please remove `Any`).

thierry-martinez:
Thank you for your feedback. I’d like to address the points you raised:

The comment in #158 was made prior to our meeting with @shinich1 in Oxford. During that meeting—and later confirmed at the workshop—we agreed on several key points:
  - We prefer not to introduce `sympy` as a dependency for graphix, which was why early commits in this PR removed it.
  - We also agreed that reimplementing a substantial part of `sympy` for handling symbolic simulation is not ideal.
  - The extensive symbolic simulation functionality will be delegated to a separate repository, `graphix-symbolic`.
  - Within `graphix`, we maintain only a small fragment for handling symbolic affine expressions, which is sufficient for our needs in symbolic transpiling and Pauli presimulation.
  
I understand your concerns about potential code complexity. However, all occurrences of `Any` in the current commits are limited to the signatures of dunder methods, whose types are predetermined by Python. Additionally, the PR passes all CI checks. While we are open to further strengthening the typing rules, such improvements would be better addressed in a dedicated follow-up PR rather than blocking this one.

Lastly, this feature has been praised multiple times during the workshop, and there is strong user interest in its inclusion. The current implementation reflects the collective decision made in Oxford and during the workshop, rather than an individual preference.

I hope this clarifies our approach and addresses your concerns.

EarlMilktea:
> the PR passes all CI checks

Please do not blindly trust the type checking CI for this repo., since most of the important rules were dropped due to massive errors when I introduced the CI.
I expect this project will also be strictly typed/linted as in fastflow and graphix-zx (`Any` is banned by ruff + optionally by mypy) in the future, meaning that somebody will anyway need to work on eliminating `Any`.
Please do not increase future tasks.

> affine expressions

I'm for including these mechanics in graphix (signal sign flip + signal addition without specifying angles), but also not sure why `.sin` or `.cos` similar are needed and cannot be externalized.

graphix/ops.py
```diff
@@ -68,7 +74,23 @@ class Ops:
     )
 
     @staticmethod
-    def rx(theta: float) -> npt.NDArray[np.complex128]:
+    def __cast_array(
```
EarlMilktea, line 77: Please use `_cast_array` as private methods seem to be not idiomatic in Python (I'm not sure why...)

graphix/ops.py
```diff
@@ -68,7 +74,23 @@ class Ops:
...
+        array: Iterable[Iterable[ExpressionOrComplex]], theta: ExpressionOrFloat
+    ) -> npt.NDArray[np.complex128] | npt.NDArray[np.object_]:
```
EarlMilktea, line 79: Isn't it possible to avoid `NDArray[np.object_]` ?

We should avoid it because

- poorly supported compared to numeric types
- there are no performance benefits
- difficult to annotate: `np.object_` tells us no info. about element type

graphix/parameter.py
```diff
@@ -0,0 +1,409 @@
+"""Parameter class.
```
EarlMilktea, line 1: MEMO: use `object` or more concrete types instead of `Any`.

graphix/parameter.py
```diff
@@ -0,0 +1,409 @@
...
+
+    a: float
+    x: Parameter
+    b: float
+
+    def offset(self, d: float) -> AffineExpression:
+        """Add `d` to the expression."""
+        return AffineExpression(a=self.a, x=self.x, b=self.b + d)
+
+    def __scale_non_null(self, k: float) -> AffineExpression:
```
EarlMilktea, line 158: Same as above (use `_scale_non_null`)

graphix/parameter.py
```diff
@@ -0,0 +1,409 @@
...
+
+    ExpressionOrSupportsComplex = typing.Union[Expression, SupportsComplex]
+
+
+T = TypeVar("T")
+
+
+# The return type could be `T | Expression | complex` since `subs` returns `Expression` only
+# if `T == Expression`, but `mypy` does not handle this yet: https://github.com/python/mypy/issues/12989
+def subs(value: T, variable: Parameter, substitute: ExpressionOrSupportsFloat) -> T | Expression | complex:
```
EarlMilktea, line 328: Maybe `overload` will help?

graphix/sim/base_backend.py
```diff
@@ -81,31 +85,38 @@ def swap(self, i: int, j: int) -> None:
         self.__dict[node_j] = i
 
 
-class State:
+class State(ABC):
     """Base class for backend state."""
 
+    @abstractmethod
+    def flatten(self) -> npt.NDArray:
```
EarlMilktea, line 92: Please add element type.

thierry-martinez: Fixed in 0cd2b36.

graphix/parameter.py
thierry-martinez: Fixed in 0cd2b36. Thanks.

graphix/parameter.py
thierry-martinez: Fixed in 0cd2b36. Thanks.

graphix/ops.py
```diff
@@ -98,12 +129,19 @@ def ry(theta: float) -> npt.NDArray[np.complex128]:
...
+    @staticmethod
+    def rz(theta: float) -> npt.NDArray[np.complex128]: ...
+
+    @overload
+    @staticmethod
+    def rz(theta: Expression) -> npt.NDArray[np.object_]: ...
 
     @staticmethod
-    def rz(theta: float) -> npt.NDArray[np.complex128]:
+    def rz(theta: ExpressionOrFloat) -> npt.NDArray[np.complex128] | npt.NDArray[np.object_]:
```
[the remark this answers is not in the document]
thierry-martinez: `rz` has already been annotated with `@overload` in the lines above (since commit 47760f2b).

graphix/ops.py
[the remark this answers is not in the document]
thierry-martinez: Fixed in 0cd2b36. Thanks.

graphix/ops.py
thierry-martinez: This is fixed in 0cd2b36, but it would be interesting to know why.

examples/mbqc_vqe.py
```diff
@@ -19,55 +19,77 @@
...
+from graphix.transpiler import Angle
 
 Z = np.array([[1, 0], [0, -1]])
 X = np.array([[0, 1], [1, 0]])
 
 
 # %%
 # Define the Hamiltonian for the VQE problem (Example: H = Z0Z1 + X0 + X1)
-def create_hamiltonian():
+def create_hamiltonian() -> npt.NDArray:
```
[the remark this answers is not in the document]
thierry-martinez: This is fixed in 0cd2b36. Thanks.

graphix/ops.py
thierry-martinez: Actually, `npt.NDArray[np.object_]` corresponds to a specific data type for a NumPy array, where the elements can be any Python object, thereby introducing an indirection in the data representation. In contrast, numeric data types are stored unboxed. Supporting NumPy arrays containing non-numeric Python objects is crucial for symbolic computation since symbolic expressions cannot be represented as unboxed numeric values. The proposed type annotation reflects that arrays will be numeric in numeric computations and can contain arbitrary Python objects (in this case, symbolic expressions) in symbolic computations.

graphix/ops.py
EarlMilktea: I suspect it would be related to [the open-closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle).
Based on this idea, methods should be protected (sibling classes can access to them) rather than private (no access) while "private variables" are still allowed because there are no way to "inherit" them.

shinich1:
@EarlMilktea if no further comments comments by the end of this week, I let authors assume this approved.

PR #163, opened 2024-06-10
Support python312

### LINKED ISSUE #127: Support Python3.12

Graphix currently supports Python 3.8 through 3.11. Python 3.12 was released last October, so It would be beneficial to support 3.12.

**Additional context**

I would like to have criteria for when to add new Python versions and when to drop older ones. It seems reasonable to me to follow the [status of Python versions](https://devguide.python.org/versions/). Following this, we will drop the oldest next October and add the latest once released. If anyone has opinions or suggestions, please share them in this issue.

EarlMilktea:
Resolves #127 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "windows-2022", "macos-latest"]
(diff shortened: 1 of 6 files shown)


shinich1 (approved):
Thanks! checked locally too, seems to work well. please squash and merge.

PR #166, opened 2024-06-17
Enable efficient contraction order simulation

nabe98:
- Format added code by `black` and `isort`
  - See `pyproject.toml` for configurations

Then, please fill in below:

**Context (if applicable):**

Enabled TN contraction based state vector simulator, which is more efficient compared with pure state-vector simulation.

**Related issue:**
#92 


graphix/random_circuit.py
```diff
@@ -84,6 +84,21 @@ def get_rand_circuit(
...
+        Number of qubits in the circuit. Must be at least 2.
+    depth : int
+        Number of layers in the circuit.
+    rng : Generator
+        Random number generator.
+    use_rzz : bool, optional
+        Whether to include RZZ gates, by default False.
+    use_ccx : bool, optional
+        Whether to include CCX gates, by default False.
+    """
```
shinich1, line 101: any particular motivation to move to main module? if this is only to use for benchmarking, I suggest copying to benchmarks folder.

mgarnier59: Random circuits are also extensively used in tests so maybe in random_objects? Or another dedicated file?

graphix/sim/tensornet.py
```diff
@@ -568,7 +572,12 @@ def expectation_value(self, op, qubit_indices, output_node_indices=None, **kwagr
             target_nodes = [output_node_indices[ind] for ind in qubit_indices]
             out_inds = output_node_indices
         op_dim = len(qubit_indices)
-        op = op.reshape([2 for _ in range(2 * op_dim)])
+        op = op.reshape(
+            [
+                2,
+            ]
+            * (2 * op_dim)
+        )
```
shinich1, line 580: use `black -l 120`?

graphix/sim/tensornet.py
```diff
@@ -502,28 +504,30 @@ def get_basis_amplitude(self, basis, **kwagrs):
...
-        """Retrieve the statevector from the tensornetwork.
-        This method tends to be slow however we plan to parallelize this.
+    def to_statevector(self, skip=False, **kwagrs) -> NDArray:
+        """Take outer product of the tensors in the network and return the statevector.
 
         Parameters
         ----------
-        indices (optional): list of int
-            target qubit indices. Default is the MBQC output nodes (self.default_output_nodes).
+        skip : bool
```
shinich1, line 512: argument name `skip` is way too generic and very difficult to understand. Consider making it `skip_tn_simp` or restructure. Also, why/who/when do we want to skip the simplification?

graphix/sim/tensornet.py
```diff
@@ -502,28 +504,30 @@ def get_basis_amplitude(self, basis, **kwagrs):
         coef = self.get_basis_coefficient(basis, **kwagrs)
         return abs(coef) ** 2
 
-    def to_statevector(self, indices=None, **kwagrs):
-        """Retrieve the statevector from the tensornetwork.
-        This method tends to be slow however we plan to parallelize this.
+    def to_statevector(self, skip=False, **kwagrs) -> NDArray:
```
shinich1, line 507: Please add very detailed description (100+ words, with equations and diagrams) to this method as this is rather uncommon way to use TN (for most people, at least afaik). 
Is this proven to work for any case or are there limitations? worth pointing out we're NOT performing the intermediate measurements according to probability distribution? what is the expected space/time cost compared to statevector sim?

graphix/sim/tensornet.py
nabe98: Yes. If I left this like `op = op.reshape([2,] * (2 * op_dim)`, cannot pass the check for black.

graphix/sim/tensornet.py
nabe98: This option is not that essential, so I removed `skip`.

graphix/sim/tensornet.py
nabe98: I added that based on notes and discussion in #92 .

shinich1:
@nabe98 can you paste benchmarking result (plot) on this page?

nabe98:
OK.
I'll share that as soon as I return to my home (around 8/7).

nabe98:
@shinich1 
Here is the benchmarking result plot. 
simulation_time

nabe98:
simulation_time
The latest benchmarking plot with error bars (for 5 iterations).

PR #170, opened 2024-06-19
Resolve deprecated type hints

### LINKED ISSUE #169: Resolve deprecated type hints

Deprecated annotation syntax, such as `typing.List`, is still used.

EarlMilktea:
Resolves #169 .

### DIFF
--- a/graphix/graphsim/basegraphstate.py
+++ b/graphix/graphsim/basegraphstate.py
@@ -20,10 +20,7 @@ try:
 
     RUSTWORKX_INSTALLED = True
 except ModuleNotFoundError:
-    # !!! INVALID !!!
-    # We cannot define "dynamic" type hints
     rx = None
-    # PyGraph is not regarded as a type hint here
     PyGraph = None
 
 if sys.version_info >= (3, 10):
--- a/graphix/graphsim/basegraphstate.py
+++ b/graphix/graphsim/basegraphstate.py
@@ -20,7 +20,10 @@ try:
 
     RUSTWORKX_INSTALLED = True
 except ModuleNotFoundError:
+    # !!! INVALID !!!
+    # We cannot define "dynamic" type hints
     rx = None
+    # PyGraph is not regarded as a type hint here
     PyGraph = None
 
 if sys.version_info >= (3, 10):
--- a/graphix/graphsim/basegraphstate.py
+++ b/graphix/graphsim/basegraphstate.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
+import sys
 from abc import ABC, abstractmethod
-from typing import Union
 
 import networkx as nx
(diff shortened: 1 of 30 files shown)


graphix/graphsim/basegraphstate.py
```diff
@@ -19,12 +20,22 @@
 
     RUSTWORKX_INSTALLED = True
 except ModuleNotFoundError:
+    # !!! INVALID !!!
+    # We cannot define "dynamic" type hints
     rx = None
+    # PyGraph is not regarded as a type hint here
     PyGraph = None
```
shinich1, line 27: `rx = None` can be removed, perhaps?
`PyGraph` part - not sure yet how best to proceed. we can, in principle, require `rustworkx` instead of `networkx`, if this structure become too much of burden for maintaining/cleaning up. What is your thoughts @EarlMilktea ?

EarlMilktea: I personally prefer to make `import` itself optional: optional module is not imported automatically, and if users import without installing necessary optional deps., exception is raised.

With this design, we don't have to keep going after `ImportError` .

shinich1: I see, does that mean rustworkx is `import`'ed only when `RxGraphState` is instantiated?

EarlMilktea: I misunderstood a little bit...
As this file is for the common base class, we cannot skip importing it.
Let me discuss workarounds later.

EarlMilktea: The situation is not so easy as I expected...

@shinich1 Could you review and merge (if OK) this PR first?

I may need to carefully examine the codes to fix the problem above.

shinich1: @EarlMilktea do we want to keep the comments here (`!!! invalid ...`) or is it cleaner to reomve this part and put in an issue?

This PR looks good overall, let me take a look again later. I also ask @thierry-martinez to have a quick look through.

EarlMilktea: May better to open a new issue.

thierry-martinez (approved):
Looks good to me! I approve this PR.

graphix/graphsim/basegraphstate.py
shinich1: Sounds good! 
please remove these lines before merging, and put in a new issue.

EarlMilktea:
⚠ I used `push -f` to reword f54da115757e5cf592104d3f99ad7bf88033bdcf immediately after push.

graphix/graphsim/basegraphstate.py
EarlMilktea: Opened #171 .

PR #178, opened 2024-07-16
Refactor backend for delegated measurement methods

jemappellesami:
Refactoring of the backend to enable delegated measurement-based QC protocols.

1. Decouple backends from patterns, to allow backend instantiation independent of any pattern
2. Better code sharing and less redundancy between backends
3. Introduction of `NodeIndex` data structure for efficient mapping between nodes and indices in the backend's internal state
4. Simulators are now parametrized by `MeasureMethod`, with default measurement method that implements MBQC. It allows customized measurement methods, for instance for delegated QC protocols.

See examples of usage of these features for verification of delegated MBQC quantum computations in https://github.com/qat-inria/veriphix .


EarlMilktea:
@jemappellesami We're now enforcing Ruff CI check. Could you merge `master` into the branch?

graphix/clifford.py
```diff
@@ -287,6 +293,17 @@ def measure(self, pauli: graphix.pauli.Pauli) -> graphix.pauli.Pauli:
...
+    @classmethod
+    def __get_pydantic_core_schema__(
+        cls, source_type: typing.Any, handler: pydantic.GetCoreSchemaHandler
+    ) -> pydantic_core.CoreSchema:
+        def check_clifford(obj) -> Clifford:
+            if not isinstance(obj, Clifford):
+                raise ValueError("Clifford expected")
+            return obj
+
+        return pydantic_core.core_schema.no_info_plain_validator_function(function=check_clifford)
```
shinich1, line 305: are we sure we want to use this, it does not seem to be the most stable feature:
https://docs.pydantic.dev/latest/concepts/types/#customizing-validation-with-__get_pydantic_core_schema__

shinich1, line 305: are we sure we want to do this? does not seem to be the most stable feature of pydantic, see https://docs.pydantic.dev/latest/concepts/types/#customizing-validation-with-__get_pydantic_core_schema__

graphix/sim/graphix.code-workspace
shinich1, line 1: please remove this file.

graphix/states.py
```diff
@@ -31,6 +33,17 @@ def get_densitymatrix(self) -> npt.NDArray:
...
+    @classmethod
+    def __get_pydantic_core_schema__(
+        cls, source_type: typing.Any, handler: pydantic.GetCoreSchemaHandler
+    ) -> pydantic_core.CoreSchema:
+        def check_state(obj) -> State:
+            if not isinstance(obj, State):
+                raise ValueError("State expected")
+            return obj
+
+        return pydantic_core.core_schema.no_info_plain_validator_function(function=check_state)
```
shinich1, line 45: same as above comment.

graphix/sim/base_backend.py
```diff
@@ -1,10 +1,77 @@
...
+    from graphix.pauli import Plane
+
+
+@dataclass
+class MeasurementDescription:
+    plane: Plane
+    angle: float
+
+
+class NodeIndex:
```
shinich1, line 26: what's wrong with a `list` as `node_index`, as in previous implementation? Perhaps it is specific to BQC (which is fine)? I'd appreciate it if you could a few lines of description to this class.

graphix/sim/base_backend.py
```diff
@@ -34,31 +101,115 @@ def perform_measure(
...
-    def __init__(self, pr_calc: bool = True, rng: np.random.Generator | None = None):
+    def __init__(
+        self, state: State, node_index: NodeIndex | None = None, pr_calc: bool = True, rng: np.random.Generator = None
+    ):
         """
         Parameters
         ----------
             pr_calc : bool
                 whether or not to compute the probability distribution before choosing the measurement result.
                 if False, measurements yield results 0/1 with 50% probabilities each.
```
shinich1, line 112: description for state and node_index?

graphix/simulator.py
```diff
@@ -6,29 +6,77 @@
...
+    def get_measurement_description(self, cmd: BaseM) -> MeasurementDescription: ...
+
+    @abc.abstractmethod
+    def get_measure_result(self, node: int) -> bool: ...
+
+    @abc.abstractmethod
+    def set_measure_result(self, node: int, result: bool) -> None: ...
+
+
+class DefaultMeasureMethod(MeasureMethod):
```
shinich1, line 41: Could you add description of this class, esp. why we need to have this in the first place? I believe this is for generalization (e.g. client-server setup), but it is hard to understand why we have this without such context, as this looks (at first sight) a rather unnecessary wrapping of measurement command.

shinich1 (changes requested):
My aplogies for the delay reviewing this!

graphix/simulator.py
jemappellesami: It is exactly for that reason (see 4th point in the description of the PR): this class is to be overwritten by a custom measurement method in the case of a client-server setup. You are right we should add a comment about it in the code.

graphix/clifford.py
thierry-martinez: This change allows the `Clifford` class to be used as a type in `pydantic` models: in this PR, we changed the `cliff_index` parameter of type `int` in `C` commands into a `clifford` parameter of type `Clifford`. I think that having a parameter of type `Clifford` instead of a raw index is a good thing. However, another option is to enable the config flag `arbitrary_types_allowed` in the `C` command (https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.arbitrary_types_allowed): I don't know if it is more stable, but it is simpler indeed.

EarlMilktea: > `Clifford` class to be used as a type in pydantic models

Why is it necessary? I feel it's very unlikely that `Clifford` should always be monitored by `pydantic` validation.

EarlMilktea: Even if it were to be really necessary, we should not rely on internal APIs as @shinich1 insists.

graphix/pauli.py
```diff
@@ -108,6 +114,15 @@ class Axis(enum.Enum):
...
     Z = 2
 
+    @property
+    def op(self) -> npt.NDArray:
+        if self == Axis.X:
+            return graphix.ops.Ops.x
+        elif self == Axis.Y:
+            return graphix.ops.Ops.y
+        elif self == Axis.Z:
+            return graphix.ops.Ops.z
```
EarlMilktea, line 124:
```suggestion
        if self == Axis.X:
            return graphix.ops.Ops.x
        if self == Axis.Y:
            return graphix.ops.Ops.y
        if self == Axis.Z:
            return graphix.ops.Ops.z
        typing_extensions.assert_never(self)
```

graphix/sim/base_backend.py
```diff
@@ -1,10 +1,104 @@
...
+    def __getitem__(self, index: int) -> int:
+        return self.__list[index]
+
+    def index(self, node: int) -> int:
+        """
+        Returns the qubit index associated with the specified node label.
+        """
+        return self.__dict[node]
+
+    def __iter__(self) -> collections.abc.Iterator[int]:
```
EarlMilktea, line 60: Please add the corresponding `import`.
(I personally prefer single `Iterable` though)

graphix/sim/base_backend.py
```diff
@@ -34,31 +128,121 @@ def perform_measure(
...
+                move_from = self.node_index.index(ind)
+                self.state.swap((i, move_from))
+                self.node_index.swap(i, move_from)
+
+    def finalize(self, output_nodes) -> None:
+        """to be run at the end of pattern simulation."""
+        self.sort_qubits(output_nodes)
+
+    @property
+    def Nqubit(self) -> int:
```
EarlMilktea, line 247: How about changing the name to PEP8 compliant one? We someday need to do so.

graphix/sim/density_matrix.py
```diff
@@ -77,19 +73,30 @@ def check_size_consistency(mat):
...
                         return
                 except TypeError:
                     pass
         statevec = Statevec(data, nqubit)
         # NOTE this works since np.outer flattens the inputs!
         self.rho = np.outer(statevec.psi, statevec.psi.conj())
-        self.Nqubit = len(statevec.dims())
+
+    @property
+    def Nqubit(self) -> int:
```
EarlMilktea, line 84: Same as above

graphix/sim/density_matrix.py
```diff
@@ -77,19 +73,30 @@ def check_size_consistency(mat):
...
+        return self.rho.shape[0].bit_length() - 1
+
+    def copy(self) -> DensityMatrix:
+        result = self.__new__(self.__class__)
```
EarlMilktea, line 88: Same as above

graphix/sim/density_matrix.py
```diff
@@ -77,19 +73,30 @@ def check_size_consistency(mat):
...
+        result.rho = self.rho
+        return result
 
     def __repr__(self):
```
EarlMilktea, line 92: This method is not valid as `__repr__`. Please use `__str__`.

graphix/sim/density_matrix.py
```diff
@@ -77,19 +73,30 @@ def check_size_consistency(mat):
...
-        return f"DensityMatrix object, with density matrix {self.rho} and shape {self.dims()}."
+        return f"DensityMatrix object, with density matrix {self.rho} and shape{self.dims()}."
+
+    def add_nodes(self, nqubit, data) -> None:
```
EarlMilktea, line 95: Please do not omit type annotation.

graphix/sim/density_matrix.py
```diff
@@ -172,20 +179,19 @@ def expectation_single(self, op, i):
         if op.shape != (2, 2):
             raise ValueError("op must be 2x2 matrix.")
 
-        st1 = deepcopy(self)
```
EarlMilktea, line 175: I feel it's more idiomatic to use `deepcopy`.

graphix/sim/density_matrix.py
```diff
@@ -208,9 +213,9 @@ def cnot(self, edge):
                 Edge to apply CNOT gate.
         """
 
-        self.evolve(CNOT_TENSOR.reshape(4, 4), edge)
+        self.evolve(graphix.sim.statevec.CNOT_TENSOR.reshape(4, 4), edge)
```
EarlMilktea, line 216: Too long.

graphix/sim/density_matrix.py
```diff
@@ -296,134 +304,29 @@ def apply_channel(self, channel: KrausChannel, qargs):
...
-            self.state = DensityMatrix(nqubit=0)
-        n = len(nodes)
-        dm_to_add = DensityMatrix(nqubit=n, data=input_state)
-        self.state.tensor(dm_to_add)
-        self.node_index.extend(nodes)
-        self.Nqubit += n
-
-    def entangle_nodes(self, edge):
-        """Apply CZ gate to the two connected nodes.
+    def __init__(self, pr_calc=True):
```
EarlMilktea, line 326:
```suggestion
    def __init__(self, pr_calc=True) -> None:
```

graphix/sim/statevec.py
```diff
@@ -4,186 +4,21 @@
...
-    elif plane == graphix.pauli.Plane.YZ:
-        vec = (0, np.cos(angle), np.sin(angle))
-    elif plane == graphix.pauli.Plane.XZ:
-        vec = (np.cos(angle), 0, np.sin(angle))
-    op_mat = np.eye(2, dtype=np.complex128) / 2
-    for i in range(3):
-        op_mat += (-1) ** (choice) * vec[i] * CLIFFORD[i + 1] / 2
-    op_mat = CLIFFORD[CLIFFORD_CONJ[vop]] @ op_mat @ CLIFFORD[vop]
-    return op_mat
+    def __init__(self, pr_calc=True, rng: np.random.Generator | None = None):
```
EarlMilktea, line 20:
```suggestion
    def __init__(self, pr_calc=True, rng: np.random.Generator | None = None) -> None:
```

graphix/sim/statevec.py
```diff
@@ -285,7 +120,16 @@ def __init__(
     def __repr__(self):
         return f"Statevec object with statevector {self.psi} and length {self.dims()}."
 
-    def evolve_single(self, op: npt.NDArray, i: int):
+    def copy(self) -> Statevec:
```
EarlMilktea, line 123: Same as above.

graphix/sim/statevec.py
```diff
@@ -312,17 +156,17 @@ def evolve(self, op: np.ndarray, qargs: list[int]):
...
         shape = [2 for _ in range(2 * op_dim)]
         op_tensor = op.reshape(shape)
-        self.psi = np.tensordot(
+        psi = np.tensordot(
             op_tensor,
             self.psi,
             (tuple(op_dim + i for i in range(len(qargs))), tuple(qargs)),
         )
-        self.psi = np.moveaxis(self.psi, [i for i in range(len(qargs))], qargs)
+        self.psi = np.moveaxis(psi, [i for i in range(len(qargs))], qargs)
```
EarlMilktea, line 164: Is it really necessary to pass as `list`?

graphix/sim/tensornet.py
```diff
@@ -1,22 +1,25 @@
...
 import quimb.tensor as qtn
-import typing_extensions
 from quimb.tensor import Tensor, TensorNetwork
 
-import graphix.clifford
-from graphix import command
-from graphix.clifford import CLIFFORD, CLIFFORD_CONJ, CLIFFORD_MUL
+import graphix.command
+
+if typing.TYPE_CHECKING:
```
EarlMilktea, line 13: Same as above.

graphix/sim/tensornet.py
```diff
@@ -67,34 +66,36 @@ def __init__(self, pattern, graph_prep="auto", input_state=BasicStates.PLUS, **k
...
         """Add nodes to the network
 
         Parameters
         ----------
         nodes : iterator of int
             index set of the new nodes.
         """
+        if data != BasicStates.PLUS:
+            raise NotImplementedError(
+                "TensorNetworkBackend currently only supports |+> input state (see https://github.com/TeamGraphix/graphix/issues/167 )."
```
EarlMilktea, line 91:
```suggestion
                "TensorNetworkBackend currently only supports |+> input state (see https://github.com/TeamGraphix/graphix/issues/167)."
```

graphix/simulator.py
```diff
@@ -6,29 +6,87 @@
...
+
 class PatternSimulator:
     """MBQC simulator
 
     Executes the measurement pattern.
     """
 
-    def __init__(self, pattern, backend="statevector", noise_model=None, **kwargs):
+    def __init__(
+        self, pattern, backend="statevector", measure_method: MeasureMethod = None, noise_model=None, **kwargs
```
EarlMilktea, line 81: You need to add `| None` to the argument type when you pass `None` as default value.

graphix/simulator.py
```diff
@@ -6,29 +6,87 @@
...
+    ):
```
EarlMilktea, line 82:
```suggestion
    ) -> None:
```

graphix/simulator.py
```diff
@@ -40,43 +98,48 @@ def __init__(self, pattern, backend="statevector", noise_model=None, **kwargs):
...
+        self.set_noise_model(noise_model)
+        self.__pattern = pattern
+        if measure_method is None:
+            measure_method = DefaultMeasureMethod(pattern.results)
+        self.__measure_method = measure_method
 
     @property
-    def results(self):
-        return self.backend.results
+    def pattern(self):
```
EarlMilktea, line 129: Please do not omit return type.

tests/test_graphsim.py
```diff
@@ -38,6 +40,43 @@ def get_state(g) -> Statevec:
     return gstate
 
 
+def meas_op(angle, vop=0, plane=graphix.pauli.Plane.XY, choice=0):
```
EarlMilktea, line 43: Please do not omit type annotation.

tests/test_statevec_backend.py
```diff
@@ -94,4 +96,112 @@ def test_init_fail(self, hadamardpattern, fx_rng: Generator) -> None:
         state = PlanarState(plane=rand_plane[0], angle=rand_angle[0])
         state2 = PlanarState(plane=rand_plane[1], angle=rand_angle[1])
         with pytest.raises(ValueError):
-            StatevectorBackend(hadamardpattern, input_state=[state, state2])
+            StatevectorBackend().add_nodes(hadamardpattern.input_nodes, data=[state, state2])
+
+    def test_clifford(self):
```
EarlMilktea, line 101:
```suggestion
    def test_clifford(self) -> None:
```

EarlMilktea (changes requested):
General comment:

- Please do not omit type annotation, no matter how it is used.
- Please verify that you're surely importing the name up to the required layer: I'm seeing many linter warnings originating from `foo.bar.baz` used without `import foo.bar` (only `import foo`). This is prohibited unless `bar` is imported in `__init__.py` of `foo`.

graphix/clifford.py
thierry-martinez: Fixed in cba7cf3: I finally set the flag `ConfigDict(arbitrary_types_allowed=True)` in the pydantic model for `C` commands, which makes pydantic do the validation with `isinstance` anyway. There is a similar discussion [here](https://github.com/TeamGraphix/graphix/pull/158#discussion_r1722512340).

graphix/simulator.py
shinich1: thanks! this kind of targeted API change would require a simple but complete example demonstrating such use and/or a dedicated tutorial page. could you write up an example?

graphix/pauli.py
```diff
@@ -227,7 +258,20 @@ def matrix(self) -> np.ndarray:
...
+            return BasicStates.PLUS if eigenvalue == 0 else BasicStates.MINUS
+        elif self.symbol == IXYZ.Y:
+            return BasicStates.PLUS_I if eigenvalue == 0 else BasicStates.MINUS_I
+        elif self.symbol == IXYZ.Z:
+            return BasicStates.ZERO if eigenvalue == 0 else BasicStates.ONE
+        # Any state is eigenstate of the identity
+        elif self.symbol == IXYZ.I:
+            return BasicStates.PLUS
+
+    def __repr__(self) -> str:
```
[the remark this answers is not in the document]
thierry-martinez: It should be the case: return values will be of the form `I`, `X`, `Y` or `Z` (declared at top-level in the module), prefixed by `-` and/or by `1j * `.

graphix/sim/density_matrix.py
thierry-martinez: `normalize` already copy the contents of the matrix, so it's preferable just to do a shallow copy here, instead of copying the whole matrix twice (first with `deepcopy`, and just after with `normalize`).

graphix/pauli.py
```diff
@@ -227,7 +258,20 @@ def matrix(self) -> np.ndarray:
         """
         return self.__unit.complex * graphix.clifford.CLIFFORD[self.__symbol.value + 1]
 
-    def __repr__(self):
+    def get_eigenstate(self, eigenvalue=0) -> BasicStates:
+        from graphix.states import BasicStates
```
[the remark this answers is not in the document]
jemappellesami: It can't be inserted at top level due to circular import issues

graphix/sim/base_backend.py
```diff
@@ -1,10 +1,104 @@
...
+
+    def __init__(self) -> None:
+        self.__dict = dict()
+        self.__list = []
+
+    def copy(self) -> NodeIndex:
+        """
+        Creates a shallow copy of the NodeIndex instance, copying both the list and dictionary.
+        """
+        result = self.__new__(self.__class__)
```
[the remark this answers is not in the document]
jemappellesami: Solved it by using `copy.copy(self)` instead. Same is done for other similar comments

graphix/sim/density_matrix.py
jemappellesami: resolved with `copy.copy(self)`

graphix/sim/statevec.py
jemappellesami: resolved with `copy.copy(self)`

graphix/pauli.py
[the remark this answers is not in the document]
EarlMilktea: It is not enough as `repr(...)` is evaluated at the caller's scope.
How about prefixing it with the class name?

graphix/sim/density_matrix.py
EarlMilktea: I see, but then we should use `copy.copy` instead: I cannot see an obvious reason why `.copy` should be provided as a dedicated method.

thierry-martinez: It’s quite common in Python for objects to provide their own copy methods—examples include `set.copy()`, `list.copy()`, and `dict.copy()`. These methods handles the copying process in a way that is specific to the object’s internal structure. Relying on the copy module (`copy.copy()` or `copy.deepcopy()`) can sometimes expose or depend on internal details of the object that we might not want to reveal.

For instance, in the case of `NodeIndex`, the `copy` method is designed to copy one level deeper than a standard shallow copy, ensuring that internal collections are duplicated into new instances rather than being shared between the original and the copy. In the case of back-ends, they currently only contain their state object, but that's an implementation detail that the caller doesn't have to know about.

EarlMilktea: > `set.copy()`, `list.copy()`, and `dict.copy()`

To my understanding, these methods serves as the "leaves" of object trees, which we are closely tied to the C implementations: they are special and inevitable.
For our implementation, they're unnecessary as we're coding in the Python layer.

> designed to copy one level deeper than a standard shallow copy

I feel it isn't conventional or idiomatic.
It might be no problem if deepcopy were to be a free lunch, but sadly not.
We may need to discuss whether slight conciseness deserves unconventionality and computational cost.

shinich1: @thierry-martinez does `normalize` really do deepcopy?  even if it does, I would be up for `copy.copy` here to be clear to anyone that it's a shallow copy. I'm not against `.copy()` method but I personally many not use it to be sure what I am doing where.

shinich1:
@thierry-martinez @EarlMilktea let us resolve the remaining two comments, and merge this soon!

graphix/simulator.py
shinich1: Thanks! tbh, I feel that delegating the explanation and justification of implementation decisions to the package that depends on graphix is a little strange. it is like rewriting some part of numpy for the sake of graphix and justifying by linking graphix example. I would like this repo to be standalone to be maintainable, and I would appreciate it if you could add short example in graphix package without relying on verphix if at all possible.

graphix/sim/density_matrix.py
thierry-martinez: Fixed in e9f2dce. I removed the copy methods from `Statevec`, `DensityMatrix`, and `NodeIndex`, and explicitly used `copy.copy`. I believe that choosing between `copy` and `deepcopy` assumes knowledge of the internals of the classes being copied, but I understand the preference for being explicit.

As for `normalize`, it indeed normalizes the array, but it's similar to a `deepcopy` in the following ways:

    There is no aliasing between the normalized array and the original array.
    The process iterates over each element of the array.

Importantly, there is no need to `deepcopy` before `normalize` (or `evolve`). If `d1` is a shallow copy of `d0` (meaning they share a reference to the same numpy array, i.e., `d0.rho is d1.rho`), then `d1.normalize()` changes `d1.rho` to reference the newly normalized array, while `d0.rho` remains unmodified.

EarlMilktea: Looks good. Thank you!

graphix/pauli.py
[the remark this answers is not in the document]
EarlMilktea: Why not use either...

- `__str__` : it can be freely designed
- use a standard design for `__repr__` : something like `f"{type(self).__name__}({...})"`

EarlMilktea (changes requested):

graphix/pauli.py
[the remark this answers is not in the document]
thierry-martinez: Fixed in 8a1c34d, where I added the prefix `graphix.pauli.`. It’s important to note that the caller's scope is not pertinent here. As you mentioned earlier, it would be bad practice for the caller to execute the result of `repr(...)`, and we cannot make any assumptions about the caller's environment anyway. The Python documentation mentions [_given an appropriate environment_](https://docs.python.org/3/reference/datamodel.html#object.__repr__), meaning that the goal is to easily copy and paste the printed value somewhere else (where we don’t know anything about the scope, just reasonable assumptions).

Overriding `__repr__` instead of defining a separate `as_str` method is very convenient, especially for debugging purposes. If we don’t override `__repr__`, the Pauli object would be printed as `<graphix.pauli.Pauli object at 0x...>`, which is neither evaluable nor readable.

EarlMilktea (approved):

graphix/simulator.py
shinich1: @jemappellesami @thierry-martinez I guess I'll write one up myself..! let us merge this now and we can get on with other PRs!

shinich1 (approved):
@thierry-martinez please resolve conflicts as appropriate and squash and merge. Thanks!

thierry-martinez:
Merged. Thanks!

PR #183, opened 2024-07-25
Migrate to Ruff and update docs

EarlMilktea:

Added badges:

- CI status
- codecov
- Ruff
- isort


shinich1:
@EarlMilktea do we remove black and isort, now that we have Ruff? we can keep them as well.

shinich1 (approved):

EarlMilktea:
I see. Let's migrate to Ruff!

EarlMilktea:
MEMO: I didn't pin ruff version since it's still under active development.

shinich1:
Is ruff's formatter stable? I am slightly worried about CI keep failing just because of ruff version difference.

EarlMilktea:
There may be some breaking changes as it does not reach v1, but as it's designed to be compatible with black, black would also have similar breaking changes at that time.

I added these lines to `CONTRIBUTING.md` to ensure that contributors and CI use the latest version of Ruff.

```bash
# Be sure to use the latest version of ruff
pip install -U ruff
# Sort imports and format
ruff check --select I --fix .
ruff format .
```

EarlMilktea:
And actually, in most cases, Ruff CI does not complain even if input codes are actually formatted by black/isort.

shinich1 (approved):

thierry-martinez (approved):
Looks good to me. Thanks!

Issue #188, opened 2024-07-28
[Bug]: Wrong Pauli measurements for axes `YZ` and `XZ`


thierry-martinez:
Axes conventions have been changed in 81e1a652 for measurements during simulation, but Pauli preprocessing was not updated, leading to inconsistencies for Pauli measurements outside the `XY` axis. The following test currently succeeds only on the `XY` axis but not on `YZ` and `XZ`. 

```python
    @pytest.mark.parametrize("plane", Plane)
    @pytest.mark.parametrize("angle", [0., 0.5, 1., 1.5])
    def test_pauli_measurement_single(self, plane: Plane, angle: float, use_rustworkx: bool = True) -> None:
        pattern = Pattern(input_nodes=[0, 1])
        pattern.add(E(nodes=[0, 1]))
        pattern.add(M(node=0, plane=plane, angle=angle))
        pattern_ref = pattern.copy()
        pattern.perform_pauli_measurements(use_rustworkx=use_rustworkx)
        state = pattern.simulate_pattern()
        state_ref = pattern_ref.simulate_pattern(pr_calc=False, rng=IterGenerator([0]))
        assert np.abs(np.dot(state.flatten().conjugate(), state_ref.flatten())) == pytest.approx(1)
```

Axes conventions are defined in `pauli.py`. The current convention is:
| Plane | cos | sin |
|-|-|-|
| XY | X | Y |
| YZ | Z | Y |
| XZ | Z | X |

The former convention was:
| Plane | cos | sin |
|-|-|-|
| XY | X | Y |
| YZ | Y | Z |
| XZ | X | Z |

PR #189, opened 2024-07-28
Fix #181 and #188: Better pauli nodes recognition

### LINKED ISSUE #181: Add Pauli plane measurements
Currently, Pauli measurements are determined by checking the angle mod $2\pi$, with non-zero possiblity of incorrect determination of Pauli-measured nodes by numerical errors. I believe it's natural to incorporate Pauli measurements as `X`, `Y`, and `Z` planes. See also implementation of rust-based flow-finding routines in https://github.com/TeamGraphix/fastflow/pull/3.

thierry-martinez:
This commit introduces a new method `is_pauli` on `M` commands, which returns a pair `(axis, sign)` if the measure is Pauli, and `None` otherwise.

This method takes an optional parameter `precision` ($\rho$, default: `1e-6`): a measure is considered to be Pauli if the angle is in an interval $[- \rho + k \cdot \frac \pi 2; \rho + k \cdot \frac \pi 2]$, addressing the precision problem raised in #181.

`axis` is computed by using the `cos` and `sin` definitions of `pauli.py`, solving the inconsistency bug reported in #188.

This commit introduces a new `Sign` class to make code manipulating `Plus` and `Minus` to be more readable than when using `bool`.

### DIFF
--- a/graphix/pattern.py
+++ b/graphix/pattern.py
@@ -2021,11 +2021,10 @@ def is_pauli_measurement(cmd: command.Command):
         if the measurement is not in Pauli basis, returns None.
     """
     assert cmd.kind == command.CommandKind.M
-    pauli = cmd.is_pauli()
+    pauli = cmd.is_close_to_pauli()
     if pauli is None:
         return None
-    axis, sign = pauli
-    return f"{sign}{axis.name}"
+    return f"{pauli.sign}{pauli.axis.name}"
 
 
 def cmd_to_qasm3(cmd):
--- a/graphix/pattern.py
+++ b/graphix/pattern.py
@@ -8,13 +8,12 @@ from copy import deepcopy
 from dataclasses import dataclass
 
 import networkx as nx
-import numpy as np
 import typing_extensions
 
 import graphix.clifford
 import graphix.pauli
 from graphix import command
-from graphix.clifford import CLIFFORD_CONJ, CLIFFORD_MEASURE, CLIFFORD_TO_QASM3
+from graphix.clifford import CLIFFORD_CONJ, CLIFFORD_TO_QASM3
 from graphix.device_interface import PatternRunner
 from graphix.gflow import find_flow, find_gflow, get_layers
 from graphix.graphsim.graphstate import GraphState
@@ -2024,54 +2023,11 @@ def is_pauli_measurement(cmd: command.Command, ignore_vop=True):
         if the measurement is not in Pauli basis, returns None.
     """
     assert cmd.kind == command.CommandKind.M
-    basis_str = [("+X", "-X"), ("+Y", "-Y"), ("+Z", "-Z")]
-    # first item: 0, 1 or 2. correspond to choice of X, Y and Z
-    # second item: 0 or 1. correspond to sign (+, -)
-    basis_index = (0, 0)
-    if np.mod(cmd.angle, 2) == 0:
-        if cmd.plane == graphix.pauli.Plane.XY:
-            basis_index = (0, 0)
-        elif cmd.plane == graphix.pauli.Plane.YZ:
-            basis_index = (1, 0)
-        elif cmd.plane == graphix.pauli.Plane.XZ:
-            basis_index = (0, 0)
-        else:
-            raise ValueError("Unknown measurement plane")
-    elif np.mod(cmd.angle, 2) == 1:
-        if cmd.plane == graphix.pauli.Plane.XY:
-            basis_index = (0, 1)
-        elif cmd.plane == graphix.pauli.Plane.YZ:
-            basis_index = (1, 1)
-        elif cmd.plane == graphix.pauli.Plane.XZ:
-            basis_index = (0, 1)
-        else:
-            raise ValueError("Unknown measurement plane")
-    elif np.mod(cmd.angle, 2) == 0.5:
-        if cmd.plane == graphix.pauli.Plane.XY:
-            basis_index = (1, 0)
-        elif cmd.plane == graphix.pauli.Plane.YZ:
-            basis_index = (2, 0)
-        elif cmd.plane == graphix.pauli.Plane.XZ:
-            basis_index = (2, 0)
-        else:
-            raise ValueError("Unknown measurement plane")
-    elif np.mod(cmd.angle, 2) == 1.5:
-        if cmd.plane == graphix.pauli.Plane.XY:
-            basis_index = (1, 1)
-        elif cmd.plane == graphix.pauli.Plane.YZ:
-            basis_index = (2, 1)
-        elif cmd.plane == graphix.pauli.Plane.XZ:
-            basis_index = (2, 1)
-        else:
-            raise ValueError("Unknown measurement plane")
-    else:
+    pauli = cmd.is_pauli()
+    if pauli is None:
         return None
-    if not ignore_vop:
-        basis_index = (
-            CLIFFORD_MEASURE[cmd.vop][basis_index[0]][0],
-            int(np.abs(basis_index[1] - CLIFFORD_MEASURE[cmd.vop][basis_index[0]][1])),
-        )
-    return basis_str[basis_index[0]][basis_index[1]]
+    axis, sign = pauli
+    return f"{sign}{axis.name}"
 
 
 def cmd_to_qasm3(cmd):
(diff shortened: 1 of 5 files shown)


EarlMilktea:
@thierry-martinez 

@shinich1 is now planning to provide a new class or enum variant for Pauli measurements.

thierry-martinez:
This PR uses `tuple[Axis, Sign]` as the type for Pauli measurements. I am open to declaring a dataclass instead of a tuple if preferred, but I believe that any choice will be isomorphic to this tuple type. If the plan is to have a dedicated `Command` subclass for Pauli measurements (either a child class or a sister class of `M`), I think the approach of this PR—having a general `M` class and a partial coercion method `is_pauli`—is the right choice. I don't see any good way to prevent the `M` class from being instantiated with Pauli angles, so the `Command` class level does not seem to be the right place to distinguish between general and Pauli measurements.

EarlMilktea:
@thierry-martinez 

I see. I'm almost agree with your approach, but have some opinions for the API.

- The name `is_pauli` is not good: we can only say that it is close to Pauli.
- If you still want to use the name `is_pauli`, we should not make `precision` optional.
- Implicit casting to Pauli should be strictly banned.

EarlMilktea:
> I don't see any good way to prevent the M class from being instantiated with Pauli angles

So do I, but at the same time I don't want it to be treated as special Pauli measurements unless explicitly specified by users.

EarlMilktea:
> expect Pauli preprocessing to handle this measurement if angle is (close to) Pauli

I feel this approach is a bit confusing.
For example, Pauli flow requires measurement angles to be exactly equal to $0$, $\pi / 2$, $\pi$, or $3\pi / 2$.
It would be better to require users to cast to dedicated classes/variants explicitly when infinite precision is required, or only do it when angle is given as `Fracion`.

graphix/pattern.py
```diff
@@ -2003,7 +2002,7 @@ def pauli_nodes(pattern: Pattern, leave_input: bool):
     return pauli_node, non_pauli_node
 
 
-def is_pauli_measurement(cmd: command.Command, ignore_vop=True):
+def is_pauli_measurement(cmd: command.Command):
```
EarlMilktea, line 2005: Let's add return type annotation (would be `Literal[XXX] | None`)

graphix/pauli.py
```diff
@@ -19,6 +21,58 @@ class IXYZ(enum.Enum):
...
+    def __neg__(self) -> Sign:
+        return Sign.minus_if(self == Sign.Plus)
+
+    @typing.overload
+    def __mul__(self, other: Sign) -> Sign: ...
+
+    @typing.overload
+    def __mul__(self, other: Number) -> Number: ...
+
+    def __mul__(self, other):
```
EarlMilktea, line 54: Please do not omit type annotations even for the base implementation.

graphix/pauli.py
```diff
@@ -19,6 +21,58 @@ class IXYZ(enum.Enum):
...
+        if isinstance(other, Sign):
+            return Sign.plus_if(self == other)
+        if isinstance(other, Number):
+            return self.value * other
+        return NotImplemented
+
+    def __rmul__(self, other) -> Number | type(NotImplemented):
```
EarlMilktea, line 61: `type(NotImplemented)` may be unnecessary as users never see it (results in exception if not implemented).

graphix/pauli.py
```diff
@@ -69,32 +120,32 @@ def prefix(self, s: str) -> str:
...
-            return COMPLEX_UNITS[self.__sign == self.__im][not self.__im]
+            return COMPLEX_UNITS[self.__sign == Sign.minus_if(self.__im)][not self.__im]
 
     def __neg__(self):
-        return COMPLEX_UNITS[not self.__sign][self.__im]
+        return COMPLEX_UNITS[self.__sign == Sign.Plus][self.__im]
 
 
-COMPLEX_UNITS = [[ComplexUnit(sign, im) for im in (False, True)] for sign in (False, True)]
+COMPLEX_UNITS = [[ComplexUnit(sign, im) for im in (False, True)] for sign in (Sign.Plus, Sign.Minus)]
```
EarlMilktea, line 148: Let's delete this line to prevent unexpected cache mutation.

graphix/pauli.py
```diff
@@ -27,7 +81,7 @@ class ComplexUnit:
     with Python constants 1, -1, 1j, -1j, and can be negated.
     """
 
-    def __init__(self, sign: bool, im: bool):
```
EarlMilktea, line 30: I prefer `is_imag` here.

EarlMilktea (changes requested):

thierry-martinez:
> I feel this approach is a bit confusing. For example, Pauli flow requires measurement angles to be exactly equal to 0 , π / 2 , π , or 3 π / 2 . It would be better to require users to cast to dedicated classes/variants explicitly when infinite precision is required, or only do it when angle is given as `Fracion`.

There is no `Fraction` equal to π / 2! ;-) In numerical simulations, it’s often unavoidable to rely on comparisons with a tolerance of ±ε due to the limitations of floating-point arithmetic. This approach is not unique to Pauli flow but is a general necessity whenever exact matches to theoretical results in the complex field aren’t feasible: Graphix simulation back-ends do that all the time!

graphix/pauli.py
thierry-martinez: Are you suggesting that the base implementation should have the following signature?
```python
def __mul__(self, other: Sign | Number) -> Sign | Number:
```
While I understand the intent, I’m not sure this is the best approach. It introduces repetition in type annotations and can obscure the relationship between the input type and the return type. An alternative could be:
```
SignNumber = TypeVar("SignNumber", Sign | Number)
def __mul__(self, other: SignNumber) -> SignNumber:
```
However, I believe this might complicate maintenance rather than simplify it. According to [PEP 484](https://peps.python.org/pep-0484/), type annotations are often omitted in the actual implementation (for example, see the implementation of `def utf8`).

graphix/pattern.py
thierry-martinez: I changed `measure_pauli` to use `Pauli` directly instead of relying on a string encoding in 654f300: the return type is now just `Pauli`.

graphix/pauli.py
thierry-martinez: Actually, `NotImplemented` may be seen if the user call `__rmul__` directly, but what I didn't know is that `NotImplemented` is an empty type for `mypy` (and `type(NotImplemented)` is just unsupported), so we never need to add explicitly `NotImplemented` in types.

thierry-martinez: Fixed in 3884d45.

graphix/pauli.py
thierry-martinez: Fixed in 453698a.

graphix/pauli.py
thierry-martinez: These global values are needed. I changed them for tuples instead of lists in 57078e7 to prevent unexpected mutations.

EarlMilktea:
> There is no Fraction equal to π / 2!

We're now excluding $\pi$ from the angles, aren't we?

> simulation back-ends do that all the time

Then we should gradually fix it...

Anyway, as long as conversion to Pauli is explicit, this PR looks good to me.

graphix/pauli.py
EarlMilktea: Why necessary? 
We may need to refactor the class using the singleton pattern.

graphix/pauli.py
EarlMilktea: > It introduces repetition in type annotations and can obscure the relationship between the input type and the return type.

I'm suggesting this fix to enable mypy check for the implementation body.
If you were to omit it, you would lose a chance to find hidden bugs, as `other` is now untyped.
I know this approach is not perfect and causes repetition, but still necessary to improve code quality.

EarlMilktea: By the way, mypy is complaining that `Operator "*" not supported for types "Literal[1, -1]" and "Number"
`.

thierry-martinez:
> We're now excluding π from the angles, aren't we?

You're correct, but since Pauli angles have exact float representations (e.g., $0.5 = 2^{-1}$), we could argue that it's safe to use `==` for comparisons with 0, 0.5, 1, and 1.5. Would you prefer to revert to using float equality in this case?

EarlMilktea:
How about discussing later, together with the angle convention mentioned in #191 ?

thierry-martinez:
I propose closing this PR to continue the discussion on #181. I will open a new PR specifically for #188.

PR #191, opened 2024-07-30
Add Open graph class 

wlcsm:
**Context:** Many researchers in MBQC use a graphical approach to reasoning about patterns. These center around the concept of a open graph which can be thought of as a graphical description of an MBQC pattern with inputs and outputs

**Description of the change:** Added the Open graph class together with functionality for converting to and from pyzx diagrams.


graphix/open_graph.py
```diff
@@ -0,0 +1,379 @@
...
+    """An MBQC measurement.
+
+    :param angle: the angle of the measurement. Should be between [0, 2pi)
+    :param plane: the measurement plane: 'XY', 'XZ', 'YZ'
+    """
+
+    angle: float
+    plane: str
+
+    def __eq__(self, other):
```
EarlMilktea, line 24: Please add type annotations whenever possible.

tests/test_open_graph.py
```diff
@@ -0,0 +1,74 @@
...
+
+import networkx as nx
+import numpy as np
+import pytest
+import pyzx as zx
+
+from graphix.open_graph import Measurement, OpenGraph
+
+
+def test_graph_no_output_measurements():
```
EarlMilktea, line 11: Please do not omit `-> None`.

requirements.txt
```diff
@@ -9,4 +9,5 @@ pydantic
 quimb>=1.4.0
 scipy
 sympy>=1.9
+pyzx==0.8.0
```
EarlMilktea, line 12: Is it really necessary to pin the version?

graphix/open_graph.py
```diff
@@ -0,0 +1,379 @@
...
+        -------
+        >>> from graphix.open_graph import Measurement
+        >>> Measurement(0.0, "XY") == Measurement(0.0, "XY")
+        True
+        >>> Measurement(0.0, "XY") == Measurement(0.0, "YZ")
+        False
+        >>> Measurement(0.1, "XY") == Measurement(0.0, "XY")
+        False
+        """
+        return np.allclose(self.angle, other.angle) and self.plane == other.plane
```
EarlMilktea, line 37: Please `return NotImplemented` when comparison does not make sense.

graphix/open_graph.py
```diff
@@ -0,0 +1,379 @@
...
+    # Input nodes:
+    # * "is_input"    - The value is always True
+    # * "input_order" - A zero-indexed integer used to preserve the ordering of
+    #                   the inputs
+    #
+    # Output nodes:
+    # * "is_output"    - The value is always True
+    # * "output_order" - A zero-indexed integer used to preserve the ordering of
+    #                    the outputs
+    inside: nx.Graph
```
EarlMilktea, line 98: Please do not expose data members if possible.

graphix/open_graph.py
```diff
@@ -0,0 +1,379 @@
...
+        >>> inputs = [0]
+        >>> outputs = [2]
+        >>>
+        >>> og = OpenGraph(g, measurements, inputs, outputs)
+        >>> og.inputs == inputs
+        True
+        """
+        inputs = [i for i in self.inside.nodes(data=True) if "is_input" in i[1]]
+        sorted_inputs = sorted(inputs, key=lambda x: x[1]["input_order"])
+        input_node_ids = [i[0] for i in sorted_inputs]
```
EarlMilktea, line 314: Personally I prefer to return `[...]` directly.

graphix/open_graph.py
```diff
@@ -0,0 +1,379 @@
...
+        ...     0: Measurement(0.0, "XY"),
+        ...     1: Measurement(0.5, "XY"),
+        ... }
+        True
+        """
+        return {n[0]: n[1]["measurement"] for n in self.inside.nodes(data=True) if "measurement" in n[1]}
+
+    def perform_z_deletions_in_place(self):
+        """Removes the Z-deleted nodes from the graph in place"""
+        z_measured_nodes = [
```
EarlMilktea, line 366: Please use genexp if you're using it only for iteration.

tests/test_open_graph.py
```diff
@@ -0,0 +1,74 @@
...
+    inputs = [0]
+    outputs = [1]
+
+    # Output node can not be measurement
+    with pytest.raises(ValueError):
+        OpenGraph(g, meas, inputs, outputs)
+
+
+def test_graph_equality():
+    file = "./tests/circuits/adder_n4.qasm"
```
EarlMilktea, line 23: Please use `pytest.fixture` for test data, if possible.

EarlMilktea (changes requested):

requirements.txt
wlcsm: PyZX breaks very frequently, so I believe it is better to pin the version to prevent troubles in future

graphix/open_graph.py
wlcsm: I can't do that here because I am removing nodes later on. If I use a genexp then it will complain that the size of `.inside.nodes` is changing during iteration.

wlcsm:
Hi @EarlMilktea , thanks for your review!
I've replied to a couple of your suggestions to get your opinion first and I've implemented the rest of the suggestions and pushed it to this PR.

Could you have a look at my responses please?

requirements.txt
EarlMilktea: @shinich1 

Is it really acceptable to include `pyzx` as a new dependency?

graphix/open_graph.py
```diff
@@ -0,0 +1,383 @@
...
+
+        g1 = self.perform_z_deletions()
+        g2 = other.perform_z_deletions()
+
+        return (
+            g1.inputs == g2.inputs
+            and g1.outputs == g2.outputs
+            and nx.utils.graphs_equal(g1._inside, g2._inside)
+            and g1.measurements == g2.measurements
+        )
```
shinich1, line 116: I wonder what's the purpose of `perform_z_deletions` here. Does this become important in the context of FBQC you're working on (eg you end up having many Z measurements in the graph)? I get that this an easiest graph simplificaiton we can quickly implement, but it feels a little strange to include only this and not the other transformations like local complementation. If `perform_z_deletions` is necessary, we should mention in the docstring that `.. are equal up to the addition of Z-measured nodes`, and/or have separate method which explicitly does 'up to z measured nodes' comparison?

graphix/open_graph.py
```diff
@@ -0,0 +1,383 @@
...
+from copy import deepcopy
+from dataclasses import dataclass
+
+import networkx as nx
+import numpy as np
+import pyzx as zx
+
+
+@dataclass
+class Measurement:
```
shinich1, line 14: why not use `graphix.command.M`? we will soon have Pauli measurement recognition under development: https://github.com/TeamGraphix/graphix/pull/189

requirements.txt
shinich1: I would keep `pyzx` optional (no inclusion in `requirements.txt` and try the import only when using `opengraph` class or equivalent.

graphix/open_graph.py
```diff
@@ -0,0 +1,383 @@
...
+        Example
+        -------
+        >>> import pyzx as zx
+        >>> from graphix.open_graph import OpenGraph
+        >>> circ = zx.qasm("qreg q[2]; h q[1]; cx q[0], q[1]; h q[1];")
+        >>> g = circ.to_graph()
+        >>> og = OpenGraph.from_pyzx_graph(g)
+        """
+        zx.simplify.to_graph_like(g)
+        zx.simplify.full_reduce(g)
```
shinich1, line 233: don't we want to check that `g` is graph_like, rather than performing `full_reduce` automatically? I think it's up to user what type of graph they want to translate (e.g. may not always want to full-reduce)

graphix/open_graph.py
```diff
@@ -0,0 +1,383 @@
...
+        return (
+            g1.inputs == g2.inputs
+            and g1.outputs == g2.outputs
+            and nx.utils.graphs_equal(g1._inside, g2._inside)
+            and g1.measurements == g2.measurements
+        )
+
+    def __init__(
+        self,
+        inside: nx.Graph,
```
shinich1, line 120: what's the motivation to make this a class rather than functions for each directions of transformations, zx-> dataset (nx.graph, in, out, measurements) and (pattern or (nx.graph, in, out, measurements)) -> zx?

shinich1:
thanks for the PR @wlcsm ! Just to understand the direction:

- is the expected workflow ZX -> opengraph-> generate_from_graph -> pattern?
- for the opposite workflow (pattern->opengraph->PYZX) do(n't) we want to check the compatibility with gflow so that we can go pat-ZX-back to pat while maintaining semantics?

requirements.txt
wlcsm: Sure

graphix/open_graph.py
wlcsm: Yes, sorry I think this is a remnant of the benchmarking that we ran before. I will remove it

requirements.txt
wlcsm: If it isn't in the requirements file though, it will be a problem when the unit tests are run. I'm not sure it will even pass the CI pipeline.
I could include it in a different requirements file if that helps? Or explicitly install pyzx in the github workflow pipeline. Though this would be confusing for people who run `tox` without knowing about this

graphix/open_graph.py
wlcsm: One reason is that since these four items (nx.graph, in, out, measurements) are all related and would not often be used separately (at least I can't think of many situations), so I felt it was better to hold them together in a class since it represents one logical object. In our compiler, the Open graph has many more methods and we have functions that take the Open graph as an argument, so it makes sense for us to hold it all together as one object.

Graphix is a bit different than our compiler though. The purpose of putting the open graph definition in graphix is so that we can standardise our interfaces and develop interchangeable components. We could define the interface for open graphs to be the tuple of (nx.graph, in, out, measurement) to simplify things, I think this is probably okay for graphix's purposes at present. Though this way has the disadvantage of our interface being an implicit convention, something new developers may not realise when they start to use our softwares.

What do you think? Do you think we should keep the class or just use the tuple?

graphix/open_graph.py
wlcsm: I have considered it, however I felt that it wasn't a good separation of concerns. From what I understand, `graphix.command.M` is a command defined at the pattern level and hence has methods and attributes useful for this layer of abstraction e.g. correction parameters.
Whereas `Measurement` is more like a tuple (angle, plane) with methods useful for a higher level of abstraction at the level of Open graphs.

Though I'm happy to discuss it more

graphix/open_graph.py
wlcsm: Yes, it is more to do with some of the things we are doing in the compiler.
In this case since we aren't performing any optimisations, we don't need the `perform_z_deletions` methods here so I have removed them.

wlcsm:
My pleasure :)

> is the expected workflow ZX -> opengraph-> generate_from_graph -> pattern?
Yes

> for the opposite workflow (pattern->opengraph->PYZX) do(n't) we want to check the compatibility with gflow so that we can go pat-ZX-back to pat while maintaining semantics?

I don't think this necessary in the open graph -> ZX or ZX -> open graph code. A ZX graph does not require flow. If you want to convert it to a ZX circuit (not a graph) then you will likely use the [`zx.extract_circuit`](https://pyzx.readthedocs.io/en/latest/api.html#pyzx.extract.extract_circuit) function which will check for flow when creating the graph.

requirements.txt
shinich1: There's actually a similar example with `rustworkx`. We include in `requirements-dev.txt` and ask [dev] installation for those contributing (see https://github.com/TeamGraphix/graphix/blob/master/CONTRIBUTING.md). It'll be fine if we add `pyzx` in `requirements-dev`.

@EarlMilktea I understand type annotations can be done similarly to `rustworkx` case, please correct my if I'm wrong about this.

graphix/open_graph.py
```diff
@@ -0,0 +1,361 @@
...
+@dataclass
+class Measurement:
+    """An MBQC measurement.
+
+    :param angle: the angle of the measurement. Should be between [0, 2pi)
+    :param plane: the measurement plane: 'XY', 'XZ', 'YZ'
+    """
+
+    angle: float
+    plane: str
```
shinich1, line 21: plane should be enum. see graphix.pauli.Plane..

graphix/open_graph.py
shinich1: Thanks for the clarification! 

> The purpose of putting the open graph definition in graphix is so that we can standardise our interfaces and develop interchangeable components.

this is indeed a good point, and generic open graph interface would help standardizing the interface. I think current implementation would do the job nicely.

graphix/open_graph.py
shinich1: Indeed they have different purposes. I think I agree.

shinich1:
>> for the opposite workflow (pattern->opengraph->PYZX) do(n't) we want to check the compatibility with gflow so that we can go pat-ZX-back to pat while maintaining semantics?

>I don't think this necessary in the open graph -> ZX or ZX -> open graph code. A ZX graph does not require flow. If you want to convert it to a ZX circuit (not a graph) then you will likely use the [zx.extract_circuit](https://pyzx.readthedocs.io/en/latest/api.html#pyzx.extract.extract_circuit) function which will check for flow when creating the graph.

yes if it's graph <> ZX, there's no flow involved. If we involve patterns, we need to care about flows, and what type (max-delayed, focus, etc). This is what @masa10-f has worked on, which I think we can incorporate as a feature-rich transpilation between open graph and pattern (to be in separate PR). 

separate from above, do we want to implement OpenGraph.to_pattern and wrap `generate_from_graph`?

shinich1:
also - @EarlMilktea @thierry-martinez do you think we should keep these qasm files? 
I get that they're useful for generating a bunch of zx diagrams and that `pyzx` has similar things (might also be good for grapihix' own testing once we have qasm input).

EarlMilktea:
@shinich1 

I personally don't prefer to include `.qasm` files or even `pyzx` due to package quality issues.
They should be replaced with self-hosted implementations, at least in the future.

requirements.txt
wlcsm: Sure. I've added it into the `requirements-dev.txt` file

wlcsm:
> > > for the opposite workflow (pattern->opengraph->PYZX) do(n't) we want to check the compatibility with gflow so that we can go pat-ZX-back to pat while maintaining semantics?
> 
> > I don't think this necessary in the open graph -> ZX or ZX -> open graph code. A ZX graph does not require flow. If you want to convert it to a ZX circuit (not a graph) then you will likely use the [zx.extract_circuit](https://pyzx.readthedocs.io/en/latest/api.html#pyzx.extract.extract_circuit) function which will check for flow when creating the graph.
> 
> yes if it's graph <> ZX, there's no flow involved. Once we involve patterns, we need to care about flows, and what type (flow/gflow/pailiflow, max-delayed, focused, other). This is what @masa10-f has worked on, which is perhaps something we can incorporate later as a feature-rich transpilation between open graph and pattern (to be in separate PR).

Are you saying that you wish to attach information about the flow to the Open Graph? I'm not familiar with the use case here so perhaps we can coordinate more with @masa10-f later on as it definitely sounds like something we can incorporate.

> separate from above, do we want to implement OpenGraph.to_pattern to wrap `generate_from_graph`, as well as write up bare minimum implementation of `OpenGraph.from_pattern` (i.e. just extarct the open graph, noting limitation of it)?

Sure. I can add a `.to_pattern` method. How would `.from_pattern` be implemented? Is there an existing function that converts patterns to the (nx.graph, in, out, meas) tuple?

wlcsm:
> I personally don't prefer to include `.qasm` files or even `pyzx` due to package quality issues. They should be replaced with self-hosted implementations, at least in the future.

I understand that it looks like an unnecessary step to hold qasm files when we are testing the PyZX graphs --- with the code the way it is now, it would be more ideal to replace the qasm with code that directly constructs the graph in the unit test.

The main reason I have included qasm is so that we can in future create tests to verify the correctness of the PyZX -> OpenGraph -> Pattern pipeline by for instance running the QASM and the resulting pattern on simulators and comparing the resulting tensors. If we directly construct the graphs using code instead, then we can't run it on an alternative simulator to compare the results.

Regarding seperating PyZX from the code. I agree that this could be a good decision so that people can use the interface and not have to install PyZX if they don't need it. @EarlMilktea @shinich1 Would it then be better to move the `to_pyzx_graph` and `from_pyzx_graph` methods into a separate file (e.g. `open_graph_pyzx.py`) and make them their own functions?

```
def to_pyzx_graph(g: OpenGraph) -> pyzx.Graph
def to_pyzx_graph(g: pyzx.Graph) -> OpenGraph
```

wlcsm:
@shinich1 I moved the `pyzx` dependancy to `requirements-dev.txt` but now tox is failing in the CI because the `minimum-deps-py` environments don't install the dependencies in `requirements-dev.txt`.

Some ways we could solve this could be to either hardcode installing pyzx in the `tox.ini` file, or alternatively, to make the `minimum-deps-py` environment skip `tests/test_open_graph.py`. What are your thoughts?

shinich1:
yup i was not endoring nor fully denying inclusion. I think we can leave 

> @shinich1 I moved the `pyzx` dependancy to `requirements-dev.txt` but now tox is failing in the CI because the `minimum-deps-py` environments don't install the dependencies in `requirements-dev.txt`.
> 
> Some ways we could solve this could be to either hardcode installing pyzx in the `tox.ini` file, or alternatively, to make the `minimum-deps-py` environment skip `tests/test_open_graph.py`. What are your thoughts?

Good catch, how about doing the follwoing:
https://github.com/TeamGraphix/graphix/blob/364be19de0b1cc3a9be559eb2891f2a2e56ff6b7/tests/test_graphsim.py#L137

shinich1:
> Sure. I can add a .to_pattern method. How would .from_pattern be implemented? Is there an existing function that converts patterns to the (nx.graph, in, out, meas) tuple?

we can, for example, do the following:
https://github.com/TeamGraphix/graphix/blob/364be19de0b1cc3a9be559eb2891f2a2e56ff6b7/graphix/gflow.py#L716-L722

wlcsm:
@shinich1 Thank you for your suggestions, it has made implementing the changes quite smooth.

I have added `OpenGraph.to_pattern` and `OpenGraph.from_pattern` methods.

Regarding the PyZX tests, the `pytest.skip` decorator works but because the `OpenGraph.to_pyzx_graph` and ``OpenGraph.from_pyzx_graph` are methods, we must import pyzx in the file containing the OpenGraph class definition, hence people wouldn't be able to import the OpenGraph without having PyZX installed.

To fix this, I have moved these two methods into a seperate file `graphix/pyzx.py` and their tests into a seperate file as well `tests/test_pyzx.py`. This way people can use OpenGraph and the `OpenGraph.to_pattern` methods without having PyZX installed.

graphix/open_graph.py
```diff
@@ -0,0 +1,248 @@
...
+    #
+    # Input nodes:
+    # * "is_input"    - The value is always True
+    # * "input_order" - A zero-indexed integer used to preserve the ordering of
+    #                   the inputs
+    #
+    # Output nodes:
+    # * "is_output"    - The value is always True
+    # * "output_order" - A zero-indexed integer used to preserve the ordering of
+    #                    the outputs
```
thierry-martinez, line 106: Why do we care about preserving the order of inputs and outputs but not the order of measurements? In the same vein, this representation forgets s-domains and t-domains: I understand that measurement orders and domains are irrelevant if the computation is deterministic but we should take care of not restricting ourselves to this case, at least when defining data structures (whereas it is fine for some algorithms to be specific to a certain case of flow for instance, such data structures should be general). I would prefer if open graphs were isomorphic to standardized patterns, that is if we preserve input orders, output orders, measurement orders and byproducts.

tests/test_open_graph.py
```diff
@@ -0,0 +1,41 @@
...
+
+import networkx as nx
+
+from graphix.open_graph import Measurement, OpenGraph
+from graphix.pauli import Plane
+
+
+# Tests whether an open graph can be converted to and from a pattern and be
+# successfully reconstructed.
+def test_open_graph_to_pattern() -> None:
```
thierry-martinez, line 11: We should enrich `random_objects.py` with random open graph generation (possible naive to begin with), and test against it. We could also add a test with random patterns (for instance, transpiled from random circuits).

thierry-martinez (changes requested):
Thank you. Having such an `OpenGraph` class will be very useful for improving the rest of the API (in particular, for flows). However, the current choices are biased in preserving some information (input and output ordering) and not others (measurement ordering, byproducts). Even if it is fine to restrict some algorithms to particular case of determinism, we should take care of providing general purpose data structures.

graphix/open_graph.py
masa10-f: Hi @thierry-martinez ! I have a little different perspective on your comment.

From the relationship between ZX Calculus and MBQC, isn’t it more natural that open graph class is equivalent to ZX diagram, especially, MBQC-LC form, not to the pattern? If we want to design universal non-Unitary pattern, I think one of the natural ways is to include freedom of correction set design into the open graph, which can be done in a separated module. In fact, some of papers studying non-Unitary computation on MBQC do not use command sequence in their study because pattern is too much information for their purpose. It may degrade the abstractness of open graph. What do you think about it?

graphix/open_graph.py
```diff
@@ -0,0 +1,248 @@
...
+        outputs = pattern.output_nodes
+
+        meas_planes = pattern.get_meas_plane()
+        meas_angles = pattern.get_angles()
+        meas = {node: Measurement(meas_angles[node], meas_planes[node]) for node in meas_angles}
+
+        return cls(g, meas, inputs, outputs)
+
+    def to_pattern(self) -> Pattern:
+        """Converts the Open graph into a pattern."""
```
shinich1, line 238: please specifically state that the conversion only succeeds if either flow, gflow or pauli flow exist, and that the generated pattern is induced by maximally-delayed flow.

graphix/open_graph.py
```diff
@@ -0,0 +1,248 @@
...
+        ...     0: Measurement(0.0, Plane.XY),
+        ...     1: Measurement(0.5, Plane.XY),
+        ... }
+        True
+        """
+        return {n[0]: n[1]["measurement"] for n in self._inside.nodes(data=True) if "measurement" in n[1]}
+
+    @classmethod
+    def from_pattern(cls, pattern: Pattern) -> OpenGraph:
+        """Initialises an OpenGraph based on a measurement pattern"""
```
shinich1, line 222:
```suggestion
        """Initialises an `OpenGraph` object based on the resource-state graph associated with the measurement pattern. """
```

graphix/open_graph.py
shinich1: @thierry-martinez I agree with @masa10-f. the exact form of correction sets are independent from open graph feature to be implemented here (main purpose is to have a common interface between patterns and zx diagrams), and is something to be dealt with in a separate PR.

graphix/open_graph.py
```diff
@@ -0,0 +1,248 @@
...
+
+if TYPE_CHECKING:
+    from graphix.pattern import Pattern
+
+
+@dataclass
+class Measurement:
+    """An MBQC measurement.
+
+    :param angle: the angle of the measurement. Should be between [0, 2pi)
```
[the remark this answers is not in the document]
wlcsm: I will change it to match graphix's convention for now to avoid potential issues with conversions

graphix/open_graph.py
```diff
@@ -0,0 +1,248 @@
...
+        g = nx.Graph()
+        nodes, edges = pattern.get_graph()
+        g.add_nodes_from(nodes)
+        g.add_edges_from(edges)
+
+        inputs = pattern.input_nodes
+        outputs = pattern.output_nodes
+
+        meas_planes = pattern.get_meas_plane()
+        meas_angles = pattern.get_angles()
```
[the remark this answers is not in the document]
wlcsm: Thanks for pointing these out!

graphix/open_graph.py
```diff
@@ -0,0 +1,248 @@
...
+        False
+        >>> Measurement(0.1, Plane.XY) == Measurement(0.0, Plane.XY)
+        False
+        """
+        if not isinstance(other, Measurement):
+            return NotImplemented
+
+        return np.allclose(self.angle, other.angle) and self.plane == other.plane
+
+    def is_z_measurement(self) -> bool:
```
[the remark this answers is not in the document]
wlcsm: I've removed this method as I realised it is more specific to our usecase for FBQC. There is no real need for it to be here

graphix/open_graph.py
```diff
@@ -0,0 +1,248 @@
...
+        True
+        >>> Measurement(0.0, Plane.XY) == Measurement(0.0, Plane.YZ)
+        False
+        >>> Measurement(0.1, Plane.XY) == Measurement(0.0, Plane.XY)
+        False
+        """
+        if not isinstance(other, Measurement):
+            return NotImplemented
+
+        return np.allclose(self.angle, other.angle) and self.plane == other.plane
```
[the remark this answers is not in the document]
wlcsm: Oh I didn't know about `math.isclose`, I'll use that instead.

I agree with your suggestion and have matde this into an `isclose` method instead.

tests/test_open_graph.py
wlcsm: Good idea.
This is a non-trivial task and so I think this would go better in another PR perhaps. I've made an issue to track it if thats fine with you? #196

graphix/open_graph.py
```diff
@@ -0,0 +1,248 @@
...
+    def __eq__(self, other):
+        """Checks the two open graphs are equal
+
+        This doesn't check they are equal up to an isomorphism"""
+
+        return (
+            self.inputs == other.inputs
+            and self.outputs == other.outputs
+            and nx.utils.graphs_equal(self._inside, other._inside)
+            and self.measurements == other.measurements
```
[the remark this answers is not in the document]
wlcsm: Ah good catch! I've removed the comparison of inputs and outputs directly and I've changed the comparison of the measurements to use the `isclose` method

wlcsm:
@shinich1 @thierry-martinez After resolving all your comments, I have simplified the OpenGraph datastructure to simply hold the nx.Graph, inputs, outputs, and measurements separately. This is primarily because it is simpler to understand.
The reason we put in all in one nx.Graph previously was actually to make some operations required by our compiler easier to perform, but I see no need here.

graphix/opengraph.py
```diff
@@ -0,0 +1,165 @@
...
+        >>> from graphix.opengraph import Measurement
+        >>> from graphix.pauli import Plane
+        >>> Measurement(0.0, Plane.XY).isclose(Measurement(0.0, Plane.XY))
+        True
+        >>> Measurement(0.0, Plane.XY).isclose(Measurement(0.0, Plane.YZ))
+        False
+        >>> Measurement(0.1, Plane.XY).isclose(Measurement(0.0, Plane.XY))
+        False
+        """
+        if not isinstance(other, Measurement):
```
EarlMilktea, line 44: `NotImplemented` is a special mechanism for operator overloading. Please do not abuse it.

graphix/opengraph.py
```diff
@@ -0,0 +1,165 @@
...
+import networkx as nx
+
+from graphix.generator import generate_from_graph
+
+if TYPE_CHECKING:
+    from graphix.pattern import Pattern
+    from graphix.pauli import Plane
+
+
+@dataclass
```
EarlMilktea, line 18: Please use `frozen=True` or `typing.NamedTuple` if you don't need mutability.

graphix/opengraph.py
```diff
@@ -0,0 +1,165 @@
...
+    >>>
+    >>> inside_graph = nx.Graph([(0, 1), (1, 2), (2, 0)])
+    >>>
+    >>> measurements = {i: Measurement(0.5 * i, Plane.XY) for i in range(2)}
+    >>> inputs = [0]
+    >>> outputs = [2]
+    >>> og = OpenGraph(inside_graph, measurements, inputs, outputs)
+    """
+
+    _inside: nx.Graph
```
EarlMilktea, line 73: Please use `__inside` unless you're planning to inherit this class.

graphix/opengraph.py
```diff
@@ -0,0 +1,165 @@
...
+    _inputs: list[int]
+    _outputs: list[int]
+    _meas: dict[int, Measurement]
+
+    def __eq__(self, other):
```
EarlMilktea, line 78: Please do not omit type annotation.

graphix/opengraph.py
```diff
@@ -0,0 +1,165 @@
...
+
+        return all(m.isclose(other.measurements[node]) for node, m in self.measurements.items())
+
+    def __init__(
+        self,
+        inside: nx.Graph,
+        measurements: dict[int, Measurement],
+        inputs: list[int],
+        outputs: list[int],
+    ):
```
EarlMilktea, line 100:
```suggestion
    ) -> None:
```

graphix/opengraph.py
```diff
@@ -0,0 +1,165 @@
...
+
+        if set(self.measurements.keys()) != set(other.measurements.keys()):
+            return False
+
+        return all(m.isclose(other.measurements[node]) for node, m in self.measurements.items())
+
+    def __init__(
+        self,
+        inside: nx.Graph,
+        measurements: dict[int, Measurement],
```
EarlMilktea, line 97: Please use `Mapping` and `Sequence` if possible.

EarlMilktea (changes requested):

graphix/opengraph.py
wlcsm: Hey, I'm not sure I understand what you mean. Could you describe the change you'd like to see a little more?

EarlMilktea: This is a rather general remark: if you only needs something that can be indexed by `int`, prefer `Mapping[int, Measurement]` for flexibility.
But anyway, the usage here is actually reasonable.
Sorry for confusing you :cry:

wlcsm: Ah I see. I was unaware of this `collections.abc.Mappings` class. It looks useful for extensibility so I've added it in.

EarlMilktea: 5e7b2aa is not the correct way to use `Mapping`, as it will obscure the concrete type.
My implication was that, to use `Mapping` for the argument type and then cast it to `dict` inside `__init__`.
If you want to make this class itself generic, you need `TypeVar`, but it may be going too further.

EarlMilktea: If you're interested in more details, please see https://github.com/TeamGraphix/fastflow/tree/master/python/fastflow for example.

graphix/opengraph.py
```diff
@@ -0,0 +1,162 @@
...
+        measurements: Mapping[int, Measurement],
+        inputs: list[int],
+        outputs: list[int],
+    ) -> None:
+        """Constructs a new OpenGraph instance
+
+        The inputs() and outputs() methods will preserve the order that was
+        original given in to this methods.
+        """
+        self._inside = inside
```
thierry-martinez, line 103: The field declared above is called `__inside` (private, with two `_`), not `_inside`. Same remark for `__inputs`, `__outputs` and `__meas`.

graphix/opengraph.py
```diff
@@ -0,0 +1,162 @@
...
+    >>> measurements = {i: Measurement(0.5 * i, Plane.XY) for i in range(2)}
+    >>> inputs = [0]
+    >>> outputs = [2]
+    >>> og = OpenGraph(inside_graph, measurements, inputs, outputs)
+    """
+
+    __inside: nx.Graph
+    __inputs: list[int]
+    __outputs: list[int]
+    __meas: Mapping[int, Measurement]
```
thierry-martinez, line 73: Now that inputs, outputs and measures are not stored in the graph, `OpenGraph` can now be declared as a `NamedTuple` with fields `inside`, `outputs` and `meas`, which would simplify the `__init__` implementation.

thierry-martinez (changes requested):
The implementation is much simpler now that datas are no longer stored in the graph, so simpler that I wonder if `OpenGraph` shouldn't just be a `NamedTuple`!

graphix/opengraph.py
wlcsm: Oh I thought the point was to obscure over the concrete type 😅 , so that we may use any object that can be indexed with `int` and return a `Measurement`.

I see how you are using `Mapping` and converting it to a `dict` in the fastflow repo. For simplicity, I think we could stick to `dict` for now. We can always change the code to use `Mapping` in future if there is a need and it will be backwards compatible.

graphix/opengraph.py
wlcsm: Thanks for pointing this out, but I actually already changed these as per EarlMilkTea's suggestion in a previous commit. Not sure why the change didn't display on your side.

graphix/opengraph.py
wlcsm: Though if it in a NamedTuple, then we cannot write a custom `__init__` method, so we can't perform checks at object initialisation time. Currently there is only one check, namely that output nodes aren't measured.

We could get around this by making a classmethod to create new instances of `OpenGraph`, though we would need to keep in mind for all future code review that people should be using this method to create Open graphs and not directly with `OpenGraph(g, in, out, meas)`.

Something like
```python
@classmethod
def from(g: nx.Graph, in: list[int], out: list[int], meas: dict[int, Measurement]) -> OpenGraph:
    # Perform all the checks
    # ....
    return OpenGraph(g, meas, in, out)
```
As a side note, we also can't prefix attributes in a NamedTuple with double underscores so that change would have to be reverted.

wlcsm: So in practice, only a few lines are saved in the `__init__` by using a NamedTuple, though we can remove the getter methods since it is now immutable. I think my main concern would be with having to ensure everyone uses the classmethod for initialisation from here one, which seems unnecessary.
What are your thoughts?

thierry-martinez: The purpose of using double underscores is to protect these fields from being modified externally. However, since `NamedTuple` makes these fields immutable, we wouldn’t need to worry about protecting them, allowing us to eliminate the `__init__` method as well as the `inputs`, `outputs`, and `measurements` properties.

If the intention behind keeping `__init__` (and the other properties) is to enable validation, it might be more appropriate to use a `pydantic.BaseModel` with validators. That said, the current scope of validation seems limited. I’m unsure if it justifies keeping the existing setup. If we decide to maintain validation, it would be beneficial to include additional checks, such as ensuring that inputs, outputs, and measurements are actual graph nodes, and verifying that the graph is self-loop free.

graphix/opengraph.py
thierry-martinez: Are you sure you pushed on the branch `opengraph` associated to this PR? The file `opengraph.py` on the last commit 5e7b2aa of this branch still contains occurrences of `_inline`: https://github.com/TeamGraphix/graphix/blob/5e7b2aab3c403b2da78dacfbad19b776600b42de/graphix/opengraph.py#L103

shinich1: @wlcsm I agree with @thierry-martinez, it seems like your changes are not on remote repo. I unresolved correpsonding @EarlMilktea 's somment. Could you try to push the changes again?

graphix/pyzx.py
```diff
@@ -0,0 +1,164 @@
...
+import networkx as nx
+import pyzx as zx
+
+from graphix.opengraph import Measurement, OpenGraph
+from graphix.pauli import Plane
+
+
+def to_pyzx_graph(og: OpenGraph) -> zx.graph.base.BaseGraph:
+    """Return a PyZX graph corresponding to the the open graph.
+
```
shinich1, line 18:
```suggestion
        .. warning::
            works with `pyzx==0.8.0` (see `requirements-dev.txt`). Other versions may not be compatible due to breaking changes
```

graphix/pyzx.py
```diff
@@ -0,0 +1,164 @@
...
+
+def from_pyzx_graph(g: zx.graph.base.BaseGraph) -> OpenGraph:
+    """Constructs an Optyx Open Graph from a PyZX graph.
+
+    This method may add additional nodes to the graph so that it adheres
+    with the definition of an OpenGraph. For instance, if the final node on
+    a qubit is measured, it will add two nodes behind it so that no output
+    nodes are measured to satisfy the requirements of an open graph.
+
```
shinich1, line 93:
```suggestion
        .. warning::
            works with `pyzx==0.8.0` (see `requirements-dev.txt`). Other versions may not be compatible due to breaking changes
```

graphix/opengraph.py
shinich1: >  If we decide to maintain validation, it would be beneficial to include additional checks, such as ensuring that inputs, outputs, and measurements are actual graph nodes, and verifying that the graph is self-loop free.

@wlcsm could you add these? then we should be fine to resolve this conversation!

shinich1 (approved):
@wlcsm Thanks for the updates! For me, this can be merged after resolving the above remaining conversations.

graphix/pyzx.py
```diff
@@ -0,0 +1,164 @@
...
+    -------
+    >>> import networkx as nx
+    >>> g = nx.Graph([(0, 1), (1, 2)])
+    >>> inputs = [0]
+    >>> outputs = [2]
+    >>> measurements = {0: Measurement(0, Plane.XY), 1: Measurement(1, Plane.YZ)}
+    >>> og = OpenGraph(g, measurements, inputs, outputs)
+    >>> reconstructed_pyzx_graph = og.to_pyzx_graph()
+    """
+    g = zx.Graph()
```
shinich1, line 29:
```suggestion
    # check pyzx availability and version
    try:
        import pyzx as zx
    except ModuleNotFoundError as e:
        msg = "Cannot find pyzx (optional dependency)."
        raise RuntimeError(msg) from e
    if zx.__version__ != "0.8.0":
        raise Warning(
            "`to_pyzx_graph` is guaranteed to work only with pyzx==0.8.0 due to possible breaking changes in `pyzx`."
        )
    g = zx.Graph()
```

graphix/pyzx.py
```diff
@@ -0,0 +1,164 @@
...
+
+    Example
+    -------
+    >>> import pyzx as zx
+    >>> from graphix.opengraph import OpenGraph
+    >>> circ = zx.qasm("qreg q[2]; h q[1]; cx q[0], q[1]; h q[1];")
+    >>> g = circ.to_graph()
+    >>> og = OpenGraph.from_pyzx_graph(g)
+    """
+    zx.simplify.to_graph_like(g)
```
shinich1, line 102:
```suggestion
    # check pyzx availability and version
    try:
        import pyzx as zx
    except ModuleNotFoundError as e:
        msg = "Cannot find pyzx (optional dependency)."
        raise RuntimeError(msg) from e
    if zx.__version__ != "0.8.0":
        raise Warning(
            "`from_pyzx_graph` is guaranteed to work only with pyzx==0.8.0 due to possible breaking changes in `pyzx`."
        )
        
    zx.simplify.to_graph_like(g)
```

shinich1 (changes requested):
@wlcsm sorry  - here's my final requests for changes. After resolving these (and remaining requests above), I will approve for merge!

graphix/pyzx.py
EarlMilktea: `raise` makes python abort (rather than warning)

shinich1: ah thanks, yes - changing now

graphix/opengraph.py
wlcsm: Okay yep, I've made it a NamedTuple now. We now don't validate whether no output nodes were measured, but this can be added in later, perhaps with a more complete set of validations

graphix/opengraph.py
wlcsm: Ah my apologies, my local repo wasn't correctly synchronised. It should be fixed now

wlcsm (commented):
Implemented

shinich1:
@wlcsm thanks! one more thing, can we replace these qasm files by `cliffordT` [generator of pyzx](https://pyzx.readthedocs.io/en/latest/api.html#pyzx.generate.cliffordT)?

thierry-martinez (approved):
Looks good to me, thanks!

tests/test_pyzx.py
```diff
@@ -0,0 +1,73 @@
+from __future__ import annotations
+
+import os
+import sys
+
+import pytest
```
shinich1, line 6:
```suggestion
import random
import sys
from copy import deepcopy

import pytest
from pyzx.generate import cliffordT
```

tests/test_pyzx.py
```diff
@@ -0,0 +1,73 @@
...
+
+try:
+    import pyzx as zx
+
+    from graphix.pyzx import from_pyzx_graph, to_pyzx_graph
+except ModuleNotFoundError:
+    pass
+
+
```
shinich1, line 15:
```suggestion
SEED = 123

```

tests/test_pyzx.py
```diff
@@ -0,0 +1,73 @@
...
+@pytest.mark.skipif(sys.modules.get("pyzx") is None, reason="pyzx not installed")
+def test_graph_equality() -> None:
+    file = "./tests/circuits/adder_n4.qasm"
+    circ = zx.Circuit.load(file)
+
+    g = circ.to_graph()
+    og1 = from_pyzx_graph(g)
+
+    g_copy = circ.to_graph()
```
shinich1, line 24:
```suggestion
    random.seed(SEED)
    g = cliffordT(4, 10, 0.1)

    og1 = from_pyzx_graph(g)

    g_copy = deepcopy(g)
```

tests/test_pyzx.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+# Converts a graph to and from an Open graph and then checks the resulting
+# pyzx graph is equal to the original.
+@pytest.mark.skipif(sys.modules.get("pyzx") is None, reason="pyzx not installed")
+def assert_reconstructed_pyzx_graph_equal(circ) -> None:
+    g = circ.to_graph()
+    zx.simplify.to_graph_like(g)
+
+    g_copy = circ.to_graph()
```
shinich1, line 37:
```suggestion
def assert_reconstructed_pyzx_graph_equal(g) -> None:
    zx.simplify.to_graph_like(g)

    g_copy = deepcopy(g)
```

tests/test_pyzx.py
```diff
@@ -0,0 +1,73 @@
...
+    og = from_pyzx_graph(g_copy)
+    reconstructed_pyzx_graph = to_pyzx_graph(og)
+
+    # The "tensorfy" function break if the rows aren't set for some reason
+    for v in reconstructed_pyzx_graph.vertices():
+        reconstructed_pyzx_graph.set_row(v, 2)
+
```
shinich1, line 44:
```suggestion

    for v in g.vertices():
        g.set_row(v, 2)
```

tests/test_pyzx.py
```diff
@@ -0,0 +1,73 @@
...
+    return circuits
+
+
+# Tests that compiling from a pyzx graph to an OpenGraph returns the same
+# graph. Only works with small circuits up to 4 qubits since PyZX's `tensorfy`
+# function seems to consume huge amount of memory for larger qubit
+@pytest.mark.skipif(sys.modules.get("pyzx") is None, reason="pyzx not installed")
+def test_all_small_circuits(all_small_circuits) -> None:
+    for circ in all_small_circuits:
+        assert_reconstructed_pyzx_graph_equal(circ)
```
shinich1, line 73:
```suggestion
def test_random_cliffordT() -> None:
    for i in range(15):
        random.seed(SEED+i)
        g = cliffordT(4, 10, 0.1)
        assert_reconstructed_pyzx_graph_equal(g)
```

tests/test_pyzx.py
```diff
@@ -0,0 +1,73 @@
...
+        filename = os.fsdecode(file)
+        if not filename.endswith(".qasm"):
+            raise Exception(f"only '.qasm' files allowed: not {filename}")
+
+        circ = zx.Circuit.load(direc + filename)
+        circuits.append(circ)
+
+    return circuits
+
+
```
shinich1, line 66:
```suggestion
```

shinich1:
> @wlcsm thanks! one more thing, can we replace these qasm files by `cliffordT` [generator of pyzx](https://pyzx.readthedocs.io/en/latest/api.html#pyzx.generate.cliffordT)? I think this will improve package quality while maintaining tests

implemented as above, please take a look.

wlcsm:
> > @wlcsm thanks! one more thing, can we replace these qasm files by `cliffordT` [generator of pyzx](https://pyzx.readthedocs.io/en/latest/api.html#pyzx.generate.cliffordT)? I think this will improve package quality while maintaining tests
> 
> implemented as above, please take a look.

Hi @shinich1 , thanks for proactively implementing the change!
I realise I neglected to mention that I was on vacation last week and so I unfortunately wasn't quick to respond to your comments.

Your change is great implementing the randomised circuit testing is great. I changed a couple of lines to ensure the pyzx dependency is optional and removed the QASM files. So now I think it is all ready to go

shinich1:
> > > @wlcsm thanks! one more thing, can we replace these qasm files by `cliffordT` [generator of pyzx](https://pyzx.readthedocs.io/en/latest/api.html#pyzx.generate.cliffordT)? I think this will improve package quality while maintaining tests
> > 
> > 
> > implemented as above, please take a look.
> 
> Hi @shinich1 , thanks for proactively implementing the change! I realise I neglected to mention that I was on vacation last week and so I unfortunately wasn't quick to respond to your comments.
> 
> Your change is great implementing the randomised circuit testing is great. I changed a couple of lines to ensure the pyzx dependency is optional and removed the QASM files. So now I think it is all ready to go

Hi @wlcsm, Thanks for the fix! Yes, this is ready to go. 
This is OSS contribution, don't be pressured too much :) Hope you enjoyed your holiday!

shinich1 (approved):

EarlMilktea:
@wlcsm Could you resolve my suggestions please?

wlcsm:
> @wlcsm Could you resolve my suggestions please?

Hey @EarlMilktea, which suggestions are referring to? I can't see any unresolved suggestions

EarlMilktea:
@wlcsm 

It might be collapsed like this:

collapse

wlcsm:
@EarlMilktea When I click to see which changes you have requested here
Screenshot 2024-08-19 at 10 37 17 PM
It takes me to this set of changes which have all been marked as resolved
Screenshot 2024-08-19 at 10 36 42 PM

> It might be collapsed like this:

I've expanded the discussions and I still can't find any unresolved changes 😅 
Sorry for the trouble but would you be able to provide a link to the unresolved suggestions please?

EarlMilktea (changes requested):

EarlMilktea:
@wlcsm 

I forgot to request changes after adding comments! :sob:

graphix/pyzx.py
```diff
@@ -0,0 +1,164 @@
...
+
+    # Open Graph's don't have boundary nodes, so we need to connect the
+    # input and output Z spiders to their corresponding boundary nodes in
+    # pyzx.
+    for pyzx_index, og_index in zip(in_verts, og.inputs):
+        g.add_edge((pyzx_index, map_to_pyzx[og_index]))
+    for pyzx_index, og_index in zip(out_verts, og.outputs):
+        g.add_edge((pyzx_index, map_to_pyzx[og_index]))
+
+    og_edges = og._inside.edges()
```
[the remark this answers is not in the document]
wlcsm: Now that OpenGraph is a NamedTuple, the `inside` attribute is not protected and is just a tuple element

graphix/opengraph.py
```diff
@@ -0,0 +1,162 @@
...
+        if set(self.measurements.keys()) != set(other.measurements.keys()):
+            return False
+
+        return all(m.isclose(other.measurements[node]) for node, m in self.measurements.items())
+
+    def __init__(
+        self,
+        inside: nx.Graph,
+        measurements: Mapping[int, Measurement],
+        inputs: list[int],
```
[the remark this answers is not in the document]
wlcsm: I agree that duplication does not make sense and so it would be nicer to use a `set`. The main reason I have used a list is actually to preserve the ordering of the inputs/outputs.

The ordering of inputs/outputs is required when checking if two OpenGraphs are equal, since the unitary transformation that the graph will be different if the inputs/outputs are in a different order.
There is of course a workaround by storing the inputs/outputs as a `set` and using a seperate data structure to store the ordering which we can then use when comparing the open graphs. However I felt this was too complicated for our current usage. What do you think?

wlcsm:
@EarlMilktea No worries 😄 

We are now using randomised circuits for testing and removed the QASM circuits, so the changes were naturally resolved.

I've resolved your changes except one, please have a look at my reply

graphix/opengraph.py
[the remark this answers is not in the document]
EarlMilktea: Mmm...Then we should use `list`.
Something like `ordered_inputs` may help users recognize the subtle difference.
Also please do not forget to add duplication check and appropriate documentations!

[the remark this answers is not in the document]
wlcsm: Okay great. I have added comments in the docstring and in the code to emphasise that the inputs/outputs are ordered.

We can't add the duplication check at the moment due to using a NamedTuple which doesn't allow us to create our own ``__init__`` method to perform validation. We discussed other options in https://github.com/TeamGraphix/graphix/pull/191#discussion_r1711041039 and decided to omit validation for now in order to maintain simplicity and postpone proper validation to future work.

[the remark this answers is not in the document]
EarlMilktea: How about `dataclass`?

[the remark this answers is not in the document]
wlcsm: It originally was a `dataclass` and we agreed to move to a `NamedTuple` because it simplified many aspects. You can see the entire discussion here https://github.com/TeamGraphix/graphix/pull/191#pullrequestreview-2229643252

graphix/opengraph.py
```diff
@@ -0,0 +1,162 @@
...
+        True
+        >>> Measurement(0.0, Plane.XY).isclose(Measurement(0.0, Plane.YZ))
+        False
+        >>> Measurement(0.1, Plane.XY).isclose(Measurement(0.0, Plane.XY))
+        False
+        """
+        return math.isclose(self.angle, other.angle, rel_tol=rel_tol, abs_tol=abs_tol) and self.plane == other.plane
+
+
+class OpenGraph:
```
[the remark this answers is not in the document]
wlcsm: Because our measurements are stored as floats, there may be OpenGraphs where `__eq__` returns true because `Measurement.isclose` returns true for all the measurements, but the hashes won't be equal because the floats are not exactly the same.

I'm not sure if this may lead to issues in future so I'm not sure if it is the best to implement a naive hash function now unless we think hashing will be required

graphix/opengraph.py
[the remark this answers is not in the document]
thierry-martinez: I am the one responsible for suggesting `NamedTuple` as it seemed like the simplest solution to provide immutable fields. However, I realize now that I could have also suggested using `@dataclass(frozen=True)`. The original code didn't use either `dataclass` or `NamedTuple`, but instead manually initialized private fields and provided read-only properties.

Switching to dataclasses is indeed trivial, and I've made a proposition here: https://github.com/TeamGraphix/graphix/commit/c1d371e9a34069fd1b2b99c802119eee85183c75 . Given that we already use dataclasses in other parts of the graphix project, it makes sense to maintain consistency by adopting @dataclass(frozen=True).

While data validation can be achieved with `NamedTuple` by overriding `__new__`, it would not be as concise or straightforward as using a frozen dataclass.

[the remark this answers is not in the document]
EarlMilktea: c1d371e9a34069fd1b2b99c802119eee85183c75 looks great to me.
Concise and idiomatic!

graphix/opengraph.py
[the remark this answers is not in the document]
EarlMilktea: I see, understood why `__hash__` is missing in the current implementation.
By the way, I strongly believe that `__eq__` should test strict equality, rather than logical equality defined by `math.isclose`.
I remember there was a discussion about that.
Where was it?

graphix/opengraph.py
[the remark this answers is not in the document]
wlcsm: Looks good to me, I've created a commit with these changes and pushed

graphix/opengraph.py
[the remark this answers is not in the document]
wlcsm: We could instead turn the `OpenGraph.__eq__` method into another `isclose` method?
How does that sound?

Testing strict equality of open graphs wouldn't work well due to the angles being implemented as floats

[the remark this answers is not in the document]
thierry-martinez: > I remember there was a discussion about that... Where was it?

We already had a discussion about `__eq__`/`isclose` in the PR about Pauli measures: https://github.com/TeamGraphix/graphix/pull/189#issuecomment-2267505193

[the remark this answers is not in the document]
EarlMilktea: > Testing strict equality of open graphs wouldn't work well due to the angles being implemented as floats

I understand it, but still we had better not hack `__eq__`.
It is really confusing for users and also unconventional.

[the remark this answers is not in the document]
wlcsm: I'm not familiar with the changes made in that PR. To keep consistent with the design decisions in that PR, is it better to make `__eq__` into an `isclose` method for open graphs? What is the way forward here?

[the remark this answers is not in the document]
wlcsm: Okay I've converted this method to be `isclose`.
OpenGraph should get a default implementation of `__eq__` from the dataclass decorator which will compare the attributes as if it was a tuple.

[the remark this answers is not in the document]
wlcsm: @EarlMilktea If all looks good to you, can you resolve this conversation and see if there is anything else to change before we merge?

EarlMilktea (changes requested):
~Looks good!~
~Only one simple suggestion.~

Sorry I was checking outdated codes...

EarlMilktea (approved):

shinich1:
@wlcsm please squash and merge!

shinich1:
will bump pypi version soon.

PR #197, opened 2024-08-09
Fix #188: Fix Pauli measurement conventions

### LINKED ISSUE #188: [Bug]: Wrong Pauli measurements for axes `YZ` and `XZ`

thierry-martinez:
This commit introduces a new named tuple, `graphix.pauli.Measure`, which contains a pair `(axis, sign)`. It includes a static method, `try_from`, that returns an instance of `Measure` if the given plane and angle represent a valid Pauli measure, or `None` otherwise.

The axis is computed using the `cos` and `sin` definitions from `pauli.py`, addressing the inconsistency bug reported in #188.

Additionally, this commit introduces a new `Sign` class, which enhances the readability of code that manipulates `Plus` and `Minus` values, as opposed to using `bool`.

Note: This PR follows up on #189, focusing on fixing #188 while preserving the equality test for recognizing Pauli measures.

### DIFF
--- a/graphix/pattern.py
+++ b/graphix/pattern.py
@@ -17,7 +17,7 @@ from graphix.clifford import CLIFFORD_CONJ, CLIFFORD_TO_QASM3
 from graphix.device_interface import PatternRunner
 from graphix.gflow import find_flow, find_gflow, get_layers
 from graphix.graphsim.graphstate import GraphState
-from graphix.pauli import Axis, PauliMeasure, Plane, Sign
+from graphix.pauli import Axis, PauliMeasurement, Plane, Sign
 from graphix.simulator import PatternSimulator
 from graphix.visualization import GraphVisualizer
 
@@ -1878,7 +1878,7 @@ def measure_pauli(pattern, leave_input, copy=False, use_rustworkx=False):
         new_inputs = pattern.input_nodes
     for cmd in to_measure:
         pattern_cmd: command.Command = cmd[0]
-        measurement_basis: PauliMeasure = cmd[1]
+        measurement_basis: PauliMeasurement = cmd[1]
         # extract signals for adaptive angle.
         s_signal = 0
         t_signal = 0
@@ -1948,7 +1948,7 @@ def measure_pauli(pattern, leave_input, copy=False, use_rustworkx=False):
     return pat
 
 
-def pauli_nodes(pattern: Pattern, leave_input: bool) -> list[tuple[command.M, PauliMeasure]]:
+def pauli_nodes(pattern: Pattern, leave_input: bool) -> list[tuple[command.M, PauliMeasurement]]:
     """returns the list of measurement commands that are in Pauli bases
     and that are not dependent on any non-Pauli measurements
 
@@ -1965,11 +1965,11 @@ def pauli_nodes(pattern: Pattern, leave_input: bool) -> list[tuple[command.M, Pa
     if not pattern.is_standard():
         pattern.standardize()
     m_commands = pattern.get_measurement_commands()
-    pauli_node: list[tuple[command.M, PauliMeasure]] = []
+    pauli_node: list[tuple[command.M, PauliMeasurement]] = []
     # Nodes that are non-Pauli measured, or pauli measured but depends on pauli measurement
     non_pauli_node: set[int] = set()
     for cmd in m_commands:
-        pm = PauliMeasure.try_from(cmd.plane, cmd.angle)  # None returned if the measurement is not in Pauli basis
+        pm = PauliMeasurement.try_from(cmd.plane, cmd.angle)  # None returned if the measurement is not in Pauli basis
         if pm is not None and (cmd.node not in pattern.input_nodes or not leave_input):
             # Pauli measurement to be removed
             if pm.axis == Axis.X:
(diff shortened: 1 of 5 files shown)


graphix/pauli.py
```diff
@@ -284,6 +338,32 @@ def parse(name: str) -> Pauli:
...
+        angle_double = 2 * angle
+        if not is_int(angle_double):
+            return None
+        angle_double_mod_4 = int(angle_double) % 4
+        if angle_double_mod_4 % 2 == 0:
+            axis = plane.cos
+        else:
+            axis = plane.sin
+        sign = Sign.minus_if(angle_double_mod_4 >= 2)
+        return Measure(axis, sign)
```
shinich1, line 364: perhaps differnet class name is good to make clear it's specifically to work with Pauli measurement?

thierry-martinez: Fixed in e374cd5.

thierry-martinez: This definitely indicates that unification is needed! However, I don’t think this necessarily dictates a specific order for merging:
- Regardless of which PR is merged first, the others can be unified during the rebasing process.
- It seems to me that the primary need for unification is between `MeasurementDescription` (in #178) and `Measurement` (in #191), as both classes are intended for general measurement (and neither of these PRs introduces a class specifically for Pauli measures).
- `pauli.py` seems like an appropriate place for defining `PauliMeasure`.

graphix/pattern.py
```diff
@@ -1974,24 +1965,24 @@ def pauli_nodes(pattern: Pattern, leave_input: bool):
     if not pattern.is_standard():
         pattern.standardize()
     m_commands = pattern.get_measurement_commands()
-    pauli_node: list[tuple[command.M, str]] = []
+    pauli_node: list[tuple[command.M, PauliMeasure]] = []
     # Nodes that are non-Pauli measured, or pauli measured but depends on pauli measurement
     non_pauli_node: set[int] = set()
     for cmd in m_commands:
-        pm = is_pauli_measurement(cmd, ignore_vop=True)
+        pm = PauliMeasure.try_from(cmd.plane, cmd.angle)
```
shinich1, line 1972:
```suggestion
        pm = PauliMeasure.try_from(cmd.plane, cmd.angle)  # None returned if the measurement is not in Pauli basis
```

graphix/pauli.py
shinich1: sorry - could you rename to `PauliMeasurement` to be consistent with other classes mentioned above?

shinich1:
@thierry-martinez thanks! will approve once the above two comments resolved.

graphix/pauli.py
thierry-martinez: Fixed in 6388d4.

thierry-martinez:
> @thierry-martinez thanks! will approve once the above two comments resolved.

Done! Thanks @shinich1 .

shinich1 (approved):

EarlMilktea:
Memo: I confirmed these `Number`-related `mypy` errors are newly introduced after this PR merge.
Let's resolve them in #201 !

```bash
graphix/pauli.py:51: error: Incompatible return value type (got "Sign", expected "SignOrNumber")  [return-value]
graphix/pauli.py:53: error: Incompatible return value type (got "int", expected "SignOrNumber")  [return-value]
graphix/pauli.py:53: error: Unsupported operand types for * ("int" and "Number")  [operator]
graphix/pauli.py:58: error: Incompatible return value type (got "int", expected "Number")  [return-value]
graphix/pauli.py:58: error: Unsupported operand types for * ("int" and "Number")  [operator]
graphix/pauli.py:342: error: No overload variant of "int" matches argument type "Number"  [call-overload]
graphix/pauli.py:356: error: Argument 1 to "is_int" has incompatible type "float"; expected "Number"  [arg-type]
```

PR #198, opened 2024-08-21
Move `rand_circuit` and `rand_gate` to `random_objects` module and pass `rng` to every random object

thierry-martinez:
This pull request contains two commits:

- Commit 1b78cd9:
    - Moves the functions defined in `tests/random_circuit.py` to `graphix/random_objects.py`.
    - Removes the module `tests/random_circuit.py`.
    - Renames `get_rand_gate` and `get_rand_circuit` to `rand_gate` and `rand_circuit`, respectively, for uniformity.
    - Placing the random circuit generation functions in the `random_objects` module, alongside other random generators, seems more logical. Additionally, the random_circuit module was located in `tests/`, making it inaccessible to users (e.g., from other packages).

- Commit 5b69b7a:
    - Adds an `rng` parameter to all functions in `random_objects`, instead of relying on the default state of `np.random`.
    - Updates tests using random_objects to utilize the common `fx_rng` fixture, similar to how `rand_gate` and `rand_circuit` were already handled.


EarlMilktea:
@thierry-martinez 

Good PR indeed. Let me review it later!

graphix/random_objects.py
```diff
@@ -1,36 +1,45 @@
...
+from graphix.transpiler import Circuit
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable, Iterator
+
+    from numpy.random import Generator
 
 
-def rand_herm(sz: int):
+def rand_herm(sz: int, rng: Generator):
```
EarlMilktea, line 21:
```suggestion
def rand_herm(sz: int, rng: Generator) -> npt.NDArray:
```

graphix/random_objects.py
```diff
@@ -1,36 +1,45 @@
...
     """
     generate random hermitian matrix of size sz*sz
     """
-    tmp = np.random.rand(sz, sz) + 1j * np.random.rand(sz, sz)
+    tmp = rng.random(size=(sz, sz)) + 1j * rng.random(size=(sz, sz))
     return tmp + tmp.conj().T
 
 
-def rand_unit(sz: int):
+def rand_unit(sz: int, rng: Generator):
```
EarlMilktea, line 29:
```suggestion
def rand_unit(sz: int, rng: Generator) -> npt.NDArray:
```

graphix/random_objects.py
```diff
@@ -1,36 +1,45 @@
...
     else:
-        return unitary_group.rvs(sz)
+        return unitary_group.rvs(sz, random_state=rng)
 
 
 UNITS = np.array([1, 1j])
 
 
-def rand_dm(dim: int, rank: int | None = None, dm_dtype=True) -> DensityMatrix | np.ndarray:
+def rand_dm(dim: int, rng: Generator, rank: int | None = None, dm_dtype=True) -> DensityMatrix | np.ndarray:
```
EarlMilktea, line 42:
```suggestion
def rand_dm(dim: int, rng: Generator, rank: int | None = None, dm_dtype=True) -> DensityMatrix | npt.NDArray:
```

graphix/random_objects.py
```diff
@@ -52,14 +61,14 @@ def rand_dm(dim: int, rank: int | None = None, dm_dtype=True) -> DensityMatrix |
...
-    evals = np.random.rand(rank)
+    evals = rng.random(size=rank)
 
     padded_evals = np.zeros(dim)
     padded_evals[: len(evals)] = evals
 
     dm = np.diag(padded_evals / np.sum(padded_evals))
 
-    randU = rand_unit(dim)
+    randU = rand_unit(dim, rng)
```
EarlMilktea, line 71: Could you change the variable name to `rand_u` or similar?

graphix/random_objects.py
```diff
@@ -69,7 +78,7 @@ def rand_dm(dim: int, rank: int | None = None, dm_dtype=True) -> DensityMatrix |
         return dm
 
 
-def rand_gauss_cpx_mat(dim: int, sig: float = 1 / np.sqrt(2)) -> np.ndarray:
+def rand_gauss_cpx_mat(dim: int, rng: Generator, sig: float = 1 / np.sqrt(2)) -> np.ndarray:
```
EarlMilktea, line 81:
```suggestion
def rand_gauss_cpx_mat(dim: int, rng: Generator, sig: float = 1 / np.sqrt(2)) -> npt.NDArray:
```

graphix/random_objects.py
```diff
@@ -1,36 +1,45 @@
 from __future__ import annotations
 
+import functools
+from typing import TYPE_CHECKING
+
 import numpy as np
```
EarlMilktea, line 6:
```suggestion
import numpy as np
import numpy.typing as npt
```

graphix/random_objects.py
```diff
@@ -173,3 +182,101 @@ def rand_Pauli_channel_kraus(dim: int, rank: int | None = None) -> KrausChannel:
...
+        if use_rzz:
+            for j, k in genpair(nqubits, 2, rng):
+                circuit.rzz(j, k, np.pi / 4)
+        if use_ccx:
+            for j, k, l in gentriplet(nqubits, 2, rng):
+                circuit.ccx(j, k, l)
+        for j, k in genpair(nqubits, 4, rng):
+            circuit.swap(j, k)
+        for j in range(nqubits):
+            rng.choice(gate_choice)(j)
```
EarlMilktea, line 281:
```suggestion
            ind = rng.integers(len(gate_choice))
            gate_choice[ind](j)
```

EarlMilktea (changes requested):
Excellent! Only few suggestions related to type hints/warnings.

EarlMilktea:
@thierry-martinez 

If you don't mind, could you resolve remaining four similar warnings?
They can be located by this:

```bash
ruff check --select NPY --extend-exclude examples --extend-exclude benchmarks
```

graphix/random_objects.py
```diff
@@ -1,36 +1,46 @@
...
+from graphix.transpiler import Circuit
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable, Iterator
+
+    from numpy.random import Generator
 
 
-def rand_herm(sz: int):
+def rand_herm(sz: int, rng: Generator) -> npt.NDArray:
```
shinich1, line 22: can you make `rng` optional? I don't think it's handy to set up rng all the time..

thierry-martinez: Fixed in 33231b6. Thanks!

thierry-martinez:
> @thierry-martinez
> 
> If you don't mind, could you resolve remaining four similar warnings? They can be located by this:
> 
> ```shell
> ruff check --select NPY --extend-exclude examples --extend-exclude benchmarks
> ```

Done!

graphix/rng.py
```diff
@@ -0,0 +1,14 @@
...
+
+from typing import TYPE_CHECKING
+
+import numpy as np
+
+if TYPE_CHECKING:
+    from numpy.random import Generator
+
+
+def ensure_rng(rng: Generator | None = None) -> Generator:
```
EarlMilktea, line 11: I believe rng is designed to be reused (i.e., typically instantiated only once), as its quality is argued for *different* draws from the *same* instance, rather than for different instances with distinct seeds.
Based on this observation, `ensure_rng` could be better implemented with thread-local storage (cache `rng`, maybe with weakref) ?

EarlMilktea (approved):
LGTM!
Let me suggest another possible design for `ensure_rng` (either will be fine though).

graphix/rng.py
thierry-martinez: I believe that when `ensure_rng` is called with a non-`None` argument, it should function as an identity. Usage scenarios can vary, and we should allow the caller to choose the implementation that best suits their needs. For example, it might be useful to run two simulations with distinct `rng` instances initialized with the same seed to compare them on the same measurement path.

As for the default behavior when `None` is passed, I am open to better suggestions. However, I think `get_default_rng()` should perform appropriately, and if it doesn't, we should consider proposing improvements to the implementation in `numpy` itself.

EarlMilktea: > `get_default_rng()` should perform appropriately

Yeah.
May be a little bit slow though, as seed initialization requires true random number.
Let's discuss if it happened to be too slow.

thierry-martinez: Sorry for the delay in understanding, but I now see your point. I've addressed it in commit 5cd6dee. Is this what you had in mind? Unfortunately, `np.random.Generator` does not support `weakref` yet.

EarlMilktea: Exactly!
Let me submit some small fixes.

EarlMilktea: How about something like this?

```python

import threading
import typing

import numpy as np

if TYPE_CHECKING:

    from numpy.random import Generator


class _RNGStorage:
    rng: Generator | None


def _ensure_rng_wrapper() -> Callable[[Generator | None], Generator]:
    """Provide a hidden scope."""
    # HACK: Force mypy to check attributes
    rng_local = typing.cast(_RNGStorage, threading.local())
    rng_local.rng = None

    def ensure_rng(rng: Generator | None = None) -> Generator:
        if rng is not None:
            return rng
        if rng := rng_local.rng:
            return rng
        rng = np.random.default_rng()
        rng_local.rng = rng
        return rng

    return ensure_rng


ensure_rng = _ensure_rng_wrapper()
```

thierry-martinez: In the line `if rng := rng_local.rng:` within `ensure_rng`, you cannot assume that the `rng` attribute exists on the thread-local object, because the initialization `rng_local.rng = None` is only performed for the current thread. This means that when `ensure_rng` is called in a different thread, the `rng` attribute may not be present, leading to an `AttributeError`. The new `test_rng.py` that I added in commit f694ad1 demonstrates this issue.

More generally, casting `threading.local()` to `_RNGStorage` is not type-safe, because the `rng` attribute does not always exist on instances of thread-local storage.

Regarding the introduction of `_ensure_rng_wrapper`, while it does hide `rng_local` from the module interface, it adds unnecessary complexity and introduces a new symbol (`_ensure_rng_wrapper`) into the module's scope. Although I understand the intent to protect `rng_local` from external modification, I don't believe this justifies the added complexity. If the goal is to prevent accidental mutation, simply renaming `rng_local` to `_rng_local` to indicate its internal use by convention should suffice. After all, if external code is determined to interfere with `ensure_rng`, it could do so regardless.

What do you think about this simpler approach?

EarlMilktea: >  you cannot assume that the `rng` attribute exists on the thread-local object

Verified this behavior.
Thank you for helping me out!
Why not adding this kind of test to detect similar errors?

```python
def test_threadpool() -> None:
    with ThreadPoolExecutor() as executor:
        tasks = executor.map(lambda _: ensure_rng(None), range(100))
    for _ in tasks:
        pass
```
As for the code, I prefer `setattr` with noqa here to highlight the hidden risks.
I admit closure design is too much here...

```python
    rng = np.random.default_rng()
    # MEMO: Cannot perform type check
    setattr(_rng_local, "rng", rng)  # noqa: B010
```

thierry-martinez: Fixed in 4d3d685 and 7fa5f00. Thanks!

shinich1 (approved):
please add docstring as mentioned below, and check that docs compile without error (add fix if there's error), before merging. Also, can you make sure `random_objects` is covered by docs?

graphix/sim/density_matrix.py
```diff
@@ -317,7 +322,14 @@ def apply_channel(self, channel: KrausChannel, qargs):
...
+    def __init__(
+        self,
+        pattern,
+        max_qubit_num=12,
+        pr_calc=True,
+        input_state: Data = graphix.states.BasicStates.PLUS,
+        rng: Generator | None = None,
+    ):
         """
         Parameters
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in e4a48f2.

graphix/sim/statevec.py
```diff
@@ -27,7 +31,7 @@ def __init__(
         input_state: Data = graphix.states.BasicStates.PLUS,
         max_qubit_num=20,
         pr_calc=True,
-        rng=None,
+        rng: Generator | None = None,
     ):
         """
         Parameters
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in e4a48f2.

graphix/sim/tensornet.py
```diff
@@ -22,7 +27,9 @@ class TensorNetworkBackend:
     Executes the measurement pattern using TN expression of graph states.
     """
 
-    def __init__(self, pattern, graph_prep="auto", input_state=BasicStates.PLUS, **kwargs):
+    def __init__(
+        self, pattern, graph_prep="auto", input_state=BasicStates.PLUS, rng: Generator | None = None, **kwargs
+    ):
         """
 
         Parameters
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in e4a48f2.

shinich1:
@thierry-martinez have you checked docs compilation? I am seeing bunch of errors.

EarlMilktea:
@shinich1 Maybe better to introduce CI for documentation build?

PR #199, opened 2024-08-22
Use PEP8 naming

EarlMilktea:
**Description of the change:**

Rename everything violating PEP8 naming convention.

PR #201, opened 2024-08-23
[WIP] Refactor new features

EarlMilktea:

Refactor recent changes to `graphix`.

### DIFF
--- a/graphix/command.py
+++ b/graphix/command.py
@@ -2,15 +2,15 @@
 
 from __future__ import annotations
 
-import abc
+import dataclasses
 import enum
 from typing import TYPE_CHECKING
 
 import numpy as np
-from pydantic import BaseModel
 
 import graphix.clifford
-from graphix.pauli import Plane
+from graphix import utils
+from graphix.pauli import MeasureUpdate, Plane
 
 if TYPE_CHECKING:
     from graphix.clifford import Clifford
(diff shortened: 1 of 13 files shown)


EarlMilktea:
## Design 1: Remove unnecessary `pydantic` validations

Remove validations that can be elided, assuming type annotations are respected.
Conversely add validations to properties that cannot be checked statistically (ex. nonnegative `int`).

EarlMilktea:
## Design 2: Resolve `ABC` abuse

Remove `ABC` inheritance from `Instruction` and `Command`, as they have no abstract methods defined.

EarlMilktea:
## Design 3: Use `isinstance` for detecting command/instruction kind

Remove `kind` from `Instruction` and `Command`, as

- they can be overwritten
- leads to type instabilities

```python
op: S | N = S(node=0, domain={1, 2}) if random.random() < 1 / 2 else N(node=0)

if isinstance(op, S):
    print(op.domain)  # OK: Type narrowed to S

if op.kind == CommandKind.S:
    print(op.domain)  # ERROR: Type not narrowed
```

EarlMilktea:
@thierry-martinez
How about co-authoring this PR?
Also let me know your opinions/ideas!

EarlMilktea:
## Design 4: Resolve type warnings introduced by `isinstance(..., Number)`

thierry-martinez:
> @thierry-martinez How about co-authoring this PR? Also let me know your opinions/ideas!

That’s a lot of changes, and I completely agree—it’s always better to approach such significant updates with input from multiple perspectives.

Generally, I think GitHub issues might be a better platform for initial discussions compared to pull requests. It might help us avoid bundling too many changes into a single pull request (something I'm guilty of as well!). Smaller, more focused pull requests can be reviewed and merged more quickly, leading to a smoother workflow.

Regarding the points you raised:

- Design 1: I share your concerns about `pydantic`. It's indeed slow and rigid, and I’m still exploring other alternatives within the design space.
- Design 2: I agree that `disable_init` is a better approach here.
- Design 3: The `kind` field was a workaround we used due to the poor performance of `isinstance`. It’s crucial that any alternative we consider does not degrade performance. On the bright side, we’re likely to see significant performance improvements in instance creation since `pydantic` has been a bottleneck there.
- Design 4: I've already started addressing this in #158. Given the uncertainty around whether this PR will be kept, we can cherry-pick the parts concerning `numbers.Number` into a new, focused PR.

EarlMilktea:
> issues might be a better platform

Agree. I'm actually tempted to close this issue as huge changes are coming in #199 and #200 ...
Let's discuss later in issues.

PR #205, opened 2024-08-27
Refactor classes

### LINKED ISSUE #204: Use `pydantic` effectively

Currently `pydantic` validations are sorely used to validate types, which can be elided assuming type annotations are respected.

- Use `pydantic` to check additional constraints not manifested in type annotations, otherwise remove it
- Remove `BaseModel` inheritance to avoid unnecessary automatic validations
- Define simple models and use them manually

EarlMilktea:
## Context (if applicable):

Resolves #204 .

## Description of the change:

This PR refactors the class structures of `graphix`.

### Reduce `pydantic` usage

Drop `pydantic.BaseModel` inheritance from internal classes, i.e., those used to implement this package itself.
In the place of `pydantic.BaseModel`, this PR uses `dataclass` with type annotations instead.
This would contribute to better performance.

This PR also removes `__get_pydantic_core_schema__` defined in `State`, which is marked as unstable in #178 .

### Declare `kind` as an abstract property

This will enforce subclasses to have a read-only property named `kind`.

### Use `Literal` for better typing

This part is based on a famous design known as tagged union.
Suppose we have two similar classes defined as follows:

```python
class Tag(Enum):
    """Tag enum."""

    A = "A"
    B = "B"


@dataclasses.dataclass
class A:
    """Class A."""

    tag: Literal[Tag.A] = dataclasses.field(default=Tag.A, init=False)
    data: int


@dataclasses.dataclass
class B:
    """Class B."""

    tag: Literal[Tag.B] = dataclasses.field(default=Tag.B, init=False)
    data: str
```

Due to the non-overlapping `Literal` type annotations, `mypy` can infer the types correctly based on the `tag` field:

```python
def do_something() -> None:
    # Need explicit annotation here
    x: A | B = A(42) if rng.uniform() < 0.5 else B("Hello, World!")

    # Need to use `x.tag` rather than `tag = x.tag` plus `tag == ...` 
    if x.tag == Tag.A:
        # type(x) is A
    elif x.tag == Tag.B:
        # type(x) is B
    else:
        typing.assert_never(x)
```


graphix/command.py
```diff
@@ -3,65 +3,92 @@
...
     E = "E"
     C = "C"
     X = "X"
     Z = "Z"
-    T = "T"
     S = "S"
 
 
-class Command(BaseModel, abc.ABC):
+class Command(ABC):
```
thierry-martinez, line 34: Tagged unions do not seem to be recognized by `mypy` if their union type is a base class. I would rename `Command` as `BaseCommand` and define a type alias `Command = N | E | M | X | Z | C | T`, that `mypy` understands.

We can see that tagged unions using a base class as union type are not recognized by `mypy` on the following minimal example:
```python
"""Test tagged union."""

import dataclasses
import typing
from dataclasses import dataclass
from enum import Enum
from typing import Literal


class Tag(Enum):
    """Tag enum."""

    A = "A"
    B = "B"


@dataclass
class Base:
    """Base class."""

    tag: Tag


@dataclass
class A(Base):
    """Class A."""

    tag: Literal[Tag.A] = dataclasses.field(default=Tag.A, init=False)
    data: int


@dataclass
class B(Base):
    """Class B."""

    tag: Literal[Tag.B] = dataclasses.field(default=Tag.B, init=False)
    data: str


def do_something(x: Base) -> str:
    """Do something depending on the actual type of `x`."""
    if x.tag == Tag.A:
        return str(x.data + 1)
    if x.tag == Tag.B:
        return x.data + "a"
    typing.assert_never(x)
```
`mypy` finds the following "errors":
```
example.py:45: error: "Base" has no attribute "data"  [attr-defined]
example.py:47: error: "Base" has no attribute "data"  [attr-defined]
```
These errors disappear if `x` is of type `A | B` instead of `Base`.
```diff
-def do_something(x: Base) -> str:
+def do_something(x: A | B) -> str:
```

graphix/command.py
```diff
@@ -3,65 +3,92 @@
...
     """
     Base command class.
     """
 
-    kind: CommandKind = None
+    @property
+    @abc.abstractmethod
+    def kind(self) -> Any: ...
```
thierry-martinez, line 41: I believe `CommandKind` is a more appropriate type for `kind` than `Any`, even though this type is refined in subclasses.

Defining `kind` as a property instead of a field affects performance. On my laptop, running `pytest` takes 20.67 seconds on the `refactor-classes` branch (commit 261acbd), whereas it only takes 17.95 seconds on commit 124e3ab (the latest commit on `master` which this branch is based on). When I replace the `kind` method with a simple field (as in [this commit](https://github.com/thierry-martinez/graphix/commit/7236126c5ef61a4011f69e0c4e55e485e5c83413)), `pytest` takes only 17.10 seconds. While the timings themselves are not entirely consistent from one run to another, the comparison suggests that using a `kind` method is slower than the `pydantic` field version, which is, in turn, slower than the `pydantic`-free field version.

graphix/instruction.py
```diff
@@ -29,181 +32,264 @@ class InstructionKind(enum.Enum):
...
     """
-    Circuit instruction base class model.
+    Circuit instruction base class.
     """
 
-    kind: InstructionKind = None
-    meas_index: int = None
+    @property
+    @abc.abstractmethod
+    def kind(self) -> Any: ...
```
thierry-martinez, line 44: The same comments as for `Command` apply for `Instruction`:
* `kind` should be of type `InstructionKind`.
* `kind` should be a field.
* The class `Instruction` should be renamed (for instance, as `BaseInstruction`) and the type `Instruction` should be a union type.

graphix/types.py
```diff
@@ -1,10 +1,5 @@
 from __future__ import annotations
 
-import annotated_types
-import typing_extensions
-
-PositiveOrNullInt = typing_extensions.Annotated[int, annotated_types.Ge(0)]  # includes 0
-
```
thierry-martinez, line 7: Why do we remove this type?

thierry-martinez (changes requested):
This PR is a good idea, but I have the following concerns:
- `Command` and `Instruction` should be union types (`N | E | ...`) to ensure that tagged unions are recognized by `mypy`. It seems that `mypy` does not consider a base class to be a tagged union). Additionally, using explicit union types would allow us to utilize `typing.assert_never`.
- For performance reasons, `kind` should be a field rather than a method or property. If we stop using `pydantic` and define `kind` as a field, performance will improve. However, if `kind` is defined as a method, performance will degrade.

graphix/command.py
```diff
@@ -3,65 +3,92 @@
...
 
-class CommandKind(enum.Enum):
+class CommandKind(Enum):
     N = "N"
     M = "M"
     E = "E"
     C = "C"
     X = "X"
     Z = "Z"
-    T = "T"
```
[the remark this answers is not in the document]
EarlMilktea: I didn't notice that. Thank you!
Let me restore it later.

graphix/types.py
EarlMilktea: Because this definition would wrongly suggest that `PositiveOrNullInt` alone comes with validation: actually it's not unless used with `BaseModel` (am I correct?).
Actually, it was used in `graphix/sim/density_matrix.py` as such.

graphix/command.py
EarlMilktea:
```python
"""Validate property."""

import inspect
import typing
from typing import Any, ClassVar, Literal


class CheckedBase:
    """Check the child class."""

    def __init_subclass__(cls, **kwargs: Any) -> None:
        """Check the child class."""
        super().__init_subclass__(**kwargs)
        if not hasattr(cls, "tag"):
            msg = f"{cls.__name__} must have a tag class attribute initialized."
            raise TypeError(msg)
        ann = inspect.get_annotations(cls)
        if "tag" not in ann:
            msg = "Tag attribute must be annotated."
            raise TypeError(msg)
        ann_tag = ann["tag"]
        if typing.get_origin(ann_tag) is not ClassVar:
            msg = "Tag attribute must be a class variable."
            raise TypeError(msg)
        (ann_tag,) = typing.get_args(ann_tag)
        if typing.get_origin(ann_tag) is not Literal:
            msg = "Tag attribute must be a literal."
            raise TypeError(msg)


class Ok(CheckedBase):
    """Valid class."""

    tag: ClassVar[Literal[42]] = 42


class Bad(CheckedBase):
    """Invalid class."""

    spam: int
```

Something like this?

graphix/types.py
thierry-martinez: Types in Python don't inherently come with validation, so there's no surprise there. In the case of `graphix/sim/density_matrix.py`, there was indeed an explicit validation. However, you're also right that `mypy` currently ignores such type annotations, which could make it misleading to use such types without tools to enforce or check them.

graphix/command.py
thierry-martinez: Yes, you're correct—the base class doesn't seem necessary in this case.

graphix/instruction.py
EarlMilktea: Fixed in d4a15c0e4de2c2ffe5cef8d6afff2c03f75eb0c9 .
As we need to very careful when defining type aliases (leads to `mypy` warning if not exhaustive/incorrect), I temporally dropped some of them.

graphix/command.py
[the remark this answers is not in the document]
EarlMilktea: Fixed.

graphix/command.py
EarlMilktea: Fixed in d6f0a7c9acd299110087467a0b4de1771ba8e481 .

graphix/command.py
EarlMilktea: Please check d6f0a7c9acd299110087467a0b4de1771ba8e481 .

shinich1:
just out of interest, is 'kind' the right (idiomatic and/or normal) word for it? 'type' might be more straightforward? or perhaps just confusing?

EarlMilktea:
I personally prefer `kind` because…

- `type` is a special keyword in Python
- this field is not exactly the same as the corresponding type name

Also let us emphasize that this field is added for faster `isinstance` check: users are unlikely to use it directly, and it is, by design, not reassigned.

thierry-martinez:
`kind` is a common field name for tagged unions; you can find examples on [Wikipedia](https://en.wikipedia.org/wiki/Tagged_union). In Python specifically, `tag` is also a common name (see the [MyPy documentation on tagged unions](https://mypy.readthedocs.io/en/stable/literal_types.html#tagged-unions) for instance). In the specific case of `Command` and `Instruction`, `name` might be even more descriptive.

EarlMilktea:
@thierry-martinez

I updated the design.
(Note that `command.py` and `instruction.py` are also now monitored by `mypy`)
Could you review again?

shinich1 (approved):
Thank you @EarlMilktea and sorry for delay. This looks good to me but we need to update:
- tutorial
- changelog
- docstring for Pattern class (l59-73 of pattern.py

shinich1:
@thierry-martinez could you also take a final look?

graphix/pattern.py
```diff
@@ -380,7 +381,7 @@ def add_correction_domain(
                 if clifford_gate := c_dict.pop(cmd.node, None):
                     new_cmd = cmd.clifford(clifford_gate)
                 else:
-                    new_cmd = cmd.model_copy()
+                    new_cmd = dataclasses.replace(cmd)
```
thierry-martinez, line 384: Using `copy.copy` would be more idiomatic in this context. Additionally, Pydantic’s `model_copy` performed a deep copy, and we need it to handle aliasing correctly.
```suggestion
                    new_cmd = deepcopy(cmd)
```

Below is a test that could be added to `test_pattern.py` to demonstrate the issue:
```python
def test_pattern_aliasing(self) -> None:
    pattern_a = Pattern(input_nodes=[0, 1])
    pattern_b = Pattern(input_nodes=[1])
    pattern_a.add(M(node=0))
    pattern_a.add(X(node=1, domain={0}))
    m = M(node=1)
    pattern_a.add(m)
    pattern_b.add(m)
    pattern_a.standardize()
    pattern_b.simulate_pattern()
```

graphix/sim/statevec.py
```diff
@@ -114,7 +114,7 @@ def __init__(
 
         else:
             if isinstance(input_list[0], graphix.states.State):
-                graphix.types.check_list_elements(input_list, graphix.states.State)
+                type_utils.check_list_elements(input_list, graphix.states.State)
```
thierry-martinez, line 117: I am not sure we need to do that anymore (it is somehow redundant with static type checking).

tests/test_gflow.py
```diff
@@ -601,12 +601,12 @@ def test_rand_graph_pauliflow(self, fx_bg: PCG64, jumps: int) -> None:
             assert valid
 
     def test_corrections_from_pattern(self) -> None:
-        pattern = Pattern(input_nodes=range(5))
+        pattern = Pattern(input_nodes=list(range(5)))
```
thierry-martinez, line 604: We shouldn't need to do that, as `Pattern` already constructs a list from `input_nodes`. The type annotation in `Pattern` could be relaxed to `Sequence` (or `Iterable`, though in that case, we should be cautious about transient iterables by using `self.__input_nodes` instead of `input_nodes` when calculating `self.__n_node` and `self.__output_nodes`).

thierry-martinez (changes requested):
It would be better to use `deepcopy` instead of `dataclasses.replace` in `standardize_direct`.

EarlMilktea:
> This looks good to me but we need to update:
> 
> * tutorial
> * changelog

@shinich1 To my understanding, this PR does not introduce visible changes to user.

As for the docstring, let me update it when I submit another PR for refactoring `Pattern`.

graphix/sim/statevec.py
EarlMilktea: Sadly we need it until we annotate `statevec.py`.

tests/test_gflow.py
EarlMilktea: Let me work on this improvement in another PR for `Pattern`.

graphix/pattern.py
EarlMilktea: Fixed in f02a260a589aa65f435d236daca476b31cfedc85 .

thierry-martinez (approved):
LGTM! Thanks.

shinich1 (approved):

PR #209, opened 2024-09-03
Move all Pauli measurements to the front before Pauli presimulation

thierry-martinez:
This commit adds a method `__extract_pauli_nodes` that applies exchange rules to move all Pauli measurements to the front of a pattern.  This allows all the Pauli measurements to be presimulated, even if their initial domains depend on other measurements.

On random circuits of 10 qubits (depth 10), number of remaining nodes after Pauli presimulation is reduced by a factor of 4 on average, and the time spent on state vector simulation is reduced by a factor 75 (including the time for Pauli presimulation).


graphix/pattern.py
```diff
@@ -1421,13 +1421,32 @@ def run_pattern(self, backend, **kwargs):
         result = exe.run()
         return result
 
-    def perform_pauli_measurements(self, leave_input=False, use_rustworkx=False):
+    def perform_pauli_measurements(self, leave_input=False, use_rustworkx=False, ignore_pauli_with_deps=False):
```
EarlMilktea, line 1424: Let’s add type annotation.

graphix/pattern.py
```diff
@@ -1421,13 +1421,32 @@ def run_pattern(self, backend, **kwargs):
...
+            If `True`, Pauli measurements with domains depending on other measures are preserved as-is in the pattern.
+            If `False`, all Pauli measurements are preprocessed. Formally, measurements are swapped so that all Pauli measurements are applied first, and domains are updated accordingly.
+
         .. seealso:: :func:`measure_pauli`
 
         """
+        if not ignore_pauli_with_deps:
+            self.standardize()
+            pauli_nodes = self.__extract_pauli_nodes()
+            first_measure_index = next((i for i, cmd in enumerate(self) if cmd.kind == CommandKind.M), len(self))
```
EarlMilktea, line 1448: Let’s use more conventional style (such as `.find`? I don’t know the precise name)

graphix/pattern.py
```diff
@@ -1421,13 +1421,32 @@ def run_pattern(self, backend, **kwargs):
...
+            self.__seq[first_measure_index:first_measure_index] = pauli_nodes.values()
```
EarlMilktea, line 1449: This is not good as `.value` returns iterator.

graphix/pattern.py
```diff
@@ -1527,14 +1546,65 @@ def to_qasm3(self, filename):
...
-        result.__seq = [cmd.model_copy() for cmd in self.__seq]
+        result.__seq = [cmd.model_copy(deep=True) for cmd in self.__seq]
         result.__input_nodes = self.__input_nodes.copy()
         result.__output_nodes = self.__output_nodes.copy()
         result.__n_node = self.__n_node
         result._pauli_preprocessed = self._pauli_preprocessed
         result.results = self.results.copy()
         return result
 
+    def __extract_pauli_nodes(self, leave_nodes: set[int] | None = None) -> dict[int, tuple[command.M]]:
```
EarlMilktea, line 1557: Let’s use `_extract` rather than `__extract`: I feel it’s not idiomatic to define a “private” method, while I cannot tell why it’s bad.

EarlMilktea, line 1557: `tuple[command.M]` is almost identical to `command.M`. Why not use `command.M`?

graphix/pattern.py
```diff
@@ -1527,14 +1546,65 @@ def to_qasm3(self, filename):
...
+                    elif pm.axis == Axis.Y:
+                        shift_domains[cmd.node] = cmd.s_domain ^ cmd.t_domain
+                    elif pm.axis == Axis.Z:
+                        shift_domains[cmd.node] = cmd.s_domain
+                    else:
+                        typing_extensions.assert_never(pm.axis)
+                    cmd.s_domain = set()
+                    cmd.t_domain = set()
+                    pauli_nodes[cmd.node] = cmd
+                    del self.__seq[index]
```
EarlMilktea, line 1603: Please do not use `del`: if `.pop()` (no additional args.)  is what you wanted here, use it instead. Otherwise, we need to revise the algorithm for better performance.

graphix/pattern.py
```diff
@@ -1527,14 +1546,65 @@ def to_qasm3(self, filename):
...
+        """
+        if leave_nodes is None:
+            leave_nodes = set()
+        pauli_nodes = {}
+        shift_domains = {}
+        index = 0
+        while index < len(self.__seq):
+            cmd = self.__seq[index]
+
+            def expand_domain(domain):
```
EarlMilktea, line 1579: Please do not omit type annotation.

tests/test_pattern.py
```diff
@@ -449,9 +449,8 @@ def test_standardize_domains_and_clifford(self, fx_bg: PCG64, jumps: int) -> Non
         pattern.add(C(node=0, clifford=clifford.TABLE[c]))
         pattern_ref = pattern.copy()
         pattern.standardize(method="direct")
-        state_ref = pattern_ref.simulate_pattern()
-        state_p = pattern.simulate_pattern()
-        assert np.abs(np.dot(state_p.flatten().conjugate(), state_ref.flatten())) == pytest.approx(1)
+        _state_ref = pattern_ref.simulate_pattern()
```
EarlMilktea, line 452: Assignment is redundant here.

tests/test_pattern.py
```diff
@@ -449,9 +449,8 @@ def test_standardize_domains_and_clifford(self, fx_bg: PCG64, jumps: int) -> Non
...
+        _state_p = pattern.simulate_pattern()
```
EarlMilktea, line 453: Same as above.

EarlMilktea (changes requested):

graphix/pattern.py
thierry-martinez: Fixed in 162b0cb.

graphix/pattern.py
thierry-martinez: Fixed in 42de1df: I changed the logic so that Pauli measurement commands are no longer removed by `extract_pauli_nodes`, so there is no longer need to keep this method private.

graphix/pattern.py
thierry-martinez: Fixed in 42de1df. I changed the logic and now, a new sequence is built with an explicit loop that finds where the first non-Pauli measurement command is.

graphix/pattern.py
thierry-martinez: Fixed in 8e92b97: it was a mistake indeed, the return type is now `dict[int, command.M]`.

graphix/pattern.py
thierry-martinez: Fixed in 42de1df: I changed the logic and a new sequence is now built where Pauli measurements are inserted at the right place.

graphix/pattern.py
thierry-martinez: Fixed in 162b0cb (and the function is moved outside the loop).

tests/test_pattern.py
thierry-martinez: Fixed in b79406d: the removal of the assertion was a mistake, indeed.

tests/test_pattern.py
thierry-martinez: Fixed in https://github.com/TeamGraphix/graphix/commit/b79406d59da795300d361a271ebeae5b36cb6a61: the removal of the assertion was a mistake, indeed.

graphix/pattern.py
```diff
@@ -1416,12 +1416,46 @@ def run_pattern(self, backend, **kwargs):
         result = exe.run()
         return result
 
-    def perform_pauli_measurements(self, leave_input=False, use_rustworkx=False):
+    def perform_pauli_measurements(self, leave_input=False, use_rustworkx=False, ignore_pauli_with_deps=False) -> None:
```
EarlMilktea, line 1419:
```suggestion
    def perform_pauli_measurements(
        self, leave_input: bool = False, use_rustworkx: bool = False, ignore_pauli_with_deps=False
    ) -> None:
```

graphix/pattern.py
```diff
@@ -1416,12 +1416,46 @@ def run_pattern(self, backend, **kwargs):
...
+        if not ignore_pauli_with_deps:
+            self.standardize()
+            pauli_nodes = self.extract_pauli_nodes()
+            # Create a new sequence with all Pauli nodes to the front
+            new_seq = []
+            pauli_nodes_inserted = False
+            for cmd in self:
+                if cmd.kind == CommandKind.M:
+                    if cmd.node in pauli_nodes:
+                        pass
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in de76d20.

graphix/pattern.py
thierry-martinez: `ruff format` enforces the single-line version, which is expected because the line is only 119 characters long, falling within the 120-character limit.

EarlMilktea: I don't care code formats as long as `: bool` is added and CI passes.

thierry-martinez: Fixed in df843d9. Sorry, I didn't notice your suggestion to add type annotations. You added them to `leave_input` and `use_rustworkx` (parameters not modified in this PR) but did not add them to `ignore_pauli_with_deps`, which is the new parameter introduced in this PR. I have added the missing type annotations to all the arguments.

EarlMilktea: Oh I overlooked that. Thanks!

EarlMilktea (approved):

graphix/pattern.py
```diff
@@ -1416,12 +1416,46 @@ def run_pattern(self, backend, **kwargs):
...
+                if cmd.kind == CommandKind.M:
+                    if cmd.node not in pauli_nodes:
+                        if not pauli_nodes_inserted:
+                            new_seq.extend(pauli_nodes.values())
+                            pauli_nodes_inserted = True
+                        new_seq.append(cmd)
+                else:
+                    new_seq.append(cmd)
+            if not pauli_nodes_inserted:
+                new_seq.extend(pauli_nodes.values())
```
shinich1, line 1457: How is this possible? I guess you need a standardization (I know we do it in `measure_pauli` but we should be explicit rather than implicit.

shinich1: Also, this might move the comand across possible dependency. please add a line of two of notes about  consequences of this skipping or if this doesn't affect at all, then why so.

thierry-martinez: I changed to a clearer structure in 7483827, providing proofs of the signal shifting: the function `extract_pauli_nodes` has been renamed to `move_pauli_measurements_to_the_front` and is now responsible for both shifting and moving the Pauli measurements in the sequence.

shinich1 (approved):
Looks great. thank you!

PR #210, opened 2024-09-09
Refactor channels

EarlMilktea:

Get rid of `dict` from `KrausChannel`.


graphix/channels.py
```diff
@@ -2,78 +2,137 @@
...
-        assert check_data_normalization(kraus_data)
-        assert check_rank(kraus_data)
+        # Check that the channel is properly normalized, i.e., \sum_K_i^\dagger K_i = Identity.
+        data = next(iter(self.__data))
+        work = np.zeros_like(data.operator, dtype=np.complex128)
+        for data in self.__data:
+            m = data.coef * data.operator
+            work += m.conj().T @ m
+        if not np.allclose(work, np.eye(2**self.__nqubit)):
+            raise ValueError("The specified channel is not normalized.")
```
thierry-martinez, line 116: This code could be vectorize as follows.
```python

        # Check that the channel is properly normalized, i.e., \sum_K_i^\dagger K_i = Identity.
        coefs = np.asarray([data.coef for data in self.__data])
        ops = np.asarray([data.operator for data in self.__data])
        op_sum = np.einsum("i,i...jk,i,i...kl->...jl", coefs, np.transpose(ops.conj(), axes=[0, *range(ops.ndim - 1, 0, -1)]), coefs, ops)
        if not np.allclose(op_sum, np.eye(2**self.__nqubit)):
            raise ValueError("The specified channel is not normalized.")
```

graphix/channels.py
```diff
@@ -2,78 +2,137 @@
...
 
-        # np.asarray(data, dtype=np.complex128)
-        # number of Kraus operators in the Channel
+    @typing.overload
+    def __getitem__(self, index: slice, /) -> list[KrausData]: ...
 
-        self.size = len(kraus_data)
+    def __getitem__(self, index: SupportsIndex | slice, /) -> KrausData | list[KrausData]:
+        """Return the Kraus operator at the given index."""
+        return copy.deepcopy(self.__data[index])
```
thierry-martinez, line 126: Do we need a `deepcopy` here? We could just return a read-only view.

graphix/linalg_validations.py
```diff
@@ -2,29 +2,58 @@
...
+    """Check if matrix is square, and then call the function."""
+
+    # wraps(_f) has very messy annotation, but still callable
+    @functools.wraps(f)
+    def _f(matrix: npt.NDArray[_T], *args: _P.args, **kwargs: _P.kwargs) -> bool:
+        if not is_square(matrix):
+            raise ValueError("Need to ensure that matrix is square first.")
+        return f(matrix, *args, **kwargs)
+
+    return _f  # type: ignore[return-value]
```
thierry-martinez, line 39: I would prefer a simple function `assert_square` which raises `ValueError` if not `is_square`.

graphix/linalg_validations.py
```diff
@@ -2,29 +2,58 @@
...
 
 
-def check_square(matrix: np.ndarray) -> bool:
+@_is_square_deco
+def is_qubitop(matrix: npt.NDArray[_T]) -> bool:
```
thierry-martinez, line 43: I would prefer that `is_qubitop` returns `False` if `matrix` is not square instead of raising a `ValueError`. We could define `assert_qubitop` that raises `ValueError` with the respective messages if the matrix is not square or if the size is not a power of two.

The same comment applies to other `is_*` predicates.

thierry-martinez (changes requested):
Good job! I propose some optimizations, and it would be better to have a family of `is_*` predicates that do not raise exceptions, and a family of `assert_*` (or `check_*` or `validate_*` if you don't like `assert` for production checks) that raises common exceptions.

graphix/linalg_validations.py
EarlMilktea: I don't want both `is_` and `assert_` to be in the pkg. as they're essentially the same and `is_` functions are mainly used internally.

> `is_qubitop` returns `False` if `matrix` is not square

For this suggestion, I'm for that.

graphix/linalg_validations.py
EarlMilktea: Same as below.

> I don't want both `is_` and `assert_` to be in the pkg. as they're essentially the same and `is_` functions are mainly used internally.

For the design (now based on decorator), there may be better alternatives.

graphix/channels.py
EarlMilktea: >  just return a read-only view

How can we do this?
I don't know any method essentially faster than `copy.deepcopy` .

graphix/channels.py
EarlMilktea: I prefer simpler codes here as `KrausChannel.__init__` is very unlikely to be frequently called.

graphix/channels.py
```diff
@@ -2,78 +2,137 @@
...
+        if n <= 2**p:
+            return p
+    raise RuntimeError("Unreachable.")
+
+
+# MEMO: Too much?
+_T = TypeVar("_T", bound=np.generic)
+
+
+class KrausData:
```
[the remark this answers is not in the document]
EarlMilktea: I prefer `astype(copy=True)` here because `fronzen=True` does not save us from accidentally modifying `np.array` with multiple references.
As this class is exposed to users, we may be extra careful here.

graphix/_db.py
```diff
@@ -2,43 +2,51 @@
...
+
 
-def _lock(m: npt.NDArray[np.complex128]) -> npt.NDArray[np.complex128]:
-    """Create a true immutable view."""
+def _lock(data: npt.NDArray[_T]) -> npt.NDArray[np.complex128]:
+    """Create a true immutable view.
+
+    data must be a "temporary" array, i.e., should not be bound to a variable.
+    """
+    m = data.astype(np.complex128)
```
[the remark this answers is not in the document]
EarlMilktea: I added `.astype` here for simpler codes.
Since we cannot omit `[...]` in `npt.NDArray` (mypy strict rule), we would end up casting array literal to `np.complex128` somewhere else if we were to omit the cast here.

graphix/channels.py
```diff
@@ -2,78 +2,137 @@
...
+        """Validate the data."""
+        if not lv.is_square(self.operator):
+            raise ValueError("Operator must be a square matrix.")
+        if not lv.is_qubitop(self.operator):
+            raise ValueError("Operator must be a qubit operator.")
+        size, _ = self.operator.shape
+        return _ilog2(size)
+
 
 class KrausChannel:
```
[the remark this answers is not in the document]
EarlMilktea: Same as above.

graphix/_db.py
```diff
@@ -2,43 +2,51 @@
...
     m.flags.writeable = False
```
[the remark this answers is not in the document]
EarlMilktea: I'm using this strange design here for maximizing safety.
I admit your design is more concise, but it's also potentially unsafe.
If conciseness beats safety, `_lock` itself would be redundant.

graphix/channels.py
```diff
@@ -2,78 +2,137 @@
...
+            If kraus_data is empty.
         """
-        # check there is data
         if not kraus_data:
             raise ValueError("Cannot instantiate the channel with empty data.")
 
-        if not isinstance(kraus_data, (list, np.ndarray, tuple)):
-            raise TypeError(f"The data must be a list, a numpy.ndarray or a tuple not a {type(kraus_data)}.")
+        self.__nqubit = self._nqubit(kraus_data)
+        self.__data = list(copy.deepcopy(data) for data in kraus_data)
```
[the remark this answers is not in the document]
EarlMilktea: > not sure we want to impose a deep copy here

I inserted `deepcopy` here based on https://en.wikipedia.org/wiki/Principle_of_least_astonishment .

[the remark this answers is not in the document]
thierry-martinez: As a user, I would not expect a data structure that performs read-only access to duplicate data. While performing a deep copy or not may be negligible in this particular case, the question of whether to copy or not arises throughout the API. I believe the more uniform we can be, the better, particularly in terms of the principle of least astonishment. The correct rule of thumb, in my opinion, is to assume that unless explicitly documented otherwise, all data passed as arguments should be treated as passed by read-only reference, and should only be copied internally if there is a need to mutate it.

graphix/channels.py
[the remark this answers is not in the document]
thierry-martinez: I think that, as a read-only data structure, we do not expect passed values to be copied.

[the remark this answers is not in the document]
EarlMilktea: I don't think so as `dataclass` itself does not makes us feel that it's read-only. 
Actually `frozen=False` is invisible to users unless they wrongly try to reassign to it.

graphix/channels.py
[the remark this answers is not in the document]
EarlMilktea: The problem is that we cannot expect users use our public APIs correctly.
As this class is defined to be public (no `_` prefix), we should make the API as ergonomic as possible, or in other words, easy to use and hard to misuse.
Let me emphasize that I can understand what you said: As a developer, I myself always bear a similar principle as yours in mind, and in fact if this class were to be defined as module-private, I would be willing to omit `deepcopy` here.

@shinich1

How do you think about the API design?
I'm focusing on safety at the cost of additional costs whereas @thierry-martinez prioritizes runtime cost assuming that users respect docstrings/Python good practices.

graphix/_db.py
```diff
@@ -229,3 +233,59 @@ def _lock(m: npt.NDArray[np.complex128]) -> npt.NDArray[np.complex128]:
     ("h", "x", "sdg"),
     ("h", "x", "s"),
 )
+
+
+class WellKnown:
+    """Collection of well-known matrices."""
```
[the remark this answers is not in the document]
EarlMilktea: Merged in a72d682f16f0ccfff325cddaaf20fbc09b20db9c .

graphix/channels.py
```diff
@@ -2,78 +2,137 @@
...
+    from collections.abc import Iterable
+
+
+def _ilog2(n: int) -> int:
+    if n < 0:
+        raise ValueError("n must be non-negative.")
+    for p in itertools.count():
+        if n <= 2**p:
+            return p
+    raise RuntimeError("Unreachable.")
```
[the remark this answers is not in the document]
EarlMilktea: Merged in addf32f2935c756bb5cec01af905a29e6011dd42 .

graphix/channels.py
```diff
@@ -2,78 +2,137 @@
...
-        list(dict): [{coef: scalar, operator: list}, {coef: scalar, operator: list}, ...]
+    __nqubit: int
+    __data: list[KrausData]
 
-    """
+    @staticmethod
+    def _nqubit(kraus_data: Iterable[KrausData]) -> int:
+        try:
+            nqubits = set(data.nqubit for data in kraus_data)
+        except ValueError as e:
```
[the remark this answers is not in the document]
EarlMilktea: I feel we should not use `@dataclass` here because public attributes are subject to modification after validation in `__init__` or `__post_init__`.

graphix/linalg_validations.py
EarlMilktea: Partially merged in bb8d30f8babbd3b6adb3137717ddeedd4a8771d9 : now `is_` don't raise `ValueError` (emit a warning instead).

graphix/channels.py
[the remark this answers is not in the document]
EarlMilktea: I decided to use your idea here for simplicity.

graphix/channels.py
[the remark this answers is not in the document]
EarlMilktea: Fixed in 6ae4eaacb5286805e3ea30e34b772eabd6e43901 .

graphix/channels.py
```diff
@@ -2,78 +2,137 @@
...
+            raise ValueError("Failed to intialize KrausChannel object.") from e
+
+        if len(nqubits) != 1:
```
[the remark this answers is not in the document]
EarlMilktea: Fixed in 6ae4eaacb5286805e3ea30e34b772eabd6e43901 .

graphix/linalg_validations.py
EarlMilktea: Resolved in bb8d30f8babbd3b6adb3137717ddeedd4a8771d9 .

graphix/_db.py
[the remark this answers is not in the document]
shinich1: If this doesn't slow us down (I presume), it is perhaps ok as is, perhaps could you add short comments to the code what this is trying to avoid (no need to be detailed) so that people can understand intention later.

shinich1 (changes requested):
thanks, and my apologies for the delay. Looks good to me except two parts commented below.

graphix/_db.py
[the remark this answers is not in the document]
EarlMilktea: Resolved in e48e7ecaa26cb747c5d64b0aebd76f60d3a6c6c4 .

shinich1 (approved):
thanks! could add a line to changelog about the new `Krausdata` data structure before merging?

thierry-martinez:
I am still unsure what kind of guarantees we are looking for.
If the user really wants to break things, they can still do so in multiple ways. For instance, even if `v.flags.writeable = True` fails, `v.base.flags.writeable = True` succeeds. Or, even more simply, replacing `CLIFFORD` with a deep copy of itself will unlock all the read-only flags.
```python
graphix._db.CLIFFORD = copy.deepcopy(graphix._db.CLIFFORD)
```
Moreover, we have no static guarantee that `_lock` is used as intended (there is no type annotation in python to prevent aliasing).
On the other hand, creating read-only views is less complex and should be sufficient to prevent accidental mutations. It also seems to strike the right balance in terms of conveying our intention for safety.

graphix/channels.py
thierry-martinez: `KrausData` _is_ a read-only view, isn't it?

EarlMilktea:
> `v.base.flags.writeable = True`

I didn't know that. Thanks!

Now that I'm aware of this hidden caveat, I'm not eager to ensure safety in `_lock` as before, but also feel that it is not so bad as to be corrected immediately.
There are many other issues that need to be fixed.

graphix/channels.py
EarlMilktea: Not completely read-only as `KrausData` contains mutable classes.

EarlMilktea: Anyway, unlike `deepcopy` in `__init__` that is always necessary, we could be optimistic here...

EarlMilktea: Fixed in 7ad9425f4933b0c2d963f060cd432e934dba1333 .

thierry-martinez: > Not completely read-only as `KrausData` contains mutable classes.

Indeed, but the mutability does not seem to be publicly exposed, isn't it? I mean, mutable data is private and accessors make the interface immutable, as far as I understand.

EarlMilktea:
@shinich1 

Updated the changelog.

@thierry-martinez 

I feel most of the comments are resolved/negotiated.
Do you approve this PR?

graphix/linalg_validations.py
```diff
@@ -2,29 +2,41 @@
...
-    if len(matrix.shape) != 2:
-        raise ValueError(f"The object has {len(matrix.shape)} axes but must have 2 to be a matrix.")
-    if matrix.shape[0] != matrix.shape[1]:
-        raise ValueError(f"Matrix must be square but has different dimensions {matrix.shape}.")
-    size = matrix.shape[0]
-    if size & (size - 1) != 0:
-        raise ValueError(f"Matrix size must be a power of two but is {size}.")
-    return True
+    if not is_square(matrix):
+        warnings.warn(f"is_square is not called before {is_qubitop.__name__}.", stacklevel=1)
```
thierry-martinez, line 25: Why such a warning? Why should I write `is_square(m) and is_qubit(m)` instead of just `is_qubit(m)` (that would call `is_square` internally in any case)? The doc-comment suggests that `is_qubit(m)` should really be equivalent to `is_square(m) and is_qubit(m)`.

graphix/sim/density_matrix.py
```diff
@@ -74,10 +74,15 @@ def check_size_consistency(mat):
                 try:
                     if isinstance(input_list[0], Iterable) and isinstance(input_list[0][0], numbers.Number):
                         self.rho = np.array(input_list)
-                        assert check_square(self.rho)
+                        if not lv.is_square(self.rho):
+                            raise ValueError("Density matrix must be square.")
+                        if not lv.is_qubitop(self.rho):
+                            raise ValueError("Cannot interpret the provided density matrix as a qubit operator.")
```
thierry-martinez, line 80: A function `ensure_qubitop(self.rho)` would remove code duplication and factor the logic of having dedicated error messages to catch not-square matrices between this method and `KrausData.__init__`.

thierry-martinez (changes requested):
I still have some concerns with checks.

graphix/linalg_validations.py
EarlMilktea: It's originating from the original implementation.
Let me refactor it!

EarlMilktea: Fixed in 65eb69a2c5db92b2b77acb19c4f55f2f27d97a56 .

thierry-martinez (approved):

PR #211, opened 2024-09-12
Start enforcing typecheck

EarlMilktea:

Use `mypy` and `pyright` to detect type warnings.


pyproject.toml
```diff
@@ -93,3 +93,31 @@ convention = "numpy"
 [tool.pytest.ini_options]
 # Silence cotengra warning
 filterwarnings = ["ignore:Couldn't import `kahypar`"]
+
+[tool.mypy]
+disable_error_code = [
+  "type-arg",
```
thierry-martinez, line 99: We can enable `type-arg` with only few changes: https://github.com/thierry-martinez/graphix/commit/d5c4d50d653082294b0eb93784cb8da46ec7145b

pyproject.toml
```diff
@@ -93,3 +93,31 @@ convention = "numpy"
...
+  "unused-ignore", # Necessary for pyright
```
thierry-martinez, line 100: Can you explain why this line is necessary? I don't see any error from either `mypy` or `pyright` when it is removed.

thierry-martinez (changes requested):
Very nice! However, I think we can remove the disabled error code.

requirements.txt
```diff
@@ -9,4 +9,6 @@ pydantic
 quimb>=1.4.0
 scipy
 sympy>=1.9
+types-networkx
+types-psutil
```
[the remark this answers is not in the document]
EarlMilktea: Sounds reasonable, but especially for `types-networkx` affecting our public APIs, I'm not sure whether users can see annotations without installing it...

pyproject.toml
EarlMilktea: I know it's not necessary for now, but would be necessary as it's really hard to resolve all the placeholders used in `npt.NDArray`: this is why I added it.
How do you think about it?

requirements.txt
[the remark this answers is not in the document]
EarlMilktea: Looks like it's unnecessary.
Fixed in 2bfc76e21f97cb133193d986abc549684e0a87e3.

pyproject.toml
EarlMilktea: This can be removed if we use `pyright: ignore` instead of `type: ignore` when only `pyright` complains.

EarlMilktea: Fixed in 74ce7925275807dd18f21b301f3f659b6d5dc0f1.

pyproject.toml
EarlMilktea: Anyway let me adopt your idea here.

thierry-martinez (approved):

thierry-martinez:
LGTM! Thanks.

shinich1 (approved):
Looks good! let's squash and merge.

PR #216, opened 2024-10-02
Refactor `clifford.py`

EarlMilktea:

- Refactor `Clifford`
  - Custom comparison/hash
  - Remove `get` and `TABLE`
  - Assign well-known Cliffords directly to the class itself
- Add type annotations
  - Enforce type checks
- Tidy up imports


graphix/clifford.py
```diff
@@ -37,19 +39,40 @@ class Domains:
 class Clifford:
```
thierry-martinez, line 39: `Clifford` can be a frozen dataclass with the single field `index`.

graphix/clifford.py
```diff
@@ -96,33 +119,31 @@ def commute_domains(self, domains: Domains) -> Domains:
...
-
-
-def get(index: int) -> Clifford:
-    """Return the Clifford gate with given index."""
-    return TABLE[index]
+    @staticmethod
+    def cliffords() -> Iterator[Clifford]:
+        """Return an iterator over the Clifford gates."""
+        for i in range(len(CLIFFORD)):
+            yield Clifford(i)
```
thierry-martinez, line 141: It would be more idiomatic to be able to write `list(Clifford)` (as we would be able to do if `Clifford` was defined as an `Enum`).

tests/test_pattern.py
```diff
@@ -443,10 +446,10 @@ def test_pauli_measurement_then_standardize(
     @pytest.mark.parametrize("jumps", range(1, 11))
     def test_standardize_two_cliffords(self, fx_bg: PCG64, jumps: int) -> None:
         rng = Generator(fx_bg.jumped(jumps))
-        c0, c1 = rng.integers(len(clifford.TABLE), size=2)
+        c0, c1 = rng.integers(len(CTAB), size=2)
         pattern = Pattern(input_nodes=[0])
-        pattern.add(C(node=0, clifford=clifford.TABLE[c0]))
-        pattern.add(C(node=0, clifford=clifford.TABLE[c1]))
+        pattern.add(C(node=0, clifford=CTAB[c0]))
```
thierry-martinez, line 451: It would be better to be able to do `Clifford(c0)`, without worrying about having a new instance.

tests/test_db.py
```diff
@@ -24,20 +29,20 @@ def classify_pauli(arr: npt.NDArray[np.complex128]) -> tuple[int, int]:
         Returns
         -------
             ind : tuple
-                tuple containing (pauli index, sign index)
+                tuple containing (Pauli string, sign)
```
thierry-martinez, line 32: `CliffordMeasure` (or `PauliMeasurement`).

thierry-martinez (changes requested):
Good job! I particularly appreciate the following improvements:
* the namespace `Clifford` (`Clifford.I`, ...),
* the use of the constructor `Clifford(index)` instead of `clifford.get(index)`.
I have the following suggestions:
* `clifford.CliffordMeasure` and `pauli.PauliMeasurement` are isomorphic: we should only keep one (and regarding the naming, there is a preference for `Measurement`: https://github.com/TeamGraphix/graphix/pull/197#discussion_r1727538078).
* I would except several calls to `Clifford(0)` to return the same instance, and I would prefer to write `iter(Clifford)` instead of `Clifford.cliffords()` (as I would do if it were an `Enum`), I would suggest the following changes in these lines: https://github.com/thierry-martinez/graphix/commit/8dc4f95fb6535ad7304c87ed831e012b442c0473

tests/test_pattern.py
EarlMilktea: I totally agree, but then it would be better implemented using the singleton pattern...

graphix/clifford.py
EarlMilktea: I updated the code to use `Enum` here.

tests/test_pattern.py
EarlMilktea: Resolved by `Enum` inheritance.

graphix/_db.py
```diff
@@ -143,38 +143,40 @@ def _lock(data: npt.NDArray[_T]) -> npt.NDArray[np.complex128]:
 # For instance, CLIFFORD[7].conj().T = - CLIFFORD[CLIFFORD_CONJ[7]]
 CLIFFORD_CONJ = (0, 1, 2, 3, 5, 4, 6, 15, 12, 9, 10, 11, 8, 13, 14, 7, 20, 22, 23, 21, 16, 19, 17, 18)
 
+
+class CliffordMeasure(NamedTuple):
+    pstr: Literal["X", "Y", "Z"]
+    sign: Literal[-1, +1]
```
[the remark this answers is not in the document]
EarlMilktea: I don't want to merge the two as doing so will complicate module structure (possibly with circular imports).

As this class is just for documentation, how about making it private (`_CliffordMeasure`) and leave it as-is?

tests/test_db.py
EarlMilktea: Fixed in bb6e8489d114667bd5389c1a05c14fd43d6ac1c2 .

thierry-martinez (approved):
Using `Enum` is a good idea, indeed! I approve this PR as is, but I wonder if we shouldn't remove the `index` property.

EarlMilktea:
I may submit another PR for refactoring `pauli.py` based on similar idea (private `Enum` variants)...

graphix/clifford.py
```diff
@@ -34,58 +35,89 @@ class Domains:
...
+    _22 = 22
+    _23 = 23
 
     @property
     def index(self) -> int:
-        """Return the index of the Clifford gate (inverse of clifford.get)."""
-        return self.__index
+        """Return the index of the Clifford gate."""
+        # mypy does not infer variant type (pyright does)
+        return self.value  # type: ignore[no-any-return]
```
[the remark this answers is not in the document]
EarlMilktea: Fixed in 0f526db75ad1abd4931747b3d44ab40080cceabb .

Please be noted that `mypy` is not intelligent enough to be able to infer the type of `.value`: this is why I added `pyright`.

EarlMilktea:
@shinich1 @thierry-martinez 

Could you check/review the latest codes?

graphix/sim/base_backend.py
```diff
@@ -101,7 +101,7 @@ def _op_mat_from_result(vec: tuple[float, float, float], result: bool) -> np.nda
     op_mat = np.eye(2, dtype=np.complex128) / 2
     sign = (-1) ** result
     for i in range(3):
-        op_mat += sign * vec[i] * graphix.clifford.CLIFFORD[i + 1] / 2
+        op_mat += sign * vec[i] * _db.CLIFFORD[i + 1] / 2
```
thierry-martinez, line 104: We could avoid the use of `_db`.

```suggestion
        op_mat += sign * vec[i] * Clifford(i + 1).matrix / 2
```

tests/test_pauli.py
```diff
@@ -56,21 +56,21 @@ def test_measure_update(
         angle: float,
         choice: bool,
     ) -> None:
-        vop = clifford.index
+        vop: int = clifford.value
         if s:
-            vop = graphix.clifford.CLIFFORD_MUL[1][vop]
+            vop = CLIFFORD_MUL[1][vop]
```
thierry-martinez, line 61:
```suggestion
            clifford = Clifford.X @ clifford
```

tests/test_pauli.py
```diff
@@ -56,21 +56,21 @@ def test_measure_update(
...
         if t:
-            vop = graphix.clifford.CLIFFORD_MUL[3][vop]
+            vop = CLIFFORD_MUL[3][vop]
```
thierry-martinez, line 63:
```suggestion
            clifford = Clifford.Z @ clifford
```

tests/test_pauli.py
```diff
@@ -56,21 +56,21 @@ def test_measure_update(
...
         vec = plane.polar(angle)
         op_mat_ref = np.eye(2, dtype=np.complex128) / 2
         for i in range(3):
-            op_mat_ref += (-1) ** (choice) * vec[i] * graphix.clifford.CLIFFORD[i + 1] / 2
-        clifford_mat = graphix.clifford.CLIFFORD[vop]
+            op_mat_ref += (-1) ** (choice) * vec[i] * CLIFFORD[i + 1] / 2
```
thierry-martinez, line 67:
```suggestion
            op_mat_ref += (-1) ** (choice) * vec[i] * Clifford(i + 1).matrix / 2
```

tests/test_pauli.py
```diff
@@ -56,21 +56,21 @@ def test_measure_update(
...
+        clifford_mat = CLIFFORD[vop]
```
thierry-martinez, line 68:
```suggestion
        clifford_mat = clifford.matrix
```

tests/test_pauli.py
```diff
@@ -56,21 +56,21 @@ def test_measure_update(
...
         op_mat_ref = clifford_mat.conj().T @ op_mat_ref @ clifford_mat
         measure_update = MeasureUpdate.compute(plane, s, t, clifford)
         new_angle = angle * measure_update.coeff + measure_update.add_term
         vec = measure_update.new_plane.polar(new_angle)
         op_mat = np.eye(2, dtype=np.complex128) / 2
         for i in range(3):
-            op_mat += (-1) ** (choice) * vec[i] * graphix.clifford.CLIFFORD[i + 1] / 2
+            op_mat += (-1) ** (choice) * vec[i] * CLIFFORD[i + 1] / 2
```
thierry-martinez, line 75:
```suggestion
            op_mat += (-1) ** (choice) * vec[i] * Clifford(i + 1).matrix / 2
```

thierry-martinez (changes requested):
I just propose some changes to use the `Clifford` abstraction instead of `_db`.

shinich1 (approved):

graphix/clifford.py
```diff
@@ -34,58 +35,84 @@ class Domains:
...
     @property
-    def matrix(self) -> np.ndarray:
+    def matrix(self) -> npt.NDArray[np.complex128]:
         """Return the matrix of the Clifford gate."""
-        return CLIFFORD[self.__index]
+        return CLIFFORD[self.value]
 
     def __repr__(self) -> str:
         """Return the Clifford expression on the form of HSZ decomposition."""
         return " @ ".join([f"graphix.clifford.{gate}" for gate in self.hsz])
```
thierry-martinez, line 82: The namespace has changed.

```suggestion
        return " @ ".join([f"Clifford.{gate}" for gate in self.hsz])
```

thierry-martinez (changes requested):

graphix/clifford.py
EarlMilktea: Fixed in 4ff8e6830c82e75aaaa27e1908cf95b09acc7de4 .

thierry-martinez (approved):
LGTM! Thanks.

PR #217, opened 2024-10-04
Tidy up imports

EarlMilktea:

- Use coherent import style
- Add coverage ignore patterns
- Prohibit relative import
- Resolve deeply-nested `import` that will lead to `mypy` warning


shinich1 (commented):
LGTM. could you also improve imports in `example`s or do you think we can leave them like they are?

EarlMilktea:
> could you also improve imports in `example`s or do you think we can leave them like they are?

We could safely leave them untouched after adding `__all__` attributes.

Let me work on this!

EarlMilktea:
Or we may remove all the `import`s in `__init__.py` ?

shinich1:
> Or we may remove all the `import`s in `__init__.py` ?
> 
> Which do you prefer?

`__all__` would be good I think.

shinich1 (approved):

EarlMilktea:
@thierry-martinez

Could you review this PR?
I would like to work on refactoring `pauli.py` after this PR.

graphix/__init__.py
```diff
@@ -1,7 +1,9 @@
 """Optimize and simulate measurement-based quantum computation."""
 
 from graphix.generator import generate_from_graph
-from graphix.graphsim import GraphState
+from graphix.graphsim.graphstate import GraphState
```
thierry-martinez, line 4: Could you clarify the rationale behind this change? I would actually prefer to keep `graphix.graphsim` as an abstraction layer that hides the exact layout of its submodules.

EarlMilktea:
@thierry-martinez

tests/test_pauli.py
```diff
@@ -1,25 +1,21 @@
...
 
 
 class TestPauli:
     @pytest.mark.parametrize(
         ("u", "p"),
         itertools.product(
-            graphix.pauli.UNITS,
-            graphix.pauli.LIST,
+            UNITS,
+            LIST,
```
thierry-martinez, line 18: I would prefer to use at least `pauli.LIST` while waiting for the refactoring of the Pauli namespace.

tests/test_pauli.py
```diff
@@ -38,7 +34,7 @@ def test_matmul(self, a: Pauli, b: Pauli) -> None:
     @pytest.mark.parametrize(
         ("plane", "s", "t", "clifford", "angle", "choice"),
         itertools.product(
-            graphix.pauli.Plane,
+            Plane,
             (False, True),
             (False, True),
             iter(Clifford),
```
thierry-martinez, line 40: I just realized that `iter` is useless here.
```suggestion
            Clifford,
```

thierry-martinez (changes requested):
I think we should keep the prefix `pauli.LIST`, while there is no better namespace for Pauli.

tests/test_pauli.py
EarlMilktea: I will soon (literally soon!) submit another PR for this.
Anyway let me use your idea here.

graphix/__init__.py
EarlMilktea: Two reasons.

1. `mypy` warning

It was marked as invalid by `mypy` until `__all__` attributes are added in this PR.

2. `graphix.graphsim.graphstate` is not hidden

If we really strive to use `graphix.graphsim` as an abstraction layer, we should rename `.graphstate` to `._graphstate` (private module).
I would like to leave it as-is since it is out of the scope of this PR.

thierry-martinez: Ok, thanks!

tests/test_pauli.py
EarlMilktea: Fixed in 7cf2e8383b0d905dd20b3cf0e275bf7523ed8c4c .

thierry-martinez (approved):

PR #222, opened 2024-10-13
Use standard python `.gitignore`

EarlMilktea:

Use https://github.com/github/gitignore/blob/main/Python.gitignore + custom stuffs for this project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,21 +1,167 @@
-**/__pycache__
-.pytest_cache
-.benchmarks
-**egg-info
-**/.DS_Store
-**/build
-dist/*
-docs/source/gallery/**
-docs/source/benchmarks/**
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,21 +1,167 @@
-**/__pycache__
-.pytest_cache
-.benchmarks
-**egg-info
-**/.DS_Store
-**/build
-dist/*
-docs/source/gallery/**
-docs/source/benchmarks/**
```
shinich1, line 9: why not keep them? they're rather useful.

EarlMilktea: > ```gitignore
> **/__pycache__
> .pytest_cache
> **egg-info
> **/build
> dist/*
> docs/source/gallery/**
> docs/source/benchmarks/**
> ```

These entries are still included in the new file.

> ```gitignore
> .benchmarks
> ```

For this one, I'm not sure why this is necessary.

Do you remember why this was added?

> ```gitignore
> **/.DS_Store
> ```

This entry is specific to MacOS: as `.gitignore` should only contain what we want *every* contributors to ignore, it's not appropriate to include this line.

Note that this also applies to `.vscode/` or similar editor configs.

EarlMilktea: Resolved in bc14615dc405e1f8b55fb32b4687bcc602996096 .

shinich1 (approved):

PR #225, opened 2024-10-14
Fix #215: Handle non-contiguous numbering in `rustworkx` backend

### LINKED ISSUE #215: [Bug]: `rustworkx` fails to handle non-contiguous node numbering
The following code raises `KeyError` in `rxgraphviews.get_node_index` when `use_rustworkx` is `True`:
```python
    @pytest.mark.parametrize("use_rustworkx", [False, True])
    def test_pauli_non_contiguous(self, use_rustworkx) -> None:
        pattern.extend([
            N(node=2, state=PlanarState(plane=Plane.XY, angle=0.0)),
            E(nodes=(0, 2)),
            M(node=0, plane=Plane.XY, angle=0.0, s_domain=set(), t_domain=set())
        ])
        pattern.perform_pauli_measurements(use_rustworkx=use_rustworkx)
```

The bug seems to be between `RXGraphState.neighbors`, `RXGraphState.subgraphs` and `RXGraphState.local_complement`: the indexes returned by `neighbors` are in the rustworkx numbering, whereas `local_complement` calls `subgraphs` on these nodes, which expects them to be in the graphix numbering.

A fix could be to add a `idx_to_num` mapping in `rxgraphviews` and convert neighbors in the graphix numbering.

Related issues: #206 (problem with `rxgraphviews`).

thierry-martinez:
This commit resolves a bug, as tested in the new test case `test_pauli_non_contiguous`, which previously passed with the `networkx` backend but failed with the `rustworkx` backend.

The method `BaseGraphState.neighbors` now returns an iteration over (Graphix) node indices, rather than `rustworkx` vertex indices.

These indices are then passed to `BaseGraphState.subgraph` in `BaseGraphState.local_complement`. `BaseGraphState.subgraph` expects node indices, as it converts them back to `rustworkx` vertex indices.


graphix/graphsim/rxgraphstate.py
```diff
@@ -81,7 +81,7 @@ def neighbors(self, node) -> Iterator:
         See :meth:`BaseGraphState.neighbors`.
         """
         nidx = self.nodes.get_node_index(node)
-        return iter(self._graph.neighbors(nidx))
+        return iter(self.nodes.idx_to_num[idx] for idx in self._graph.neighbors(nidx))
```
EarlMilktea, line 84: `iter` might be redundant as genexp itself is iterator.

tests/test_pattern.py
```diff
@@ -78,6 +78,18 @@ def test_minimize_space(self, fx_rng: Generator) -> None:
         state_mbqc = pattern.simulate_pattern(rng=fx_rng)
         assert np.abs(np.dot(state_mbqc.flatten().conjugate(), state.flatten())) == pytest.approx(1)
 
+    @pytest.mark.parametrize("use_rustworkx", [False, True])
+    def test_pauli_non_contiguous(self, use_rustworkx) -> None:
```
EarlMilktea, line 82:
```suggestion
    def test_pauli_non_contiguous(self, use_rustworkx: bool) -> None:
```

EarlMilktea (changes requested):

graphix/graphsim/rxgraphstate.py
thierry-martinez: Fixed in 33b2941. Thanks!

EarlMilktea (approved):

shinich1 (approved):

EarlMilktea:
@thierry-martinez 

I merged `master` to run CI against this PR.

thierry-martinez:
Thanks. Everything is green: may I merge?

thierry-martinez:
Merged! Thanks.

PR #237, opened 2024-11-23
Refactor `BaseM`

EarlMilktea:
**Context (if applicable):**

Related to #178 .

**Description of the change:** (WIP)

- Respect the original intention to introduce `BaseM`

### DIFF
--- a/graphix/command.py
+++ b/graphix/command.py
@@ -51,10 +51,19 @@ class N(_KindChecker):
 
 
 @dataclasses.dataclass
-class M(_KindChecker):
-    """Measurement command. By default the plane is set to 'XY', the angle to 0, empty domains and identity vop."""
+class BaseM:
+    """Base class for measurement command.
+
+    This class does not contain any data except for the node to measure.
+    """
 
     node: Node
+
+
+@dataclasses.dataclass
+class M(BaseM, _KindChecker):
+    """Measurement command. By default the plane is set to 'XY', the angle to 0, empty domains and identity vop."""
+
     plane: Plane = Plane.XY
     angle: float = 0.0
     s_domain: set[Node] = dataclasses.field(default_factory=set)
@@ -136,8 +145,6 @@ if sys.version_info >= (3, 10):
 else:
     Command = Union[N, M, E, C, X, Z, S, T]
     Correction = Union[X, Z]
-
-BaseM = M
 
 
 @dataclasses.dataclass


EarlMilktea:
@jemappellesami 

I opened this PR to respect your initial motivation to introduce `BaseM` (add a way to simply specify the node subject to measure and nothing else), which is not properly reflected in the current code due to refactoring in process.

In addition to that, I have the following suggestion:

- Remove `MeasureMethod` and related things

I feel it's clearer to use `BaseM` (or something derived from this) to specify *everything* about measurements, as the data required to perform measurement are essentially the same (angle, plane) though they can be missing at first.
I would like to know in detail how you are trying to utilize the simulators *without* specifying the details, even while the simulators are initially designed for non-blind measurements.

graphix/command.py
```diff
@@ -51,10 +51,19 @@ class N(_KindChecker):
 
 
 @dataclasses.dataclass
-class M(_KindChecker):
-    """Measurement command. By default the plane is set to 'XY', the angle to 0, empty domains and identity vop."""
+class BaseM:
+    """Base class for measurement command.
+
+    This class does not contain any data except for the node to measure.
+    """
```
thierry-martinez, line 58: The [documentation](https://github.com/TeamGraphix/graphix/commit/ad182700f4788771e8ff0a8581ba4468bca1e518#diff-30c3360a99df7f4c7b7d4e732b003eb6187cdec903d6b4a34d5c686b3b2d55f0R50-R60) added in commit ad182700 and later deleted in #205 was more informative. I believe we should revert to that version.

```suggestion
class BaseM:
    """Base measurement command.

    Represent a measurement of a node. In MBQC, a measure is an instance of `M`,
    with given plane, angles, and domains. In the context of blind computations,
    the server only knows which node is measured, and the parameters are given
    by the :class:`graphix.simulator.MeasureMethod` provided by the client.
    """

    """
```

EarlMilktea: I intentionally trimmed them to ensure that no outdated info. is left.
Of course willing to revert to the original after discussion.

mgarnier59:
As far as I remember we didn't agree on anything regarding measurement commands. This strusture `BaseM`and `MeasureMethod` has been done _on purpose_ for our needs and we heavily rely on this.

Furthermore, your PR message 
> Respect the original intention to introduce `BaseM`
is cryptic to me.

EarlMilktea:
> The structure `BaseM` and `MeasureMethod` has been done on purpose for our needs

I opened this PR to improve this structure itself before aggressively refactoring simulators.
I know we need to care about backward compatibility, but I *never* believe that we can implement efficient, usable, and typed simulators without introducing breaking changes.
Additionally I would like to emphasize that, though I created it without discussion I never intend to merge it without discussion: that's why I'm using draft PR.
Please view this PR as a developer-friendly issue (can suggest by example) or something.

PR #241, opened 2025-01-06
Cleaning up device interface

thierry-martinez:
In the discussion on #230, there was a comment about improving the device interface implementation (see
https://github.com/TeamGraphix/graphix/pull/230#discussion_r1857006471 ).

In particular, I removed the `except Exception` path, being unable to reproduce an example where this path is reached. If we want to handle some errors (such as `TranspilerError`?), we should be more specific.

### DIFF
--- a/graphix/device_interface.py
+++ b/graphix/device_interface.py
@@ -17,7 +17,7 @@ class PatternRunner:
     Executes the measurement pattern.
     """
 
-    def __init__(self, pattern: Pattern, backend: str = "ibmq", **kwargs) -> None:
+    def __init__(self, pattern: Pattern, backend: str = "ibmq", instance: str | None = None, resource: str | None = None, save_statevector: bool = False, optimizer_level: int = 1, shots: int = 1024) -> None:
         """Instantiate a pattern runner.
 
         Parameters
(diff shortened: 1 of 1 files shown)


graphix/device_interface.py
```diff
@@ -26,35 +26,39 @@ def __init__(self, pattern: Pattern, backend: str = "ibmq", **kwargs) -> None:
...
-                save_statevector = kwargs.get("save_statevector", False)
-                optimization_level = kwargs.get("optimizer_level", 1)
-                self.backend.to_qiskit(save_statevector)
-                self.shots = kwargs.get("shots", 1024)
+            kwargs_get_backend = {}
+            if instance is not None:
+                kwargs_get_backend["instance"] = instance
+            if resource is not None:
+                kwargs_get_backend["resource"] = resource
+            self.backend.get_backend(**kwargs_get_backend)
```
shinich1, line 58: note: interface here is different in the latest commit of `graphix-ibmq` that is compatible with qiskit 1.0, to be merged soon.

graphix/device_interface.py
```diff
@@ -17,7 +17,7 @@ class PatternRunner:
     Executes the measurement pattern.
     """
 
-    def __init__(self, pattern: Pattern, backend: str = "ibmq", **kwargs) -> None:
+    def __init__(self, pattern: Pattern, backend: str = "ibmq", instance: str | None = None, resource: str | None = None, save_statevector: bool = False, optimizer_level: int = 1, shots: int = 1024) -> None:
```
EarlMilktea, line 20: This update sadly does not contribute to type-stable interface (better than `**kwargs` though).
My ideas:

- define ABC for backends
- remove `= "ibmq"` : we do not need to add information about the name of the class inheriting `ABC`.

thierry-martinez:
Subsumed by #261.

Issue #239, opened 2024-12-06
Pretty print for circuit and patterns


benjvmin93:
**Describe the feature you'd like**
A nice way to print `Circuit` and `Pattern` objects.

1. Ascii art (for circuits only)
2. LaTeX representation (using Quantikz package for circuits or qpic package)

See QIP renderers section from QuTiP paper: https://arxiv.org/abs/2412.04705
See also: https://github.com/qpic/qpic

Ongoing work at: https://github.com/benjvmin93/graphix/tree/pretty-print

thierry-martinez:
Closed by #277.

PR #245, opened 2025-02-05
Pretty print

benjvmin93:

- Format added code by `ruff`

Then, please fill in below:

**Context (if applicable):**

https://github.com/TeamGraphix/graphix/issues/239

New `draw()` methods in `Circuit` and `Pattern` classes.
- `Circuit.draw()` generates the corresponding Qasm3 code and pass it to `qiskit.draw()` method. It relates on the same arguments (see `qiskit.draw()` method documentation).
- `pattern.draw()` locally generates the appropriate visualization object (in the ascii, unicode, png, or latex formats)
- `pattern.__repr__()` returns the pythonic representation useful to initialize the same object. It involves handling `__repr__` method for commands. Thus, directly related to the writing of a `class Command`.

More details on the different changes are described in the `CHANGELOG.md` file.

**Related issue:**


benjvmin93:
One last thing that prevents the checks to be successful is the adding of qiskit module for drawing circuit. 
The `test_ibmq_runner` test is not skipped anymore since the qiskit module is present and fails because of the absence of the `save_statevector()` method, which I suppose, is due to Qiskit changes in the new versions.

I didn't want to change anything regarding this, since my work didn't involve anything in the runner tests so maybe we should discuss how to correct this issue to make sure we agree on what to do. Maybe either by replacing this method with the new appropriate one (in the `Aer` module I think, so without calling it from a `QuantumCircuit` object) or even treating this issue in a new PR.

@thierry-martinez

shinich1:
@benjvmin93 was qiskit added to required module in this PR? I agree updating tests to be compatible with latest qiskit would be good. 
@thierry-martinez what do you think is the first thing to do? completing https://github.com/TeamGraphix/graphix-ibmq/pull/16 before updating tests etc in this repo?

graphix/command.py
```diff
@@ -20,6 +20,62 @@
 Node = int
 
 
+def command_to_latex(cmd: Command) -> str:
```
EarlMilktea, line 23: Maybe better to be implemented as methods?

graphix/pattern.py
```diff
@@ -201,6 +209,114 @@ def __eq__(self, other: Pattern) -> bool:
             and self.output_nodes == other.output_nodes
         )
 
+    def _latex_file_to_image(self, tmpdirname, tmpfilename) -> PIL.Image.Image:
+        """Convert a latex file located in `tmpdirname/tmpfilename` to an image representation."""
+        import PIL
+
+        try:
+            subprocess.run(
```
EarlMilktea, line 217: Let's setup ruff rule S to check insecure operations.

graphix/transpiler.py
```diff
@@ -76,6 +80,43 @@ def __init__(self, width: int):
         self.instruction: list[instruction.Instruction] = []
         self.active_qubits = set(range(width))
 
+    def __str__(self) -> str:
+        """Return a string representation of the Circuit."""
+        return self.draw()
+
+    def draw(self, output: str = "text") -> TextDrawing | matplotlib.figure | PIL.Image | str:
```
EarlMilktea, line 87: Let's use `@typing.overload` for better UX.

requirements.txt
```diff
@@ -6,7 +7,9 @@ networkx>3.0
 numpy>=1.24,<2
 opt_einsum>=3.2
 pydantic
+pylatexenc
```
EarlMilktea, line 10: Please make this optional since all users need to set up LaTeX if pylatexenc is listed here.

requirements.txt
```diff
@@ -6,7 +7,9 @@ networkx>3.0
...
 quimb>=1.4.0
 scipy
 sympy>=1.9
 typing_extensions
+qiskit_qasm3_import
```
EarlMilktea, line 15: Same as above (import only when necessary)

tests/test_pattern.py
```diff
@@ -668,3 +669,22 @@ def test_remove_qubit(self) -> None:
 
 def assert_equal_edge(edge: Sequence[int], ref: Sequence[int]) -> bool:
     return any(all(ei == ri for ei, ri in zip(edge, other)) for other in (ref, reversed(ref)))
+
+
+def test_draw_pattern():
```
EarlMilktea, line 674:
```suggestion
def test_draw_pattern() -> None:
```

tests/test_pattern.py
```diff
@@ -668,3 +669,22 @@ def test_remove_qubit(self) -> None:
...
+    randpat = rand_circuit(5, 5).transpile().pattern
+    try:
+        randpat.draw("ascii")
+        randpat.draw("unicode")
+    except Exception as e:
+        pytest.fail(str(e))
+
+
+@pytest.mark.skipif(which("latex") is None, reason="latex not installed")
+def test_draw_pattern_latex():
```
EarlMilktea, line 684:
```suggestion
def test_draw_pattern_latex() -> None:
```

EarlMilktea:
MEMO: It may better to make big packages optional (ex. qiskit, matplotlib, etc.) for better usability.

graphix/command.py
benjvmin93: I agree with you but there is no `class Command`. We have `class _KindChecker` which is not suitable to implement these methods.

thierry-martinez: We should leave things as is for now, and later open a PR to refactor `command.py` with a proper `Command` class.

thierry-martinez:
@benjvmin93, could you mark `test_ibmq_backend` as skipped for now, since the failure is not related to this PR? Simply add the following line before the test:

```
@pytest.mark.skip(reason="graphix-ibmq support is broken #251")
```

requirements.txt
thierry-martinez: Dependencies such as `pylatexenc` and `qiskit` should be moved to a dedicated `requirements-full.txt` file, which CI should check in the same way it currently checks `requirements.txt` and `requirements-dev.txt`.

requirements.txt
thierry-martinez: Dependencies such as `pylatexenc` and `qiskit` should be moved to a dedicated `requirements-full.txt` file, which CI should check in the same way it currently checks `requirements.txt` and `requirements-dev.txt`.

graphix/transpiler.py
thierry-martinez: Could you elaborate? Since the output type depends on the value of `output`, using `@overload` doesn't seem appropriate here (we could encode the mapping using `Literal` for typing output, but that seems like overkill).

graphix/pattern.py
```diff
@@ -201,6 +209,114 @@ def __eq__(self, other: Pattern) -> bool:
...
+
+            return self._latex_file_to_image(tmpdirname, tmpfilename)
+
+    def __str__(self) -> str:
+        """Return a string representation of the pattern."""
+        return "\n".join([command_to_str(cmd) for cmd in self.__seq])
+
+    def to_unicode(self) -> str:
+        """Return the unicode string representation of the pattern."""
+        return "".join([command_to_unicode(cmd) for cmd in self.__seq])
```
thierry-martinez, line 318: I think inserting a space between commands (`" "` instead of `""`) would be the output more readable.

tests/test_transpiler.py
```diff
@@ -130,3 +132,22 @@ def simulate_and_measure() -> int:
...
+
+
+@pytest.mark.skipif(which("latex") is None, reason="latex not installed")
+def test_circuit_draw_latex() -> None:
+    circuit = Circuit(10)
+    try:
+        circuit.draw("latex")
+        circuit.draw("latex_source")
+    except Exception as e:
+        pytest.fail(str(e))
```
thierry-martinez, line 153: I would like the following test to pass:
```python
def test_to_qasm3(fx_bg: PCG64, jumps: int) -> None:
    nqubits = 5
    depth = 4
    circuit = rand_circuit(nqubits, depth, rng)
    qasm = circuit.to_qasm3()
    import pyzx as zx
    print(qasm)
    z = zx.qasm(qasm)
    g = z.to_graph()
    og = OpenGraph.from_pyzx_graph(g)
    pattern = to_pattern(og)
    state = circuit.simulate_statevector().statevec
    state_mbqc = pattern.simulate_pattern()
    assert np.abs(np.dot(state_mbqc.flatten().conjugate(), state.flatten())) == pytest.approx(1)
```

tests/test_pattern.py
```diff
@@ -668,3 +669,22 @@ def test_remove_qubit(self) -> None:
...
+    randpat = rand_circuit(5, 5).transpile().pattern
+    try:
+        randpat.draw("latex")
+        randpat.draw("png")
+    except Exception as e:
+        pytest.fail(str(e))
```
thierry-martinez, line 690: I would like the following tests to pass (the output can be adjusted):
```python
def test_draw_pattern_j_alpha():
    p = Pattern()
    p.add(N(1))
    p.add(N(2))
    p.add(E((1, 2)))
    p.add(M(1))
    p.add(X(2, domain={1}))
    assert str(p) == "N(1) N(2) E(1,2) M(1) X(2,{1})"
    assert p.to_unicode() == "N₁ N₂ E₁₋₂ M₁ X₂¹"
    assert p.to_latex() == r"\(N_{1} N_{2} E_{1,2} M_{1} X_{2}^{1}\)"

def test_draw_pattern_measure():
    p = Pattern()
    p.add(N(1))
    p.add(N(2))
    p.add(M(1, Plane.YZ, 0.5))
    p.add(M(2, Plane.XZ, -0.25))
    p.add(M(3, Plane.XY, 0.1, s_domain={1}, t_domain={2}))
    assert str(p) == "N(1) N(2) M(1,YZ,pi/2) M(2,XZ,-pi/4) [1]M(3,0.31)[2]"
    assert p.to_unicode() == "N₁ N₂ M₁(YZ,π/2) M₂(XZ,-π/4) ₁M₃(0.31)²"
    assert p.to_latex() == r"\(N_{1} N_{2} E_{1,2} M_{1}^{YZ,\frac{\pi}{2}} M_{2}^{XZ,-\frac{\pi}{4}} {}_1[M_{3}^{0.31}]^2\)"
```

thierry-martinez (changes requested):
I noticed some bugs and opportunities for readability improvements; please see my individual comments.

tests/test_pattern.py
thierry-martinez: I should have written:
```
assert str(p) == "N(1) N(2) M(1,YZ,pi/2) M(2,XZ,-pi/4) {1}[M(3,0.31)]{2}"
assert p.to_unicode() == "N₁ N₂ M₁(YZ,π/2) M₂(XZ,-π/4) ₁[M₃(0.31)]²"
```

thierry-martinez: assert p.to_latex() == r"\(N_{1}\,N_{2}\,E_{1,2}\,M_{1}^{YZ,\frac{\pi}{2}}\,M_{2}^{XZ,-\frac{\pi}{4}}\,{}_2[M_{3}^{0.31}]^1\)"

graphix/command.py
thierry-martinez: See #253.

tests/test_transpiler.py
benjvmin93: It seems like my version of Graphix doesn't have the `from_pyzx_graph()` method, would it be another way to do this ?

requirements.txt
thierry-martinez: Actually, there is already a `requirements-extra.txt` file, which if I understand correctly corresponds to what I would like to put in `requirements-full.txt`. There is `graphix-ibmq` already in it, and you could add `pylatexenc` as well in it. However, `requirements-extra.txt` aren't currently checked by the CI: we should change `".[dev]"` in `noxfile.py` into `".[dev,extra]"`.

graphix/transpiler.py
EarlMilktea: > overkill

I don't think so. In general we should avoid union types in return type annotations, or users must manually narrow the type to use class-specific members. Not users but we developers should bear the burden of writing long codes.

.github/workflows/ci.yml
```diff
@@ -29,6 +29,49 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
 
+      - name: Setup TeX Live
```
EarlMilktea, line 32: Please check that this stage is somehow "cached" to reduce server workloads.

.github/workflows/cov.yml
```diff
@@ -18,11 +18,26 @@ jobs:
         with:
           python-version: "3.12"
 
+      - name: Setup TeX Live
+        uses: teatimeguest/setup-texlive-action@v3
```
EarlMilktea, line 22: Same as above

graphix/command.py
```diff
@@ -22,6 +22,165 @@
...
+SUBSCRIPTS = str.maketrans("0123456789+", "₀₁₂₃₄₅₆₇₈₉₊")
+SUPERSCRIPTS = str.maketrans("0123456789+", "⁰¹²³⁴⁵⁶⁷⁸⁹⁺")
+
+
+def _angle_to_str(angle: ExpressionOrFloat, latex: bool = False) -> str:
+    if not isinstance(angle, float):
+        return str(angle)
+
+    angle_map = {np.pi: "π", np.pi / 2: "π/2", np.pi / 4: "π/4"}
+    angle_map_latex = {np.pi: "\\pi", np.pi / 2: "\\frac{\\pi}{2}", np.pi / 4: "\\frac{\\pi}{4}"}
```
EarlMilktea, line 34: Please use raw string for readability.

graphix/command.py
```diff
@@ -52,6 +211,10 @@ class N(_KindChecker):
     state: State = dataclasses.field(default_factory=lambda: BasicStates.PLUS)
     kind: ClassVar[Literal[CommandKind.N]] = dataclasses.field(default=CommandKind.N, init=False)
 
+    def __repr__(self) -> str:
+        """Return the representation of a N command."""
+        return f"N(node={self.node})"
```
EarlMilktea, line 216: MEMO: `=` formatting may help:

```py
>>> node = 3
>>> f"{node=}"
'node=3'
```

graphix/pattern.py
```diff
@@ -195,9 +202,11 @@ def reorder_input_nodes(self, input_nodes: list[int]):
     # TODO: This is not an evaluable representation. Should be __str__?
     def __repr__(self) -> str:
         """Return a representation string of the pattern."""
-        return (
-            f"graphix.pattern.Pattern object with {len(self.__seq)} commands and {len(self.output_nodes)} output qubits"
-        )
+        r = [
+            f"Pattern({'' if not self.input_nodes else f'input_nodes={self.input_nodes}'})",
+            f"self.__seq = {self.__seq}",
```
EarlMilktea, line 207: Please do not expose private member `__seq` .

graphix/pattern.py
```diff
@@ -207,6 +216,151 @@ def __eq__(self, other: Pattern) -> bool:
...
+
+        try:
+            subprocess.run(
+                [
+                    "pdflatex",
+                    "-halt-on-error",
+                    f"-output-directory={tmpdirname}",
+                    f"{tmpfilename}.tex",
+                ],
+                stdout=subprocess.PIPE,
```
EarlMilktea, line 231: Let's use `capture_output` instead of pipes.

graphix/pattern.py
```diff
@@ -207,6 +216,151 @@ def __eq__(self, other: Pattern) -> bool:
...
+                stderr=subprocess.DEVNULL,
+                check=True,
+            )
+        except OSError as exc:
+            # OSError should generally not occur, because it's usually only triggered if `pdflatex`
+            # doesn't exist as a command, but we've already checked that.
+            raise Exception("`pdflatex` command could not be run.") from exc
+        except subprocess.CalledProcessError as exc:
+            with open("latex_error.log", "wb") as error_file:
```
EarlMilktea, line 240: Please use pathlib (ruff rule PTH may help)

graphix/pattern.py
EarlMilktea: Have you looked for latex wrappers? I feel the approach here (raw subprocess call) is a little bit too direct.

graphix/pattern.py
```diff
@@ -207,6 +216,151 @@ def __eq__(self, other: Pattern) -> bool:
...
+\begin{document}
+"""
+
+        output = io.StringIO()
+        output.write(header_1)
+        output.write(header_2)
+
+        output.write(self.to_latex(left_to_right))
+
+        output.write("\n\\end{document}")
```
EarlMilktea, line 312: Raw string

graphix/pattern.py
```diff
@@ -207,6 +216,151 @@ def __eq__(self, other: Pattern) -> bool:
...
+        left_to_right: bool
+            whether or not represent the pattern from left to right representation. Default is left to right, otherwise it's right to left
+        """
+        tmpfilename = "pattern"
+
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            tmppath = Path(tmpdirname) / tmpfilename
+            tmppath = tmppath.with_suffix(".tex")
+
+            with open(tmppath, "w") as latex_file:
```
EarlMilktea, line 332: pathlib

graphix/pattern.py
```diff
@@ -207,6 +216,151 @@ def __eq__(self, other: Pattern) -> bool:
...
+
+    def to_unicode(self, left_to_right: bool = True) -> str:
+        """Return the unicode string representation of the pattern.
+
+        Parameters
+        ----------
+        left_to_right: bool
+            whether or not represent the pattern from left to right representation. Default is left to right, otherwise it's right to left
+        """
+        seq = self.__seq[::-1] if not left_to_right else self.__seq
```
EarlMilktea, line 361: `reversed` may help here.

EarlMilktea:
@benjvmin93 Could you turn on ruff rule `PTH` before final review?

.github/workflows/ci.yml
benjvmin93: As specified in https://github.com/teatimeguest/setup-texlive-action, cache should be set by default by the action.

graphix/command.py
EarlMilktea: Either will do.

shinich1:
if possible, can we move the implementation of drawing methods out of `Pattern` class (`pattern` module)? This has become a bit too large.

EarlMilktea:
@benjvmin93 (CC: @shinich1 )

Additionally, if you're planning to add bunch of features to `Pattern`, I may request you to turn on `mypy` check against the file.

@thierry-martinez 

Are there any updates on `Pattern` refactoring? It would be better to merge it before this large diff. is introduced.

PR #261, opened 2025-03-31
abstract class for device interface

d1ssk:

- Format added code by `ruff`

Then, please fill in below:

**Context (if applicable):**
Refactoring and modularization of the device backend interface is required.

- Refactored the `device_interface` API structure into a formal abstract class system.
- Defined abstract classes in `graphix.device_interface`:
  - `DeviceBackend`: an interface for executing MBQC patterns on a backend (real or simulated).
  - `JobHandler`: an interface for managing job lifecycle (status, cancellation, result retrieval).
- These abstract interfaces are now the only contracts `graphix` defines for device execution.
- All device-specific implementations (e.g. for IBMQ) are completely delegated to external packages such as `graphix-ibmq`, which implement and register concrete subclasses.
- Introduced `graphix.backend_factory.get_backend(name: str)`:
  - Supports lazy import of external device backends (e.g., `name="ibmq"`).
  - This allows preserving one-way dependency flow from external backend libraries to `graphix`, while still enabling centralized backend access for users.

**Related issue:**

### DIFF
--- a/graphix/device_backend.py
+++ b/graphix/device_backend.py
@@ -1,0 +1,18 @@
+from typing import Optional
+from graphix.device_interface import DeviceBackend
+from graphix.pattern import Pattern
+
+
+def get_backend(name: str, pattern: Optional[Pattern] = None, **kwargs) -> DeviceBackend:
+    if name == "ibmq":
+        try:
+            from graphix_ibmq.backend import IBMQBackend
+        except ImportError:
+            raise ImportError("Please install graphix-ibmq via `pip install graphix-ibmq`.")
+
+        backend = IBMQBackend(**kwargs)
+        if pattern is not None:
+            backend.set_pattern(pattern)
+        return backend
+    else:
+        raise ValueError(f"Unknown backend: {name}")
--- a/graphix/pattern.py
+++ b/graphix/pattern.py
@@ -18,7 +18,7 @@ import typing_extensions
 from graphix import command, parameter
 from graphix.clifford import Clifford
 from graphix.command import Command, CommandKind
-from graphix.device_interface import PatternRunner
+# from graphix.device_interface import PatternRunner
 from graphix.fundamentals import Axis, Plane, Sign
 from graphix.gflow import find_flow, find_gflow, get_layers
 from graphix.graphsim.graphstate import GraphState
@@ -1282,26 +1282,6 @@ class Pattern:
         sim = PatternSimulator(self, backend=backend, **kwargs)
         sim.run(input_state)
         return sim.backend.state
-
-    def run_pattern(self, backend, **kwargs):
-        """Run the pattern on cloud-based quantum devices and their simulators.
-
-        Available backend: ['ibmq']
-
-        Parameters
-        ----------
-        backend : str
-            parameter to select executor backend.
-        kwargs: keyword args for specified backend.
-
-        Returns
-        -------
-        result :
-            the measurement result,
-            in the representation depending on the backend used.
-        """
-        exe = PatternRunner(self, backend=backend, **kwargs)
-        return exe.run()
 
     def perform_pauli_measurements(
         self, leave_input: bool = False, use_rustworkx: bool = False, ignore_pauli_with_deps: bool = False
--- a/graphix/device_interface.py
+++ b/graphix/device_interface.py
@@ -1,125 +1,42 @@
-"""Quantum hardware device interface.
-
-Runs MBQC command sequence on quantum hardware.
-"""
-
-from __future__ import annotations
-
-from typing import TYPE_CHECKING, Any
-
-if TYPE_CHECKING:
-    from graphix.pattern import Pattern
+from abc import ABC, abstractmethod
+from typing import Any, Optional
+from graphix.pattern import Pattern
 
 
-class PatternRunner:
-    """MBQC pattern runner.
+class JobHandle(ABC):
+    @abstractmethod
+    def get_id(self) -> str:
+        pass
 
-    Executes the measurement pattern.
-    """
+    @abstractmethod
+    def is_done(self) -> bool:
+        pass
 
-    def __init__(self, pattern: Pattern, backend: str = "ibmq", **kwargs) -> None:
-        """Instantiate a pattern runner.
+    @abstractmethod
+    def cancel(self) -> None:
+        pass
 
-        Parameters
-        ----------
-        pattern: :class:`graphix.pattern.Pattern` object
-            MBQC pattern to be executed.
-        backend: str
-            execution backend (optional, default is 'ibmq')
-        kwargs: dict
-            keyword args for specified backend.
-        """
+
+class CompileOptions(ABC):
+    """Base class for compile options."""
+    pass
+
+
+class DeviceBackend(ABC):
+    def __init__(self) -> None:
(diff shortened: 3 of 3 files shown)


graphix/device_backend.py
```diff
@@ -0,0 +1,18 @@
+from typing import Optional
+from graphix.device_interface import DeviceBackend
+from graphix.pattern import Pattern
+
+
+def get_backend(name: str, pattern: Optional[Pattern] = None, **kwargs) -> DeviceBackend:
```
EarlMilktea, line 6: `**kwargs` is evil. Please list up all the possible arguments.

d1ssk:
@EarlMilktea Thank you for your comment. I decided to remove the `get_backend` functionality after comparing the convenience it provides against the reduced clarity in code dependencies. I’ve also addressed the other points you mentioned.

thierry-martinez (changes requested):
These changes appear to be linked with modifications in the [[device_interface_abc](https://github.com/TeamGraphix/graphix-ibmq/tree/device_interface_abc)](https://github.com/TeamGraphix/[graphix-ibmq](https://github.com/TeamGraphix/graphix-ibmq)/tree/device_interface_abc) branch of [`graphix-ibmq`](https://github.com/TeamGraphix/graphix-ibmq). The overall idea seems to be moving most implementation details to `graphix-ibmq` while keeping only abstract classes in `graphix`. Thanks for this work—it leads to better organization and fits with our goal of reducing code in `graphix` when it can be better handled by specific plugins.

Note that tests previously performed in `test_runner.py` have been removed. The new tests in `test_device_interface.py` only verify that the classes can be instantiated; they do not check their functionality. I suggest:

- Adding tests in the [`[noxfile.py](https://github.com/TeamGraphix/graphix/blob/device_interface_abc/noxfile.py)`](https://github.com/TeamGraphix/graphix/blob/device_interface_abc/noxfile.py) to ensure in the CI that both the `graphix` and `graphix-ibmq` repositories are in sync—specifically, that `pytest` passes for `graphix-ibmq` (in the same manner that `tests_symbolic` verifies that `pytest` passes for `graphix-symbolic`).
- Including the `test_ibmq_backend` test from `test_runner.py` in the `graphix-ibmq` test suite.

I also have some remarks about the design:

- `CompileOptions` should be merged with `DeviceBackend` to avoid maintaining two parallel class hierarchies and performing dynamic checks.
- There should be a dedicated class for the compiled object to ensure that a `DeviceBackend` does not carry a pattern and a compiled object that do not match.
- `JobHandler` should be renamed to `Job` and should be able to retrieve its own results.

Finally, I have a more general design question: Do we need an abstract `DeviceBackend` class at all? This class does not appear to be used anywhere else in `graphix`, and I do not see a use case for reasoning generically about a `DeviceBackend`. Wouldn't it be sufficient to remove the `DeviceBackend` class from `graphix` and simply implement an `IBMQDeviceBackend` in the `graphix-ibmq` repository, with other backends defined elsewhere as needed? What is the benefit of having a common interface in this case?

d1ssk:
@thierry-martinez Thank you for your review.
> * Adding tests in the [`noxfile.py`](https://github.com/TeamGraphix/graphix/blob/device_interface_abc/noxfile.py) to ensure in the CI that both the `graphix` and `graphix-ibmq` repositories are in sync—specifically, that `pytest` passes for `graphix-ibmq` (in the same manner that `tests_symbolic` verifies that `pytest` passes for `graphix-symbolic`).
> * Including the `test_ibmq_backend` test from `test_runner.py` in the `graphix-ibmq` test suite.

I believe this point is important and plan to implement it next.

> * `CompileOptions` should be merged with `DeviceBackend` to avoid maintaining two parallel class hierarchies and performing dynamic checks.
> * There should be a dedicated class for the compiled object to ensure that a `DeviceBackend` does not carry a pattern and a compiled object that do not match.
> * `JobHandler` should be renamed to `Job` and should be able to retrieve its own results.

Since the compile options vary depending on the actual device, I think the current implementation is inevitable in order to maintain both robustness and simplicity. I would appreciate it if you clarify what you mean specifically by "integrating into DeviceBackend"?

I’ve decided to deprecate `set_pattern` and instead include the pattern as an argument to `compile`. The compiled result will now also contain the pattern. On IBMQ, the raw result includes measurements of ancillary qubits, so in order to extract results for logical qubits only, information about the original pattern is required. Therefore, I’ve decided to pass the entire compilation result to the `Job` that implements `retrieve_result`. If you have a better idea for how to handle this, I’d appreciate your input.

> Finally, I have a more general design question: Do we need an abstract `DeviceBackend` class at all? This class does not appear to be used anywhere else in `graphix`, and I do not see a use case for reasoning generically about a `DeviceBackend`. Wouldn't it be sufficient to remove the `DeviceBackend` class from `graphix` and simply implement an `IBMQDeviceBackend` in the `graphix-ibmq` repository, with other backends defined elsewhere as needed? What is the benefit of having a common interface in this case?

Since graphix is an upstream package, I believe it's meaningful for it to define what and how things should be implemented structurally in the downstream packages that handle specific devices.

tests/test_device_interface.py
```diff
@@ -0,0 +1,42 @@
+from __future__ import annotations
```
EarlMilktea, line 1: Please check this file by static type checkers.

EarlMilktea (changes requested):

EarlMilktea:
> everse dependency check

Why not kick the CI externally from this repo via HTTP, as in general reverse dependency is not preferable (and this PR resolves this issue, right?) ?

EarlMilktea:
Thank your for the issue. For now I'm happy with @thierry-martinez 's suggestion (add combined CI) as long as `graphix-ibmq` is not imported inside this package (including tests) .

graphix/pattern.py
```diff
@@ -18,7 +18,8 @@
 from graphix import command, parameter
 from graphix.clifford import Clifford
 from graphix.command import Command, CommandKind
-from graphix.device_interface import PatternRunner
+
+# from graphix.device_interface import PatternRunner
```
shinich1: remove?

shinich1 (approved):
@d1ssk thanks a lot for the PR and sorry for the huge delay.. I now checked with @thierry-martinez that graphix-perceval side is also fine with this direction. Please resolve conflict and address my comment before merging.

graphix/pattern.py
d1ssk: resolved in [fa9f302](https://github.com/TeamGraphix/graphix/pull/261/commits/fa9f3023b4476d3324d7ca030a22d0704495f5be)

docs/source/sg_execution_times.rst
```diff
@@ -0,0 +1,64 @@
+
```
EarlMilktea: Do not add this file. It's ignored in `.gitignore` .

EarlMilktea (changes requested):

docs/source/sg_execution_times.rst
d1ssk: resolved in [45abedf](https://github.com/TeamGraphix/graphix/pull/261/commits/45abedf7b04a060c221a9dc95a812d62bc3f1a64)

EarlMilktea (approved):

PR #263, opened 2025-04-10
Enforce `from __future__ import annotations`

EarlMilktea:

Enable ruff `I002` to ensure that all files have `from __future__ import annotations`, which is required for lazy evaluation of type annotations.


EarlMilktea:
MEMO: mypy errors are not related to this PR and fixed in #262 .

shinich1 (approved):

thierry-martinez (approved):
LGTM! Thanks

PR #267, opened 2025-04-16
Annotate `noise_models/*.py`

EarlMilktea:

Annotate `noise_models/*.py` and check them by `mypy` / `pyright` .


shinich1 (approved):
LGTM!

PR #270, opened 2025-04-21
Update linters

EarlMilktea:

- use `select = ["ALL"]` for blacklist-based `ruff` configurations
- minor CI fixes/updates
- use prettier to format text files (`.yaml` / `.md`)
- use `pre-commit` to check text files


graphix/pattern.py
```diff
@@ -319,10 +320,10 @@ def add_correction_domain(
                     new_cmd = new_cmd.clifford(clifford_gate)
                 if t_domain := z_dict.pop(cmd.node, None):
                     # The original domain should not be mutated
-                    new_cmd.t_domain = new_cmd.t_domain ^ t_domain
+                    new_cmd.t_domain ^= t_domain
                 if s_domain := x_dict.pop(cmd.node, None):
                     # The original domain should not be mutated
-                    new_cmd.s_domain = new_cmd.s_domain ^ s_domain
+                    new_cmd.s_domain ^= s_domain
```
thierry-martinez, line 326: The comments
```python
# The original domain should not be mutated
```
implies that updates are _not_ meant to happen in place. Indeed, the changes now break this test:
```python
    def test_standardize_copy(self) -> None:
        circuit = Circuit(1)
        circuit.h(0)
        circuit.rx(0, np.pi / 4)
        pattern_copy = copy.deepcopy(pattern)
        pattern2 = pattern.copy()
        pattern2.standardize()
        assert list(pattern) == list(pattern_copy)
```
Both `pattern2` and the original `pattern` share the same domains, that `pattern2.standardize()` modifies.

thierry-martinez (changes requested):
This PR improves things in a lot of places, thanks! I only spotted one aliasing issue in `standardize_direct`.

graphix/pattern.py
EarlMilktea: Nice catch! Indeed `--unsafe-fixes` didn't respect that comments. Fixed in 05014669bbc5d70af116f118f5f5968ef264fef4 .

README.md
```diff
@@ -83,7 +82,7 @@ state_out = pattern.simulate_pattern(backend="statevector")
 
 ## Citing
 
-> Shinichi Sunami and Masato Fukushima, Graphix. (2023) <https://doi.org/10.5281/zenodo.7861382> 
+> Shinichi Sunami and Masato Fukushima, Graphix. (2023) <https://doi.org/10.5281/zenodo.7861382>
 
 ## Contributing
```
shinich1, line 87: @EarlMilktea Could you add instructions for setting up environment for contribution (reference to contributing.md might be enough) here? I think we have lots of requirements now., with addition of `prettier` here for docs updates. Otherwise this PR look nice, thanks!

graphix/visualization.py
```diff
@@ -1171,14 +1171,14 @@ def _edge_intersects_node(start, end, node_pos, buffer=0.2):
     @staticmethod
     def _control_point(start, end, node_pos, distance=0.6):
         """Generate a control point to bend the edge around a node."""
-        edge_vector = np.array(end) - np.array(start)
+        edge_vector = np.fromiter((e - s for s, e in zip(start, end)), dtype=np.float64)
```
thierry-martinez, line 1174: Could you explain this change? The original code is more concise, easier to understand, and about 20 % faster in the following micro-benchmark.
```python
a = [i for i in range(1000)]
b = [i for i in range(0, 2000, 2)]
print(timeit(lambda: np.array(a) - np.array(b)))
print(timeit(lambda: np.fromiter((u - v for u, v in zip(a, b)), dtype=np.float64)))
```

tests/test_parameter.py
```diff
@@ -149,9 +149,8 @@ def test_density_matrix_xreplace() -> None:
     assert np.allclose(dm.xreplace({alpha: 1, beta: 2}).rho, np.array([[1, 2], [1, 2]]))
 
 
-@pytest.mark.parametrize("jumps", range(1, 11))
 @pytest.mark.parametrize("use_xreplace", [False, True])
-def test_random_circuit_with_parameters(fx_rng: Generator, jumps: int, use_xreplace: bool) -> None:
+def test_random_circuit_with_parameters(fx_rng: Generator, use_xreplace: bool) -> None:
```
thierry-martinez, line 153: The intention was to use `fx_bg` instead of `fx_rng`.
```suggestion
def test_random_circuit_with_parameters(fx_bg: PCG64, jumps: int, use_xreplace: bool) -> None:
    rng = Generator(fx_bg.jumped(jumps))
```

thierry-martinez (changes requested):
This looks good. I just have a couple of questions.

README.md
EarlMilktea: Now fixable linter errors are all handled by reviewdog-suggester.

graphix/visualization.py
EarlMilktea: Fixed in 00df966ffce140e043b3121777cb29d7e3f4cb82 . Actually the problem here was that element types are unknown.

tests/test_parameter.py
EarlMilktea: Fixed in edcdac27d261dc2a6a541e8f71771e9ef14ac9bc .

thierry-martinez (approved):

shinich1 (approved):

PR #271, opened 2025-04-22
Transpiler API for noise models

thierry-martinez:
Introduce a new API for noise models, which transpile perfect patterns
into noisy ones. Noisy patterns use an additional command, `A`, to
apply a noise operator to a subset of qubits.

The existing `Command` type remains unchanged and does not include `A`.
We introduce:

- `CommandOrNoise` — a union of all pattern commands plus `A`
- `NoiseCommands` — a sequence of `CommandOrNoise`, i.e. a noisy pattern

A noise model must implement the three abstract methods of `NoiseModel`:

- `input_nodes(self, nodes: list[int]) -> NoiseCommands`
  Noisy commands to apply to the input nodes.

- `command(self, cmd: CommandOrNoise) -> NoiseCommands`
  Generate a noisy pattern sequence for a single command. Noise may
  be inserted before or after the command, or the command itself
  may be modified (e.g., rerouted to a different qubit). The input
  can be either a perfect command or a noise command, enabling
  composition of multiple noise effects.

- `confuse_result(self, cmd: BaseM, result: bool) -> bool`
  Optionally modify the outcome of a measurement.

We provide two concrete noise models:

- `DepolarisingNoiseModel`
- `ComposeNoiseModel` — compose multiple noise models in sequence

Abstract base class `Noise` encapsulates a noise operator:

- `nqubits(self) -> int`
  Number of qubits targeted by this noise (must match the arity of `A`).

- `to_kraus_channel(self) -> KrausChannel`
  Returns the Kraus description, used by the density‑matrix backend.
  Backends can also inspect the `Noise` subclass (e.g. `DepolarisingNoise`)
  directly without parsing Kraus operators.


graphix/noise_models/depolarising_noise_model.py
```diff
@@ -0,0 +1,103 @@
...
+    ) -> None:
+        self.prepare_error_prob = prepare_error_prob
+        self.x_error_prob = x_error_prob
+        self.z_error_prob = z_error_prob
+        self.entanglement_error_prob = entanglement_error_prob
+        self.measure_error_prob = measure_error_prob
+        self.measure_channel_prob = measure_channel_prob
+        self.rng = ensure_rng(rng)
+
+    def input_nodes(self, nodes: list[int]) -> NoiseCommands:
```
EarlMilktea, line 76:
```suggestion
    def input_nodes(self, nodes: Iterable[int]) -> NoiseCommands:
```

graphix/noise_models/noise_model.py
```diff
@@ -2,63 +2,89 @@
...
+NoiseCommands = list[CommandOrNoise]
+
+
+class NoiseModel(ABC):
+    """Abstract base class for all noise models."""
 
-    @abc.abstractmethod
-    def confuse_result(self, result: bool) -> bool:
+    @abstractmethod
+    def input_nodes(self, nodes: list[int]) -> NoiseCommands:
```
EarlMilktea, line 53:
```suggestion
    def input_nodes(self, nodes: Iterable[int]) -> NoiseCommands:
```

graphix/noise_models/noise_model.py
```diff
@@ -2,63 +2,89 @@
...
+        return [n_cmd for cmd in sequence for n_cmd in self.command(cmd)]
+
+
+@dataclass(frozen=True)
+class ComposeNoiseModel(NoiseModel):
+    """Compose noise models."""
+
+    models: list[NoiseModel]
+
+    def input_nodes(self, nodes: list[int]) -> NoiseCommands:
```
EarlMilktea, line 75:
```suggestion
    def input_nodes(self, nodes: Iterable[int]) -> NoiseCommands:
```

graphix/noise_models/noiseless_noise_model.py
```diff
@@ -15,44 +16,14 @@ class NoiselessNoiseModel(NoiseModel):
...
-    @typing_extensions.override
-    def prepare_qubit(self) -> KrausChannel:
-        """Return the channel to apply after clean single-qubit preparation. Here just identity."""
-        return KrausChannel([KrausData(1.0, np.eye(2))])
-
-    @typing_extensions.override
-    def entangle(self) -> KrausChannel:
-        """Return noise model to qubits that happens after the CZ gates."""
-        return KrausChannel([KrausData(1.0, np.eye(4))])
+    def input_nodes(self, nodes: list[int]) -> NoiseCommands:
```
EarlMilktea, line 19:
```suggestion
    def input_nodes(self, nodes: Iterable[int]) -> NoiseCommands:
```

graphix/noise_models/depolarising_noise_model.py
```diff
@@ -0,0 +1,103 @@
+"""Depolarising noise model."""
```
EarlMilktea, line 1: Please add `@typing_extensions.override` to all the overriding method implementations.

graphix/sim/density_matrix.py
```diff
@@ -202,7 +203,7 @@ def expectation_single(self, op: npt.NDArray, i: int) -> complex:
 
         return np.trace(rho_tensor.reshape((2**self.nqubit, 2**self.nqubit)))
 
-    def dims(self) -> list[int]:
+    def dims(self) -> tuple[int, ...]:
```
EarlMilktea, line 206: Isn't it possible to somehow maintain the length of tuple?

graphix/sim/density_matrix.py
```diff
@@ -366,6 +369,11 @@ def apply_channel(self, channel: KrausChannel, qargs: Collection[int]) -> None:
         indices = [self.node_index.index(i) for i in qargs]
         self.state.apply_channel(channel, indices)
 
+    def apply_noise(self, nodes: list[int], noise: Noise) -> None:
```
EarlMilktea, line 372: `Iterable`

thierry-martinez: Fixed in d9f6e2a (with `Collection`, since nodes are supposed to be indexable in `apply_channel`).

graphix/simulator.py
```diff
@@ -60,7 +57,7 @@ def set_measure_result(self, node: int, result: bool) -> None:
 class DefaultMeasureMethod(MeasureMethod):
     """Default measurement method implementing standard measurement plane/angle update for MBQC."""
 
-    def __init__(self, results=None):
+    def __init__(self, results: dict[int, bool] | None = None) -> None:
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in d9f6e2a.

graphix/noise_models/noise_model.py
thierry-martinez: Fixed in d573e7f6.

graphix/noise_models/noise_model.py
```diff
@@ -2,63 +2,89 @@
...
-    from graphix.simulator import PatternSimulator
 
 
-class NoiseModel(abc.ABC):
-    """Abstract base class for all noise models."""
+class Noise(ABC):
+    """Abstract base class for noise."""
+
+    @abstractmethod
+    def nqubits(self) -> int:
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in a17adf7.

graphix/noise_models/depolarising_noise_model.py
```diff
@@ -0,0 +1,103 @@
...
+
+if TYPE_CHECKING:
+    from numpy.random import Generator
+
+
+@dataclass
+class DepolarisingNoise(Noise):
+    """One-qubit depolarising noise with probabibity `prob`."""
+
+    prob: float
```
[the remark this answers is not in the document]
thierry-martinez: I used a descriptor in 043d250.

graphix/noise_models/depolarising_noise_model.py
thierry-martinez: Fixed in 17bca10.

shinich1 (commented):
Thank you! could you,
- update changelog
- update docs
- update or add example

thierry-martinez:
I updated the CHANGELOG and the documentation. In particular, the documentation now includes an updated example on how to define a noise model. I don't know if it is worth adding an example in `examples/`, since what I can share is already in the documentation for now. I think more examples will come when we will implement more interesting noise models. @shinich1, are you ok if we wait for other PRs to have more examples? I can open an issue for that if needed.

graphix/noise_models/depolarising_noise_model.py
```diff
@@ -0,0 +1,118 @@
+"""Depolarising noise model."""
```
EarlMilktea: I feel the file name is too long.

graphix/noise_models/depolarising_noise_model.py
```diff
@@ -0,0 +1,118 @@
...
+
+    from graphix.measurements import Outcome
+
+
+@dataclass
+class DepolarisingNoise(Noise):
+    """One-qubit depolarising noise with probabibity ``prob``."""
+
+    # Probability is a descriptor
+    prob: Probability = Probability()  # noqa: RUF009
```
EarlMilktea: Please do not mix `dataclass` and custom descriptors!
As clearly written in fluent python, we must not use `dataclass` when it's not a simple "data class" .

graphix/noise_models/depolarising_noise_model.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __init__(
+        self,
+        prepare_error_prob: float = 0.0,
+        x_error_prob: float = 0.0,
+        z_error_prob: float = 0.0,
+        entanglement_error_prob: float = 0.0,
+        measure_channel_prob: float = 0.0,
+        measure_error_prob: float = 0.0,
+        rng: Generator | None = None,
```
EarlMilktea: Can't we pass `rng` when necessary rather than storing in `class` ?

graphix/random_objects.py
```diff
@@ -42,9 +41,7 @@ def rand_unit(sz: int, rng: Generator | None = None) -> npt.NDArray:
 UNITS = np.array([1, 1j])
 
 
-def rand_dm(
-    dim: int, rng: Generator | None = None, rank: int | None = None, dm_dtype=True
-) -> DensityMatrix | npt.NDArray:
+def rand_dm(dim: int, rng: Generator | None = None, rank: int | None = None) -> npt.NDArray:
```
EarlMilktea: Can't we specify the array element type?

graphix/sim/base_backend.py
```diff
@@ -480,6 +476,24 @@ def swap(self, qubits: tuple[int, int]) -> None:
             (control, target) qubit indices
         """
 
+    def apply_noise(self, qubits: list[int], noise: Noise) -> None:  # noqa: ARG002,PLR6301
```
EarlMilktea: Let's use `collections.abc` rather than `list` .

graphix/utils.py
```diff
@@ -82,3 +85,58 @@ def iter_empty(it: Iterator[_T]) -> bool:
     This function consumes the iterator.
     """
     return all(False for _ in it)
+
+
+ValueT = TypeVar("ValueT")
```
EarlMilktea: Let's make it private.

graphix/utils.py
```diff
@@ -82,3 +85,58 @@ def iter_empty(it: Iterator[_T]) -> bool:
...
+    https://docs.python.org/3/howto/descriptor.html#custom-validators
+    """
+
+    def __set_name__(self, owner: object, name: str) -> None:
+        """Set private field name."""
+        self.private_name = "_" + name
+
+    def __get__(self, obj: object, objtype: object = None) -> ValueT:
+        """Get the validated value from the private field."""
+        return cast("ValueT", getattr(obj, self.private_name))
```
EarlMilktea: Please don't cast.

graphix/utils.py
```diff
@@ -82,3 +85,58 @@ def iter_empty(it: Iterator[_T]) -> bool:
     This function consumes the iterator.
     """
     return all(False for _ in it)
+
+
+ValueT = TypeVar("ValueT")
+
+
+class Validator(ABC, Generic[ValueT]):
```
EarlMilktea: I don't think this ABC is necessary, as

- descriptor is rarely used in a dynamic way (ex. passed to generic functions)
- currently there is only one sibling class

I prefer simpler code base to generic abstraction too complex for `mypy` .

graphix/utils.py
```diff
@@ -82,3 +85,58 @@ def iter_empty(it: Iterator[_T]) -> bool:
...
+    https://docs.python.org/3/howto/descriptor.html#custom-validators
+    """
+
+    minvalue: float | None = None
+    maxvalue: float | None = None
+
+    @typing_extensions.override
+    def validate(self, value: object) -> None:
+        """Validate the assigned value."""
+        if not isinstance(value, (int, float)):
```
EarlMilktea: This check would be redundant as it's clearly described in annotations.

graphix/utils.py
```diff
@@ -82,3 +85,58 @@ def iter_empty(it: Iterator[_T]) -> bool:
...
+        self.validate(value)
+        setattr(obj, self.private_name, value)
+
+    @abstractmethod
+    def validate(self, value: ValueT) -> None:
+        """Validate the assigned value."""
+
+
+@dataclass
+class Number(Validator[float]):
```
EarlMilktea: Name `Number` could be too general (actually it's conflicting with `number.Number` )

graphix/utils.py
```diff
@@ -82,3 +85,58 @@ def iter_empty(it: Iterator[_T]) -> bool:
...
+    """Descriptor to validate value.
+
+    https://docs.python.org/3/howto/descriptor.html#custom-validators
+    """
+
+    def __set_name__(self, owner: object, name: str) -> None:
+        """Set private field name."""
+        self.private_name = "_" + name
+
+    def __get__(self, obj: object, objtype: object = None) -> ValueT:
```
EarlMilktea: This `__get__` does not correctly handle attribute accesses via class rather than instance.

EarlMilktea (changes requested):
WIP: Tests are not checked for now!

thierry-martinez:
> WIP: Tests are not checked for now!

Thank you very much for your review! I just added some new tests in bf9f11ad.

graphix/noise_models/depolarising_noise_model.py
thierry-martinez: I renamed it to `depolarising.py` in d9188da7. I moved the `NoiselessNoiseModel` class to `noise_model.py` to avoid the longer module name `noiseless_noise_model.py`, and I thought that `NoiselessNoiseModel` was too trivial to deserve its own module!

graphix/noise_models/depolarising_noise_model.py
thierry-martinez: Fixed in 9898f34d.

graphix/utils.py
thierry-martinez: Fixed in 9898f34d.

graphix/noise_models/depolarising_noise_model.py
thierry-martinez: Fixed in dad7b98b. The fact that `rng` was stored in the noise model was inherited from the previous noise model API, but I extended the new API to pass the `rng` from the simulator as a method argument to avoid storing it in the noise model classes.

graphix/random_objects.py
thierry-martinez: Fixed in c60bc1ac for all `random_objects` functions. Note, the lack of annotation was inherited and not introduced in this PR.

graphix/sim/base_backend.py
thierry-martinez: Fixed in c0c657d8.

graphix/utils.py
thierry-martinez: Fixed in 32659174.

graphix/utils.py
thierry-martinez: Fixed in 9045143a.

graphix/utils.py
thierry-martinez: There is an abstract method `validate`, so `ABC` is needed to prevent `Validator` from being instantiated directly. `mypy` handles abstract base classes correctly.

graphix/utils.py
thierry-martinez: Fixed in c5d96be9.

graphix/utils.py
thierry-martinez: Renamed into `BoundedFloat` in 395a09c5.

graphix/sim/density_matrix.py
```diff
@@ -386,6 +386,20 @@ def apply_channel(self, channel: KrausChannel, qargs: Sequence[int]) -> None:
...
+        Parameters
+        ----------
+        qubits : sequence of ints.
+            Target qubits
+        noise : Noise
+            Noise to apply
+        """
+        channel = noise.to_kraus_channel()
+        self.apply_channel(channel, qubits)
+
```
shinich1, line 402: Do we need this? Minor comment but I feel like this conversion should be handled in backend and not the simulator itself.

shinich1 (commented):
Sorry for the delay. I have two major comments:

1. Why separaete `Noise` and `KrausChannel`? It feels one-to-one
2. the name `A` for the new command is a bit confusing. is it A from 'apply noise'? How about two letters - like `ER` for error, or `CH` for channel? or, `K` for Kraus?

graphix/sim/density_matrix.py
thierry-martinez: Sorry, I missed this comment: the method you cite is indeed defined in the density matrix backend.

graphix/utils.py
```diff
@@ -100,22 +102,33 @@ def __set_name__(self, owner: object, name: str) -> None:
         """Set private field name."""
         self.private_name = "_" + name
 
-    def __get__(self, obj: object, objtype: object = None) -> ValueT:
+    @overload
+    def __get__(self, obj: None, objtype: type | None = None) -> Validator[_ValueT]:  # access on class
```
EarlMilktea: Let's use `typing_extensions.Self` .

graphix/utils.py
```diff
@@ -100,22 +102,33 @@ def __set_name__(self, owner: object, name: str) -> None:
...
 
-    def __set__(self, obj: object, value: ValueT) -> None:
+    def __set__(self, obj: object, value: _ValueT) -> None:
         """Validate and set the value in the private field."""
         self.validate(value)
         setattr(obj, self.private_name, value)
 
     @abstractmethod
-    def validate(self, value: ValueT) -> None:
+    def validate(self, value: _ValueT) -> None:
```
EarlMilktea, line 127: One suggestion: how about changing its its signature to `(self, value: object) -> TypeGuard[_ValueT]` and perform type validations inside?
This might make `Any` -> `_ValueT` casting a bit safer.

```py
# Any type narrowed without validation!
result: _ValueT = getattr(obj, self.private_name)
```

graphix/utils.py
thierry-martinez: Fixed in 5d190d4.

graphix/utils.py
thierry-martinez: Adding a `TypeGuard` in the signature of validate won’t make the `getattr` call in `__get__` any safer, since validate is only invoked in the setter, not in the getter. Calling validate in the getter would defeat the purpose of the validator, which is to enforce the value invariant once, without rechecking it on every access.

The unsafety of using getattr/setattr here is inherent to the descriptor/validator protocol. I could revert to the pre-descriptor version (commit a2675b94, implementing your suggestion: https://github.com/TeamGraphix/graphix/pull/271#discussion_r2054210711), but I still think this kind of validation is valuable to keep. The `utils.py` module is typically where we encapsulate unsafe typing hacks into safe, reusable interfaces.

thierry-martinez: Another point: adding a `TypeGuard` does not actually make the code safer, since type checkers do not verify that the declared type guard matches the actual checks performed in the function body. TypeGuards are useful for strengthening the contract expressed by a function’s signature, bypassing the type checker’s inference. In this code, however, the validator invariant is already enforced by the contract exposed through the getter and setter.

EarlMilktea: OK I see. Let's keep going!

graphix/noise_models/__init__.py
```diff
@@ -2,7 +2,24 @@
...
-__all__ = ["NoiseModel", "NoiselessNoiseModel"]
+__all__ = [
+    "ApplyNoise",
+    "CommandOrNoise",
+    "ComposeNoiseModel",
+    "DepolarisingNoise",
+    "DepolarisingNoiseModel",
+    "Noise",
+    "NoiseModel",
+    "NoiselessNoiseModel",
```
EarlMilktea: Why not reduce `__all__` entries?
Especially I believe `NoiselessNoiseModel` and `CommandOrNoise` do not contribute to better UX as they're rarely necessary in user codes.

thierry-martinez: I removed `NoiselessNoiseModel` from `__all__` in 324d597. However, `CommandOrNoise` belongs to the API that users must implement to define a new `NoiseModel` subclass, so it makes sense to keep it.

PR #273, opened 2025-04-28
Fix #252: pyzx 0.9.0, move to extra, comment graphix-{ibmq,perceval}

### LINKED ISSUE #252: [Bug]: Compatibility with pyzx 0.9
Changes introduced in pyzx 0.9 have broken compatibility with how pyzx is used in `opengraph.py` (see #250). We need to investigate how to restore compatibility with the latest version of pyzx.

thierry-martinez:
This commit moves `pyzx` from `requirements-dev.txt` to `requirements-extra.txt`, because this is not a development requirement but an optional dependency.

Tests succeed with pyzx 0.9.0. I am no longer able to reproduce the errors mentioned in #252 and in
https://github.com/TeamGraphix/graphix/pull/250#issuecomment-2676562457

graphix-ibmq and graphix-perceval are broken (in particular, they are not installable with Python 3.12). They should be fixed in #261; I comment them for now.

Dependabot can now track pyzx again.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,5 +8,3 @@ updates:
       python-packages:
         patterns:
           - "*"
-    ignore:
-      - dependency-name: "pyzx"
--- a/.github/workflows/typecheck.yml
+++ b/.github/workflows/typecheck.yml
@@ -24,7 +24,7 @@ jobs:
 
       - run: |
           python -m pip install --upgrade pip
-          pip install -e .[dev]
+          pip install -e .[dev,extra]
 
       - run: mypy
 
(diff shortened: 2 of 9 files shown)


wlcsm:
LGTM. Tests all pass

wlcsm (approved):

shinich1 (approved):

thierry-martinez:
In this PR:
- We fixed the angle conventions: gate RZ(α) is transpiled into a green vertex with phase α is PyZX, which is transpiled into a node with measure -α in OpenGraph, which is transpiled into M(α) in the pattern.
- We fixed the order of output nodes in patterns returned by `generate_from_graph`.
Simulations between circuits transpiled via PyZX and those transpiled via our transpiler now match.
@shinich1 , since they are quite substantial changes, are you still agree to merge?

shinich1:
@thierry-martinez thanks, makes sense to me. Could you clearly state the changes on changelog?

PR #277, opened 2025-05-13
Pretty-print circuits and patterns

thierry-martinez:
- Add dedicated pretty-printing functions for patterns.
- Improve `str()` and `repr()` for circuits, patterns, and core data types (`Sign`, `Axis`, `Plane`, `Command`, `Instruction`).

This is a focused extract from PR #245, removing any external dependencies:
- Render patterns as ASCII, Unicode, or LaTeX snippets (PDF/image output removed).
- Provide evaluable representations for patterns, circuits, and other core types.

Per Shinichi’s suggestion (https://github.com/TeamGraphix/graphix/pull/245#issuecomment-2794591891), move most code from `pattern.py` into a new module `pretty_print.py`.

Per EarlMilktea’s suggestion (https://github.com/TeamGraphix/graphix/pull/245#issuecomment-2795217009), add type checking to `pretty_print.py` and its tests (`test_pretty_print.py`).

### DIFF
--- a/graphix/command.py
+++ b/graphix/command.py
@@ -10,6 +10,7 @@ from typing import ClassVar, Literal, Union
 
 import numpy as np
 
+import graphix.pretty_print
 from graphix import utils
 from graphix.clifford import Clifford
 from graphix.fundamentals import Plane, Sign
@@ -52,6 +53,10 @@ class N(_KindChecker):
     state: State = dataclasses.field(default_factory=lambda: BasicStates.PLUS)
     kind: ClassVar[Literal[CommandKind.N]] = dataclasses.field(default=CommandKind.N, init=False)
 
+    def __repr__(self) -> str:
+        """Return the representation of a N command."""
+        return graphix.pretty_print.pretty_repr_dataclass(self)
+
 
 @dataclasses.dataclass
 class M(_KindChecker):
@@ -79,6 +84,10 @@ class M(_KindChecker):
             domains.t_domain,
         )
 
+    def __repr__(self) -> str:
+        """Return the representation of an M command."""
+        return graphix.pretty_print.pretty_repr_dataclass(self, unit_of_pi=True)
+
 
 @dataclasses.dataclass
 class E(_KindChecker):
@@ -86,6 +95,10 @@ class E(_KindChecker):
 
     nodes: tuple[Node, Node]
     kind: ClassVar[Literal[CommandKind.E]] = dataclasses.field(default=CommandKind.E, init=False)
+
+    def __repr__(self) -> str:
+        """Return the representation of an E command."""
+        return graphix.pretty_print.pretty_repr_dataclass(self)
 
 
 @dataclasses.dataclass
@@ -96,6 +109,10 @@ class C(_KindChecker):
     clifford: Clifford
     kind: ClassVar[Literal[CommandKind.C]] = dataclasses.field(default=CommandKind.C, init=False)
 
+    def __repr__(self) -> str:
+        """Return the representation of a C command."""
+        return graphix.pretty_print.pretty_repr_dataclass(self)
+
 
 @dataclasses.dataclass
 class X(_KindChecker):
@@ -104,6 +121,10 @@ class X(_KindChecker):
     node: Node
     domain: set[Node] = dataclasses.field(default_factory=set)
     kind: ClassVar[Literal[CommandKind.X]] = dataclasses.field(default=CommandKind.X, init=False)
+
+    def __repr__(self) -> str:
+        """Return the representation of an X command."""
+        return graphix.pretty_print.pretty_repr_dataclass(self)
 
 
 @dataclasses.dataclass
@@ -114,6 +135,10 @@ class Z(_KindChecker):
     domain: set[Node] = dataclasses.field(default_factory=set)
     kind: ClassVar[Literal[CommandKind.Z]] = dataclasses.field(default=CommandKind.Z, init=False)
 
+    def __repr__(self) -> str:
+        """Return the representation of a Z command."""
+        return graphix.pretty_print.pretty_repr_dataclass(self)
+
 
 @dataclasses.dataclass
 class S(_KindChecker):
@@ -122,6 +147,10 @@ class S(_KindChecker):
     node: Node
     domain: set[Node] = dataclasses.field(default_factory=set)
     kind: ClassVar[Literal[CommandKind.S]] = dataclasses.field(default=CommandKind.S, init=False)
+
+    def __repr__(self) -> str:
+        """Return the representation of an S command."""
+        return graphix.pretty_print.pretty_repr_dataclass(self)
 
 
 @dataclasses.dataclass
--- a/graphix/command.py
+++ b/graphix/command.py
@@ -45,21 +45,17 @@ class _KindChecker:
         utils.check_kind(cls, {"CommandKind": CommandKind, "Clifford": Clifford})
 
 
-@dataclasses.dataclass
-class N(_KindChecker):
+@dataclasses.dataclass(repr=False)
+class N(_KindChecker, graphix.pretty_print.DataclassMixin):
     """Preparation command."""
 
     node: Node
     state: State = dataclasses.field(default_factory=lambda: BasicStates.PLUS)
     kind: ClassVar[Literal[CommandKind.N]] = dataclasses.field(default=CommandKind.N, init=False)
 
-    def __repr__(self) -> str:
-        """Return the representation of a N command."""
-        return graphix.pretty_print.pretty_repr_dataclass(self)
 
-
-@dataclasses.dataclass
-class M(_KindChecker):
+@dataclasses.dataclass(repr=False)
+class M(_KindChecker, graphix.pretty_print.DataclassMixin):
     """Measurement command. By default the plane is set to 'XY', the angle to 0, empty domains and identity vop."""
 
     node: Node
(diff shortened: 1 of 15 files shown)


EarlMilktea (changes requested):

graphix/pretty_print.py
```diff
@@ -0,0 +1,263 @@
...
+            arguments = [str(cmd.node), *arguments]
+            if command_domain:
+                arguments.append(domain_to_str(command_domain))
+                command_domain = None
+            out.append(f"({','.join(arguments)})")
+        if cmd.kind == command.CommandKind.M and (cmd.s_domain or cmd.t_domain):
+            out = ["[", *out, "]"]
+            if cmd.t_domain:
+                if output == OutputFormat.LaTeX:
+                    t_domain_str = f"{{}}_{{{','.join([str(node) for node in cmd.t_domain])}}}"
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in 4a4bdd7. Thanks.

graphix/pattern.py
```diff
@@ -80,11 +81,16 @@ class Pattern:
         total number of nodes in the resource state
     """
 
-    def __init__(self, input_nodes: list[int] | None = None) -> None:
+    def __init__(self, input_nodes: list[int] | None = None, cmds: list[Command] | None = None) -> None:
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in commit 74f1d82, and I also corrected the type annotation in the `extend` method.

graphix/pretty_print.py
```diff
@@ -0,0 +1,263 @@
...
+
+    num, den = frac.numerator, frac.denominator
+    sign = "-" if num < 0 else ""
+    num = abs(num)
+
+    if output == OutputFormat.LaTeX:
+        pi = r"\pi"
+
+        def mkfrac(num: str, den: str) -> str:
+            return f"\\frac{{{num}}}{{{den}}}"
```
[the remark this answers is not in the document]
thierry-martinez: There is no raw f-string. I think the proposed code is more readable than:
```python
    return r"\frac{" + str + "}{" + den + "}"
```

graphix/transpiler.py
```diff
@@ -67,18 +69,67 @@ class Circuit:
         List containing the gate sequence applied.
     """
 
-    def __init__(self, width: int):
+    def __init__(self, width: int, instr: list[Instruction] | None = None):
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in 3246c62. Thanks.

graphix/pretty_print.py
```diff
@@ -0,0 +1,263 @@
...
+    from _typeshed import DataclassInstance
+
+    from graphix.command import Node
+    from graphix.pattern import Pattern
+
+
+class OutputFormat(Enum):
+    """Enumeration of the output format for pretty-printing."""
+
+    ASCII = "ASCII"
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in 0911cba. Thanks.

graphix/command.py
```diff
@@ -52,6 +53,10 @@ class N(_KindChecker):
     state: State = dataclasses.field(default_factory=lambda: BasicStates.PLUS)
     kind: ClassVar[Literal[CommandKind.N]] = dataclasses.field(default=CommandKind.N, init=False)
 
+    def __repr__(self) -> str:
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in 65c7952. Thanks.

graphix/instruction.py
```diff
@@ -53,6 +54,10 @@ class CCX(_KindChecker):
     controls: tuple[int, int]
     kind: ClassVar[Literal[InstructionKind.CCX]] = dataclasses.field(default=InstructionKind.CCX, init=False)
 
+    def __repr__(self) -> str:
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in 65c7952. Thanks.

graphix/fundamentals.py
```diff
@@ -40,6 +41,10 @@ def __str__(self) -> str:
             return "+"
         return "-"
 
+    def __repr__(self) -> str:
```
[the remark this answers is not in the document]
thierry-martinez: Fixed in f27ee88. Thanks.

graphix/pretty_print.py
```diff
@@ -0,0 +1,283 @@
...
+            # We use `SupportsFloat` since `isinstance(cmd.angle, float)`
+            # is `False` if `cmd.angle` is an integer.
+            if isinstance(cmd.angle, SupportsFloat):
+                angle = float(cmd.angle)
+                if not math.isclose(angle, 0.0):
+                    arguments.append(angle_to_str(angle, output))
+            else:
+                # If the angle is a symbolic expression, we can only delegate the printing
+                # TODO: We should have a mean to specify the format
+                arguments.append(str(cmd.angle * math.pi))
```
shinich1, line 139: have we tested for symbolic expressions included in pattern?

graphix/pretty_print.py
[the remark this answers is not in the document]
EarlMilktea: No, there is.

```py
>>> num = 3
>>> den = 5
>>> rf"\frac{{{num}}}{{{den}}}"
'\\frac{3}{5}'
>>> f"\frac{{{num}}}{{{den}}}"
'\x0crac{3}{5}'
```

graphix/pattern.py
```diff
@@ -97,6 +104,9 @@ def __init__(self, input_nodes: list[int] | None = None) -> None:
         # output nodes are initially input nodes, since none are measured yet
         self.__output_nodes = list(input_nodes)
 
+        if cmds:
```
EarlMilktea, line 107: `Iterable` may not implement `__bool__` .

```suggestion
        cmds = list(cmds)
        if cmds:
```

graphix/transpiler.py
```diff
@@ -67,18 +69,67 @@ class Circuit:
...
         width : int
             number of logical qubits for the gate network
+        instr : list[instruction.Instruction] | None
+            Optional. List of initial instructions.
         """
         self.width = width
-        self.instruction: list[instruction.Instruction] = []
+        self.instruction: list[Instruction] = []
         self.active_qubits = set(range(width))
+        if instr:
```
EarlMilktea, line 86: Same as above

graphix/pretty_print.py
```diff
@@ -0,0 +1,283 @@
...
+        Returns
+        -------
+        str
+            A string in the form `ClassName.MEMBER_NAME`.
+        """
+        # Equivalently (as of Python 3.12), `str(value)` also produces
+        # "ClassName.MEMBER_NAME", but we build it explicitly here for
+        # clarity.
+        # Mypy ignores that the instance is a field and hence has a name.
+        name = self.name  # type: ignore[attr-defined]
```
EarlMilktea, line 282: `type: ignore` could be avoided by ambient declaration (`name: str` in the class definition)

graphix/pretty_print.py
```diff
@@ -0,0 +1,283 @@
...
+    Use with `@dataclass(repr=False)` on the target class.
+    """
+
+    def __repr__(self) -> str:
+        """Return a representation string for a dataclass."""
+        cls_name = type(self).__name__
+        arguments = []
+        saw_omitted = False
+        # Mypy will see `self` as a non-dataclass, so ignore arg-type here
+        for field in dataclasses.fields(self):  # type: ignore[arg-type]
```
EarlMilktea, line 243: `type: ignore` could be avoided by explicitly annotating `self` as `_typeshed.DataClassInstance` .

graphix/pretty_print.py
EarlMilktea: Let's add explicit type assertion:

```py
        if not isinstance(self, Enum):
            msg = "EnumMixin can only be used with Enum classes."
            raise TypeError(msg)
```

graphix/pattern.py
thierry-martinez: Indeed, thank you for reporting this. Fixed in 74d5a5d. I preferred to check that `cmds is not None` instead of instantiating a list.

graphix/transpiler.py
thierry-martinez: Fixed in 74d5a5d.

graphix/pretty_print.py
thierry-martinez: Fixed in 7b090a0.

graphix/pretty_print.py
thierry-martinez: Clever! I never thought that `self` could be type-annotated! Fixed in b9990f9.

graphix/pretty_print.py
thierry-martinez: `cmd.angle` has type `ExpressionOrFloat`, so if it isn't `SupportsFloat`, then it's a symbolic expression.

graphix/pretty_print.py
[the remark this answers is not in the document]
thierry-martinez: Wow, nice! Fixed in c617563. Thank you.

graphix/instruction.py
```diff
@@ -45,149 +62,149 @@ def __init_subclass__(cls) -> None:
         utils.check_kind(cls, {"InstructionKind": InstructionKind, "Plane": Plane})
 
 
-@dataclasses.dataclass
```
EarlMilktea, line 48: We may have to fix import conventions to suppress this kind of regressions (`dataclasses.dataclass` is used here because it is function + I'm always using `mod.f` styles for functions).

graphix/pattern.py
```diff
@@ -170,97 +191,90 @@ def n_node(self):
...
         Parameters
         ----------
-        input_nodes: list of int
+        input_nodes: iterable of int
             input nodes order determined by user. each index corresponds to that of logical qubits.
         """
+        input_nodes = list(input_nodes)  # make our own copy (allow iterators to be passed)
         assert_permutation(self.__input_nodes, input_nodes)
-        self.__input_nodes = list(input_nodes)
+        self.__input_nodes = input_nodes
```
EarlMilktea, line 216: It's bad to store abstract types in class without generics because it leads to repeated type narrowing each time we call something not in ABC. Please revert it.

graphix/transpiler.py
```diff
@@ -67,18 +69,67 @@ class Circuit:
         List containing the gate sequence applied.
     """
 
-    def __init__(self, width: int):
+    def __init__(self, width: int, instr: Iterable[Instruction] | None = None):
```
EarlMilktea, line 72:
```suggestion
    def __init__(self, width: int, instr: Iterable[Instruction] | None = None) -> None:
```

graphix/pretty_print.py
```diff
@@ -0,0 +1,286 @@
...
+    short_command_list = command_list[: limit - 1] if truncated else command_list
+    result = separator.join(command_to_str(command, output) for command in short_command_list)
+    if output == OutputFormat.LaTeX:
+        result = f"\\({result}\\)"
+    if truncated:
+        return f"{result}...({len(command_list) - limit + 1} more commands)"
+    return result
+
+
+class DataclassMixin:
```
EarlMilktea, line 228: I noticed that the name `DataClassMixin` leads to confusion when not prefixed with the module `pretty_print`. How about including `Print` or similar words to clarify that it's for pretty printing?

graphix/pretty_print.py
```diff
@@ -0,0 +1,286 @@
...
+            value_str = custom_repr(value) if custom_repr else repr(value)
+            if saw_omitted:
+                arguments.append(f"{field.name}={value_str}")
+            else:
+                arguments.append(value_str)
+        arguments_str = ", ".join(arguments)
+        return f"{cls_name}({arguments_str})"
+
+
+class EnumMixin:
```
EarlMilktea, line 262: Same as above.

graphix/pretty_print.py
```diff
@@ -0,0 +1,286 @@
...
+            out.append(domain_str)
+    return f"{''.join(out)}"
+
+
+def pattern_to_str(
+    pattern: Pattern,
+    output: OutputFormat,
+    left_to_right: bool = False,
+    limit: int = 40,
+    target: Container[command.CommandKind] | None = None,
```
EarlMilktea, line 197: `Container` is inappropriate here because `__bool__` is not provided (maybe `Collection`, please check https://docs.python.org/3.13/library/collections.abc.html ).
Additionally I personally prefer `Iterable` + `set` collection here because `pattern_to_str` is unlikely to be a bottleneck (it makes sense to use derived ABCs in performance-critical codes, because they make codes zero-copy.)

EarlMilktea (changes requested):

graphix/transpiler.py
thierry-martinez: Fixed in 9bdf94b.

graphix/instruction.py
thierry-martinez: I'm not sure I understand where the regression is. `@dataclass` is more concise and idiomatic; most libraries use this convention. It avoids repeating the module name, is recommended by PEP 557, and follows the pattern used by other common decorators like `@staticmethod`, `@classmethod`, and `@property`.

graphix/pattern.py
thierry-martinez: I completely agree with your point, but I think line [214](https://github.com/TeamGraphix/graphix/blob/f572fad856e3e4c63e8c45959228a3deab8a579e/graphix/pattern.py#L214) might have been overlooked:
```python
        input_nodes = list(input_nodes)  # make our own copy (allow iterators to be passed)
```

graphix/pretty_print.py
thierry-martinez: Renamed into `DataclassPrettyPrintMixin` in d111615.

graphix/pretty_print.py
thierry-martinez: Renamed into `EnumPrettyPrintMixin` in https://github.com/TeamGraphix/graphix/commit/d11161572d8e3e6647fc7f05bdfb87ad81dd94a6.

graphix/pretty_print.py
thierry-martinez: I changed the condition to `target is not None` in 256af8c.
Regarding the use of `Container`, I’m not sure I understand why `Iterable` would be preferable in this case: both are ABCs, and here we only rely on `__contains__`, not iteration. Since we don’t need to iterate over the object, `Container` seems more semantically appropriate than `Iterable`. Could you clarify your reasoning?

graphix/instruction.py
EarlMilktea: Checked the PEP. I understood your intention, but still, we would need to use simple (minimal exceptions) and coherent styles for preventing the situation "someone added `dataclasses.` prefix for his own convention, then another removed for his own". Let's discuss later.

graphix/pattern.py
EarlMilktea: Oops, sorry!

graphix/pretty_print.py
EarlMilktea: My concern here is that, while `Iterable` is in general efficient, `Container` is sometimes not (ex. `list.__contains__` is $O(N)$ ) so I sometimes use `Iterable` as annotation and collect into `set` for fast member checking.

thierry-martinez: I think `Container` is preferable here in all cases:
- `pattern_to_str` is unlikely to be performance-critical, so I would prioritize semantic correctness.
- If the caller truly cares about performance, they can still pass a structure with an efficient `__contains__` implementation. In contrast, requiring an `Iterable` would force a copy, which could be less efficient.

EarlMilktea (approved):

Issue #157, opened 2024-05-27
`minimize_space` breaks runnability for patterns that are inconsistent with their graph flow


thierry-martinez:
**Describe the bug**
If a pattern has a graph with a flow but is not consistent with this flow, `minimize_space` leads to a pattern with is not runnable anymore.

**To Reproduce**
```
import graphix
pattern = graphix.pattern.Pattern(input_nodes=[0])
pattern.add(["N", 1])
pattern.add(["N", 2])
pattern.add(["E", (0, 1)])
pattern.add(["E", (1, 2)])
pattern.add(["M", 1, "XY", 0, [], []])
pattern.add(["M", 0, "XY", 0, [1], []])
pattern.add(["Z", 2, [0]])
pattern.simulate_pattern() # OK
pattern.minimize_space()
pattern.simulate_pattern() # KeyError: 1
```

**Expected behavior**
`minimize_space` should preserve pattern runnability.

mgarnier59:
It is not an emergency but it is a common issue in Graphix. We need to know if the pattern is consistent with a gflow before applying some operations. If it turns out not to be the case then problems arise. 

I think @shinich1 had an example.
I planned on diving into what exactly happens in the graphical version of Elliott et al. but since you know that better than I do, might be a good idea. Another topic we can adress in July!

mgarnier59:
What's the status on this? Recent refactors should solve the problem? Let's discuss it.

Issue #120, opened 2024-02-09
Add optional argument to pass gflow in `generate_from_graph()`


pafloxy:
**Is your feature request related to a problem? Please describe.**
NA

**Describe the feature you'd like**
Currently the facility to generate a Pattern from a graph is provided by the function `generator.generate_from_graph()` which takes as arguments the following `generate_from_graph(graph, angles, inputs, outputs, meas_planes=None)` . I think it would be good if we can have an option to pass a `gflow`  of our choice rather than invoking the `find_gflow()` internally. 

Something of the form `generate_from_graph(graph, angles, inputs, outputs, meas_planes=None, set_gflow= None)` would be convenient. Incase `None` is passed it should proceed as usual. 

**Additional context**
NA

shinich1:
@pafloxy Nice, I would suggest to create a separate function 'generate_from_gflow' instead.

pafloxy:
A seperarate function as in ? Like are the things that you want to be different in this case ?

As for me, a change like this 
![image](https://github.com/TeamGraphix/graphix/assets/61856561/a29b6930-4acf-4722-80cc-868bf2d6960f)
would be enough.

shinich1:
I get the change may be small, but as a user interface, `generate_from_graph` should generate from graph, not from (g)flow. Adding special case that you cannot guess from the function name is not useful to everyone.

we can simply call `generate_from_gflow` inside `generate_from_graph`, if gflow is found.

masa10-f:
> generate_from_graph should generate from graph, not from (g)flow.

It seems a bit strange to me. The `generate_from_graph` actually constructs pattern using both (g)flow and a graph with planes and angles, not just from graph alone. In fact `generate_from_graph` uses `find_gflow` internally. I understand the method automatically finds (g)flow just for user convenience. Isn't it more logical to include the optional argument in the method rather than creating a new one, because the current version just limits the freedom of byproduct correction?

masa10-f:
Although it might be out of context but it could be possible to create a more general method, which construct a pattern from graph, 'x_flow', and 'z_flow', not limited to (g)flow. Here 'x_flow' and 'z_flow' mean byproduct correction for X and Z. For a deterministic pattern, they are equivalent to f/g and Odd(f/g). This approach would offer a more convenient way to create non-deterministic patterns than using a manual pattern addition method.

shinich1:
[...]

PR #280, opened 2025-05-14
Add `fastflow`

### LINKED ISSUE #276: Refactor graph-related modules + introduce `fastflow`
### my suggestions

- completely remove linalg stuff
  - drop `sympy` / `galois` from `requirements.txt`
- refactor `OpenGraph` to be able to have `PauliMeasurement` inside
- implement mutual conversion between `Pattern` and `OpenGraph`
- simplify/isolate modules
  - remove `generator` and make them methods of graph or flow
  - remove `extraction` and make them methods of graph
  - remove `Pattern` API from visualization: draw from graph directly
  - remove `from_pattern` / `from_graph` stuff if appropriate
  - define a dedicated module for complex pattern manipulations
    - complex = go back and forth between pattern and graph

EarlMilktea:
Closes #276 .

### MEMO

- currently using `git+` in requirements.


EarlMilktea:
@shinich1 @thierry-martinez @mgarnier59 (CC: @WingCode )

Sharing my new API design plan. Could you take a look?

### Changes

- Users need to explicitly cast `Plane` to `PauliPlane` before calling `xxx_pauliflow` .
  - Robust against `float` precision/angle convention problems.
- Flow function is now of type `dict[int, int]` .
- Redundant functions are all pruned.
- Redundant arguments are all dropped.

EarlMilktea:
> why we're not extending `Plane` class?

Because `Enum` does not support inheritance, if I remember correctly.

mgarnier59:
Hey! We haven't had a chance to have a look yet so I don't really know what this `PauliPlane` is but from the top of my head, we need decide make everything consistent with what is done in `pauli.py` where there is a `Pauli`class with `axis`and `sign`attributes.

EarlMilktea:
> make everything consistent with what is done in `pauli.py` where there is a `Pauli`class with `axis`and `sign`attributes.

> * Refactor `M` command
>   * Hold `Measurement` or `PauliMeasurement` inside for coherence
> * Refactor `OpenGraph` to accept `PauliMeasurement`

This could be covered later with my suggestions. Anyway we need to "hurry up" as this PR is blocking #294 .

thierry-martinez:
If possible, I would suggest splitting implementation changes and API changes into separate PRs. This is generally a good principle, and it becomes especially helpful if we want to move quickly with the implementation changes.

More specifically for API changes:

| Users need to explicitly cast `Plane` to `PauliPlane` before calling `xxx_pauliflow`.

I think the convention change of introducing a dedicated `PauliPlane` class to distinguish Pauli measurements could be postponed to a future PR. In the context of this PR, we can handle the identification of Pauli measurements at the interface level with `fastflow`. In a later PR, we could better assess the trade-offs.

The precision issue is mostly theoretical: although Python doesn’t formally guarantee IEEE-754 compliance, it’s unclear in what realistic scenarios `graphix` would run on systems where float handling differs. More practically, numerical identification of Pauli measurements can be very useful. For instance, a parameterized measure α / 2 should be recognized as Pauli if and only if α is substituted by an integer.

Moreover, since angle conventions aren’t yet unified between circuits and patterns, any changes related to specialized representations of Pauli angles should also consider the circuit context.

| Flow function is now of type `dict[int, int]`.

Do you mean that only the layer information is returned, and not the full flow? While we mostly use the layer data right now, having access to the full flow could be useful (for example, for circuit extraction).

It might also be worthwhile to define dedicated classes for different types of flows. That way, each flow object could carry metadata about its type and provide methods like `get_corrections()`, making `generate_from_graph` cleaner and more modular.

| Redundant functions are all pruned.
| Redundant arguments are all dropped.

That’s always a welcome improvement!

Finally, it would be better if the API used `OpenGraph` instead of requiring a `networkx` graph along with separate input and output node arguments.

EarlMilktea:
> dedicated classes for different types of flows

This is just the thing I wanted to do in the future!

> precision issue
> parameterized measure

This could be later discussed along with type divergence issues. Anyway, currently I feel `PauliPlane` is a good staring point because

- it's core logic is all delegated to `PauliMeasurement.try_from`
- this approach is robust against numerical errors, regardless of `float` behaviors/angle conventions.

> used `OpenGraph` instead of requiring a `networkx`

Once I tried it but current codebase is not compatible with `PauliMeasurement` .

EarlMilktea:
@mgarnier59 

I misunderstood your intention. Please check 2432b3d0c955dd0a536b44ffd029f694c34b0b2c .

graphix/gflow_shim.py
```diff
@@ -0,0 +1,202 @@
...
+from graphix.fundamentals import Axis, Plane
+from graphix.measurements import PauliMeasurement
+
+if typing_extensions.TYPE_CHECKING:
+    from graphix.pattern import Pattern
+
+
+def get_pauli_nodes(
+    meas_planes: dict[int, Plane], meas_angles: dict[int, float]
+) -> tuple[set[int], set[int], set[int]]:
```
thierry-martinez, line 21: This function does not appear to be used anywhere. It is unclear whether we intend to keep it, and if we do keep it for now, perhaps we should deprecate it.

graphix/gflow_shim.py
```diff
@@ -0,0 +1,202 @@
...
+            l_y |= {node}
+        elif pm.axis == Axis.Z:
+            l_z |= {node}
+        else:
+            typing_extensions.assert_never(pm.axis)
+    return l_x, l_y, l_z
+
+
+def flow_from_pattern(pattern: Pattern) -> tuple[dict[int, set[int]], dict[int, int]]:
+    """Check if the pattern has a valid flow. If so, return the flow and layers.
```
thierry-martinez, line 57: Why is this function moved from `gflow` to `gflow_shim`? For me, a module `gflow_shim` would be a good place if we need a compatibility layer between the new `gflow` module and the former API. But here, `flow_from_pattern` and `gflow_from_pattern` implement flow extraction from patterns, which is independent from the flow-finding algorithms implemented in `gflow`. I don't see why we would want to separate them in another module. More generally, perhaps `gflow` is not a good module name in the first place and should be renamed just as `flow`, to group all the flow-related stuff, but this renaming can be considered out of the scope of this PR.

graphix/gflow.py
```diff
@@ -12,228 +12,123 @@
...
     -------
-    g: dict
-        gflow function. g[i] is the set of qubits to be corrected for the measurement of qubit i.
-    l_k: dict
-        layers obtained by gflow algorithm. l_k[d] is a node set of depth d.
+    d: int
+        minimum depth of graph
+    layers: dict of set
+        components of each layer
     """
```
thierry-martinez, line 118: Better documentation of the input and output mappings would be helpful to understand what this function computes.

```python

    """Group nodes by their layers.

    This function converts a mapping from each node to its layer (as
    obtained from flow or gflow algorithms) into a mapping from each
    layer to the set of nodes it contains.

    Parameters
    ----------
    l_k: dict
        Mapping from nodes to layers, as produced by flow or gflow algorithms.

    Returns
    -------
    d: int
        Minimum depth of graph.
    layers: dict of set
        Mapping from each layer to the set of nodes it contains.
    """
```

graphix/gflow.py
```diff
@@ -12,228 +12,123 @@
...
-):
-    """Find one layer of the gflow.
 
-    Ref: Backens et al., Quantum 5, 421 (2021).
+def _default_construct(keys: Iterable[int]) -> dict[int, Measurement]:
+    # Random angle for safety
+    return dict.fromkeys(
+        keys,
+        Measurement(Plane.XY, 0.5014943209046647),
+    )
```
thierry-martinez, line 93: Having a dummy value for measurement angles looks a bit awkward and is a sign that `Measurement` is not the right data structure to use in this module. I hope that we can eliminate it with the refactoring I propose in this review: use `Measurement` in `OpenGraph` and `generate_from_graph`, but use `Plane` and `PPlane` in `gflow`.

graphix/gflow.py
```diff
@@ -12,228 +12,123 @@
...
 def find_flow(
-    graph: nx.Graph,
-    iset: set[int],
-    oset: set[int],
-    meas_planes: dict[int, Plane] | None = None,
-) -> tuple[dict[int, set[int]], dict[int, int]]:
+    graph: nx.Graph[int],
+    iset: AbstractSet[int],
+    oset: AbstractSet[int],
+    meas: Mapping[int, Measurement] | None = None,
```
thierry-martinez, line 130: The former type for `meas_planes` (`dict[int, Plane]`) expresses precisely what is required to compute a flow. I think we should keep it, in its abstract form `Mapping[int, Plane]`, rather than requiring unnecessary information. The function `generate_from_graph` will be responsible for converting `Measurement` to `Plane` before calling `find_flow`. The name `meas_planes` seemed appropriate but if we want something shorter, we could use just `planes` (although `meas` on its own would be misleading).

graphix/gflow.py
```diff
@@ -245,123 +140,71 @@ def find_flow(
...
-    meas_planes: dict(int, Plane)
-        measurement planes for each qubits. meas_planes[i] is the measurement plane for qubit i.
-        Note that an underlying graph has a causal flow only if all measurement planes are Plane.XY.
-        If not specified, all measurement planes are interpreted as Plane.XY.
+    graph : `networkx.Graph`
+        The underlying graph.
+    iset : `collections.abc.Set`
+        Input nodes.
+    oset : `collections.abc.Set`
+        Output nodes.
```
thierry-martinez, line 149: The documentation of the parameter `meas` is missing. Is that what you meant on line 154?

graphix/gflow.py
```diff
@@ -245,123 +140,71 @@ def find_flow(
...
     """
-    check_meas_planes(meas_planes)
-    nodes = set(graph.nodes)
-    edges = set(graph.edges)
-
-    if meas_planes is None:
-        meas_planes = dict.fromkeys(nodes - oset, Plane.XY)
+    # meas is left undocumented because it is essentially redundant.
+    if meas is None:
+        meas = _default_construct(graph.nodes - oset)
```
thierry-martinez, line 156: The former implementation, where `meas_planes` defaulted to `Plane.XY`, worked fine, and was more principled.

graphix/gflow.py
```diff
@@ -245,123 +140,71 @@ def find_flow(
...
-    l_k = dict.fromkeys(nodes, 0)
-    f = {}
-    k = 1
-    v_c = oset - iset
-    return flowaux(nodes, edges, iset, oset, v_c, f, l_k, k)
+def find_gflow(
+    graph: nx.Graph[int],
+    iset: AbstractSet[int],
+    oset: AbstractSet[int],
+    meas: Mapping[int, Measurement] | None = None,
```
thierry-martinez, line 166: As for `find_flow`, I would prefer to retain the earlier version `Mapping[int, Plane]`, and perform the conversion in `generate_from_graph`.

graphix/gflow.py
```diff
@@ -245,123 +140,71 @@ def find_flow(
...
-    iset: set[int],
-    oset: set[int],
-    meas_planes: dict[int, Plane],
-    meas_angles: dict[int, float],
-    mode: str = "single",
-) -> tuple[dict[int, set[int]], dict[int, int]]:
+    graph: nx.Graph[int],
+    iset: AbstractSet[int],
+    oset: AbstractSet[int],
+    meas: Mapping[int, AnyMeasurement] | None = None,
```
thierry-martinez, line 206: The type we want here is something isomorphic to `Mapping[int, Plane | Axis]`, and to perform the conversion from `Measurement` in `generate_from_graph`.

graphix/gflow_shim.py
EarlMilktea: Because I'm planning to someday remove this.
As `Pattern` -> graph conversion is already implemented, there is no good reason for maintaining redundant direct transformations.
I suppose, this kind of redundant transformation APIs are one of the chief reasons for current densely-connected module/class structures.

thierry-martinez: These functions do not perform the same computation.

- The functions `find_flow`, `find_gflow`, and `find_pauliflow` search for a flow in a given graph, but there may be multiple valid flows in the same graph with different correction strategies.
- The functions `flow_from_pattern`, `gflow_from_pattern`, and `pauliflow_from_pattern`, return the unique flow (if it exists) that corresponds to the correction strategy implemented in a given pattern.

Thus, there is no redundancy between these functions. Here are some consequences of this difference:

- The round-trip `OpenGraph.to_pattern∘find_pauliflow∘OpenGraph.from_pattern` does not necessarily yield the original pattern.
- By contrast, starting from a pattern $p$, if $g$ is the open graph of $p$ and $f$ is the flow of $p$, we can reconstruct $p$ with $g$ and $f$ modulo commutation rules. Currently, we cannot provide a fixed flow to `generate_from_graph`, but there is an open issue addressing this (#120).
- The function `visualize_from_pattern` shows the actual correction strategy of a given pattern, extracted with `flow_from_pattern`. By contrast, if we just use `visualize`, the correction strategy that is drawn may differ from the one implemented by the pattern.
- If these functions are removed, we will have no way to check if a given pattern has a flow: using `find_flow` may return a valid flow from the pattern's open graph, but there is currently no way to check if the pattern itself implements the correction strategy of that flow (it would be interesting to have a function that checks whether a given pattern implements a given flow, #306).

thierry-martinez: For instance, the bug #157 occurs because `find_flow` is used in place of `flow_from_pattern`. I have included an example at: https://github.com/TeamGraphix/graphix/issues/157#issuecomment-3006400824.

masa10-f: Hi @thierry-martinez (CC: @mgarnier59 @shinich1 @EarlMilktea )
Let me have two comments.

1. gflow_from_pattern does not always find an existing gflow
As far as I remember correctly, this function is just a trial patch to find the flow consistent with the Pattern corrections in the `visualizer` module.  There’s a case where `gflow` can find one, but `gflow_from_pattern`  fails. In this case, a pattern is actually generated from some gflow, but the appearance is modified by Pauli simplification described in the Measurement Calculus paper. I mean, if the Graphix team plans to use this in many places, I hope you remember this context.

2. Multiple solutions of gflow
The bug reported in #157 seems to come from the ambiguity of input_nodes rather than the flow itself. If there are multiple solutions for flow (including gflow and Pauli flow) under a sufficient number of input and output, they are all equivalent to each other modulo stabilizer operators. Therefore, it’s in principle not the problem to use only `gflow` method. I agree there are some problems when you do something from Pattern to Pattern, though.

thierry-martinez: Hi @masa10-f. In #157, only causal flows are relevant, as if there is only a gflow but no causal flow, we do not know optimal bounds on space. There is no ambiguity regarding input node (which is fixed). I agree that all flows are equivalent with respect to the unitary that is computed, but the point is that they differ in the measurement order they impose by their correction strategies.

Your first point suggests some bugs in `gflow_from_pattern`: it would be interesting if you could fill an issue with an example. There is also a bug in `pauliflow_from_pattern`, since this function calls `find_pauliflow` rather than considering whether the pattern uses a different correction strategy. Anyway, if these functions have bugs, that means that the bugs should be fixed, not that the functions should be removed!

EarlMilktea: I understand this function cannot be removed now.

@thierry-martinez 

> someday remove this

Do you think it is ever possible in the future? I still believe that we should not define conversions for all the pairs (or module structure becomes too dense)

masa10-f: @thierry-martinez Thank you for your reply.

1. Ambiguity of input_nodes

Sorry for the confusion. What I meant is that when the number of input nodes is smaller than the number of output nodes, there is flow ambiguity corresponding to the missing input nodes.
For example, in the case of #157, there are two possible input states, resulting in two valid solutions. The starting points of the two flow paths differ:

- One is: 3 → 1 → 4 → 5 and 0 → 2
- The other is: 3 → 0 → 2 and 1 → 4 → 5

Even if the full set of input qubits is not specified in the pattern, the flow always assumes a sufficient number of input nodes. Therefore, these represent fundamentally different types of patterns.
I meant that the difference lies not only in the measurement order, but also in the input states themselves, which are not necessarily in the |+⟩ state.
I agree with you that `minimize_space` has a critical bug. This is also an immature patch at the early stage of development.

masa10-f: 2. flow from graph

I assumed the following situation, but I found `find_pauliflow` inside the `pauliflow_from_pattern` successfully avoided my concern.

```python
# %%
from graphix.command import N, E, M, X, Z
from graphix.gflow import (
    flow_from_pattern,
    pauliflow_from_pattern,
)

# %%
# pattern construction
# simple chain graph: 0 (in) - 1 - 2 (out)

pattern = Pattern(
    input_nodes=[0],
    cmds=[
        N(1),
        N(2),
        E((0, 1)),
        E((1, 2)),
        M(0, Plane.XY, 1 / 3, set(), set()),
        M(1, Plane.XY, 0, {0}, set()),  # Pauli X measurement
        X(2, {1}),
        Z(2, {0}),
    ],
    output_nodes=[2],
)

# apparently, this pattern has a trivial flow {0: 1, 1: 2}
flow, l_k = flow_from_pattern(pattern)
print("Flow:", flow)

# %%
# This pattern can be simplified with Pauli simplification procedure.
# The equations are described in the top of 15p [Measurement Calculus paper](https://arxiv.org/pdf/0704.1263)

pattern_simp = Pattern(
    input_nodes=[0],
    cmds=[
        N(1),
        N(2),
        E((0, 1)),
        E((1, 2)),
        M(0, Plane.XY, 1 / 3, set(), set()),
        M(
            1, Plane.XY, 0, set(), set()
        ),  # Pauli X measurement. The s_domain can be eliminated
        X(2, {1}),
        Z(2, {0}),
    ],
    output_nodes=[2],
)

# The correction flow acquired is:
# flowlike = {1, 2}. Here, the oddneighbors are Odd(flowlike[0]) = {}, Odd(flowlike[1]) = {} (I eliminated the self-loop for simplicity)
# They are not consistent with the original pattern_simp strategy: Z{0: 2}
# Moreover, the correction strategy in pattern_simp is not consistent with Pauli flow conditions.

# The below will fail if the method directly extract corrections from pattern. However, it succeeds now because it calls `find_pauliflow` so that it properly finds ignored_nodes.
pauli_flow, l_k = pauliflow_from_pattern(pattern_simp)
print("Pauli flow:", pauli_flow)

# This suggests that the correction strategy on the pattern should be interpreted considering the ignored nodes. If you generate a pattern from the direct correction strategy, it will not work.

# %%
```

EarlMilktea:
I'll close this PR and create a new one for easier reviews.

PR #286, opened 2025-05-28
Graph state 

### LINKED ISSUE #281: [Application] Investigate resource-state protocols
On the one hand, graph state are ubiquitous in quantum information science and technology. Indeed, they play a role in quantum computation (resource state engineering), quantum error correction and quantum communications (entanglement distribution) among others. This is due to the fact that they have a clear entanglement structure are suitable to theoretical processing either in their graph form or using their relationship with stabiliser states. On the other hand, their manipulation is natural in Graphix.

For example, here are some recent papers on these topics

  - Freund, Pirker, Vandré and Dür,  Graph state extraction from two-dimensional cluster states (2025) [preprint](https://arxiv.org/abs/2505.08396).
  - Claudet and Perdix, Local equivalence of stabilizer states: a graphical characterisation (2024) [preprint](https://arxiv.org/abs/2409.20183).
  - Claudet, Mhalla and Perdrix, Small k-pairable states (2023) [preprint](https://arxiv.org/abs/2309.09956).

The idea would be to

- [ ] choose one of the above examples (or propose something along the same lines)
- [ ] add an example, to `examples/resource-state.py` for  `graphix`. Please leave the implementation of the algorithm in the `examples/resource-state.py` directory and not in main graphix package. 
- [ ] Confirm the scalability of such a protocol by increasing the number of nodes associated and checking the scaling of the computing time vs problem size.
- [ ] Make sure the docs render correctly. 
- [ ] Suggest what could be improved that could led to upgrades/new features in `graphix`.

AbdullahKazi500:

- Format added code by `ruff`

Then, please fill in below:

**Context (if applicable):**
This PR introduces a new example script that demonstrates graph state extraction from 2D cluster states and performs local equivalence analysis between stabilizer states. The implementation is based on recent advancements in quantum information theory, particularly in the areas of resource state engineering and equivalence classification of stabilizer states.

Added graph_state_analysis.py, which:

- Constructs 2D cluster states of arbitrary dimensions using graphix.

- Extracts target graph states via local measurements simulating the protocol in Freund et al. (2025).

- Computes local equivalence invariants for stabilizer states as proposed by Claudet and Perdrix (2024).

- Analyzes small k-pairable structures for understanding graph-state scalability.

- Includes a run_scalability_analysis() method to benchmark extraction and analysis over increasing cluster sizes.

- Utility method analyze_resource_graph helps inspect ResourceGraph objects returned from graphix.extraction.

- Added timing metrics for extraction, equivalence checking, and fusion generation for performance profiling.

**Related issue:**
Fixes #281

### DIFF
--- a/examples/resource_state.py
+++ b/examples/resource_state.py
@@ -7,10 +7,8 @@ This module demonstrates how to extract and analyze a resource graph from a
 
 from __future__ import annotations
 
-import itertools
 import time
-from dataclasses import dataclass, field
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 import networkx as nx
 


examples/resource-state.py
```diff
@@ -0,0 +1,594 @@
+"""Graph State Extraction and Local Equivalence Analysis.
```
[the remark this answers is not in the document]
EarlMilktea: How about `resource_state.py` ?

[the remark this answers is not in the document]
EarlMilktea: No, please also do not use capital `R`: it's specified in the PEP8 style guide.

EarlMilktea:
@AbdullahKazi500 Merged `master` to this PR as it was heavily outdated.

examples/resource_state.py
```diff
@@ -0,0 +1,272 @@
...
+    elif hasattr(resource_graph, "type"):
+        info["resource_type"] = resource_graph.type
+
+    return info
+
+
+# %%
+class GraphStateExtractor:
+    """Extract target graph states from 2D cluster states and analyze local equivalence."""
+
```
EarlMilktea, line 94: Please add annotations of `extraction_times` and `equivalence_times` here.

examples/resource_state.py
```diff
@@ -0,0 +1,272 @@
+"""Script for creating and analyzing graph states using Graphix.
```
EarlMilktea, line 1: Please resolve `ruff` errors, or this PR will never be merged.

mgarnier59:
Hi @AbdullahKazi500 

Thanks for participating and writing this PR. We have not yet been able to review your PR and we will let you know when we will have done so and provide you with feedback.

Also, please try to keep the number of notifications at a reasonable level :).

examples/resource_state.py
```diff
@@ -0,0 +1,150 @@
+"""
+===============================
```
EarlMilktea: Please use `numpy` style.

examples/resource_state.py
```diff
@@ -0,0 +1,272 @@
...
+    max_degree: int
+    min_degree: int
+    k: int
+    total_k_subsets: int
+    pairable_subsets: int
+    pairable_ratio: float
+
+
+# %%
+def analyze_resource_graph(resource_graph: Any) -> ResourceGraphInfo:
```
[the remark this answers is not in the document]
EarlMilktea: Still `Any` .

examples/resource_state.py
EarlMilktea: Not done.

examples/resource_state.py
```diff
@@ -0,0 +1,150 @@
...
+from dataclasses import dataclass, field
+from typing import Any
+
+import networkx as nx
+
+from graphix import GraphState
+
+
+@dataclass
+class ResourceGraphInfo:
```
EarlMilktea: Could you explain why this stuff is necessary? For me it seems you're just copying data for no reason.

examples/resource_state.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+    Information about a resource graph.
+    """
+
+    type: str | None = None
```
EarlMilktea: Please do not use the name `type` because it's in conflict with `type` builtin.

examples/resource_state.py
```diff
@@ -0,0 +1,150 @@
...
+                if i < rows - 1:
+                    edges.append((current, current + cols))
+
+        gs.add_edges_from(edges)
+        return gs
+
+    def extract_target_graph_state(
+        self,
+        cluster_state: GraphState,
+        target_edges: list[tuple[int, int]],
```
EarlMilktea: Please use appropriate `collections.abc` classes for annotations, rather than `list` .

examples/resource_state.py
```diff
@@ -0,0 +1,150 @@
...
+        target_nodes: list[int] | None = None,
+    ) -> tuple[GraphState, list[int]]:
+        """
+        Extract a target graph state using local measurements.
+        """
+        start = time.perf_counter()
+
+        if target_nodes is None:
+            target_nodes = list(set(itertools.chain.from_iterable(target_edges)))
```
EarlMilktea: Why both `list` and `set` are required here?

examples/resource_state.py
```diff
@@ -0,0 +1,150 @@
...
+        """
+        info = ResourceGraphInfo()
+        graph = nx.Graph(gs.edges)
+
+        info.nodes = graph.number_of_nodes()
+        info.edges = graph.number_of_edges()
+
+        # FIX: Cast to list to guarantee iterability
+        degree_view = list(graph.degree)
+        info.degree_sequence = sorted([int(d) for _, d in degree_view])
```
EarlMilktea:
```suggestion
        info.degree_sequence = sorted(int(d) for _, d in degree_view)
```

examples/resource_state.py
```diff
@@ -0,0 +1,150 @@
...
+        Compute invariants for local graph state equivalence.
+        """
+        info = ResourceGraphInfo()
+        graph = nx.Graph(gs.edges)
+
+        info.nodes = graph.number_of_nodes()
+        info.edges = graph.number_of_edges()
+
+        # FIX: Cast to list to guarantee iterability
+        degree_view = list(graph.degree)
```
EarlMilktea: `DegreeView` is iterable.

examples/resource_state.py
```diff
@@ -0,0 +1,150 @@
...
+        info.degree_sequence = sorted([int(d) for _, d in degree_view])
+
+        info.spectrum = [float(x) for x in sorted(nx.adjacency_spectrum(graph).real)]
+        info.triangles = sum(nx.triangles(graph).values()) // 3
+        info.is_connected = nx.is_connected(graph)
+        info.num_components = nx.number_connected_components(graph)
+        info.max_degree = max(info.degree_sequence) if info.degree_sequence else None
+        info.min_degree = min(info.degree_sequence) if info.degree_sequence else None
+
+        return info
```
EarlMilktea: Could you add some codes here in the top-level, so that we can make sure that your code runs correctly?

examples/resource_state.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def analyze_resource_graph(resource_graph: Any) -> ResourceGraphInfo:
+    """
+    Analyze a resource graph object and extract basic metadata.
+    """
+    info = ResourceGraphInfo(type=type(resource_graph).__name__)
+
+    # Safely access attributes using getattr with default fallback
+    graph = getattr(resource_graph, "graph", None)
```
EarlMilktea: Please do not use `getattr` / `hasattr` here because we know the exact type `graph` should have.

examples/resource_state.py
AbdullahKazi500: done

examples/resource_state.py
AbdullahKazi500: done

AbdullahKazi500: done

examples/resource_state.py
AbdullahKazi500: done

examples/resource_state.py
AbdullahKazi500: ```set(...)``` removes duplicates.

```list(...)``` converts back to a list, because later logic (target_gs.add_nodes_from(target_nodes)) may rely on it being indexable.

File: examples/resource_state.py (hunk 8 above)
Comment (AbdullahKazi500) [reply examples/resource_state.py]: done

File: examples/resource_state.py (hunk 7 above)
Comment (AbdullahKazi500) [reply examples/resource_state.py]: ok

File: examples/resource_state.py (hunk 6 above)
Comment (AbdullahKazi500) [reply examples/resource_state.py]: A comprehensive collection of metrics and metadata about a quantum resource graph can be encapsulated in a structured container using the ResourceGraphInfo dataclass. This design offers a number of significant advantages:

Code is easier to comprehend and maintain when relevant resource graph properties are grouped into a single, clearly defined object. All pertinent graph properties are grouped together neatly, eliminating the need to juggle several loosely connected variables.

The dataclass enhances developers' comprehension of what each attribute represents and aids in the early detection of flaws through static analysis tools by explicitly stating types for each attribute (using Python type hints).

Flexible and scalable analysis of graph states is made possible by the dataclass's easy extension with more resource metrics in the future without interfering with already written code.

Comment (AbdullahKazi500) [reply examples/resource_state.py]: I would also like to know what do you want to do with this section @EarlMilktea ?

Comment (AbdullahKazi500) [reply examples/resource_state.py]: @EarlMilktea I would like to know the context behind not done in the previous comment 
```
# %%
class GraphStateExtractor:
    """Extract target graph states from 2D cluster states and analyze local equivalence."""
```
can you elaborate

File: examples/resource_state.py [hunk 14]
```diff
@@ -0,0 +1,157 @@
...
+
+class GraphStateExtractor:
+    """
+    Utility class for creating and analyzing graph states.
+    """
+
+    def __init__(self) -> None:
+        """Initialize timing logs."""
+        self.extraction_times: list[float] = []
+        self.equivalence_times: list[float] = []
```
EarlMilktea:
```suggestion
    extraction_times: list[float]
    equivalence_times: list[float]

    def __init__(self) -> None:
        """Initialize timing logs."""
        self.extraction_times = []
        self.equivalence_times = []
```

EarlMilktea:
For example you could do like this.

```py
"""Graph State Resource Extraction

This module demonstrates how to extract and analyze a resource graph from a
2D cluster state using the GraphState class from the `graphix` library.
"""

import itertools
import time

from graphix import GraphState

if TYPE_CHECKING:
    from collections.abc import Iterable


def create_2d_cluster_state(rows: int, cols: int) -> GraphState:
    """
    Create a 2D cluster state as a rectangular lattice.
    """
    gs = GraphState()

    edges: list[tuple[int, int]] = []
    for i in range(rows - 1):
        for j in range(cols - 1):
            current = i * cols + j
            edges.append((current, current + 1))  # noqa: FURB113
            edges.append((current, current + cols))

    gs.add_edges_from(edges)
    gs.add_nodes_from(itertools.chain.from_iterable(edges))
    return gs


def extract_graph_state(
    cluster_state: GraphState,
    edges: Iterable[tuple[int, int]],
    nodes: Iterable[int] | None = None,
) -> tuple[GraphState, set[int]]:
    """
    Extract a graph state using local measurements.
    """
    if nodes is None:
        nodes = itertools.chain.from_iterable(edges)

    gs = GraphState()
    gs.add_nodes_from(nodes)
    gs.add_edges_from(edges)

    return gs, cluster_state.nodes.keys() - nodes


if __name__ == "__main__":
    cluster = create_2d_cluster_state(rows=3, cols=3)

    edges = [(0, 1), (1, 2)]

    t0 = time.perf_counter_ns()

    gs, measured = extract_graph_state(cluster, edges)

    print(f"Elapsed time: {time.perf_counter_ns() - t0} ns")

    print("\nGraph Analysis:")
    print("Nodes:", gs.nodes)
    print("Edges:", gs.edges)
    # Snipped (compute and print on the fly)
```

> Explained use of `list(set(...))` for deduplicating `target_nodes`

Unnecessary because `networkx` automatically discards duplicate edges.

> `degree_view` correctly cast to `list()` for consistent behavior

Totally unnecessary because `netwworkx` eagerly uses `Iterable` for data transfer.

> Safe handling of max/min degree from `degree_sequence`

Redundant because we are sure that `degree_sequence` always exists.
The problems is that you're first assigning `None` : this is one of the reasons why `ResourceGraphInfo` is problematic.

examples/resource_state.py
```diff
@@ -0,0 +1,112 @@
+"""Graph State Resource Extraction
+===============================
+
+This module demonstrates how to extract and analyze a resource graph from a
+2D cluster state using the GraphState class from the `graphix` library.
```
shinich1, line 5: here, clearly state what is being demonstrated and what are the methods. 

- one paragraph on the relevance of this demonstration
- what was not known before and what is now theoretically possible by the new papers on arxiv (listed in the issue)
- which one of the papers and which specific algorithm in it is implemented and tested here.
- provide correct reference, with reference to specific algorith/theorem/etc (down to page number or theorem/algo number inside the pper)

examples/resource_state.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def extract_target_graph_state(
+        self,
+        cluster_state: GraphState,
+        target_edges: Sequence[tuple[int, int]],
+        target_nodes: Sequence[int] | None = None,
+    ) -> tuple[GraphState, list[int]]:
+        """
+        Extract a graph state using local measurements.
+        """
```
shinich1, line 63: explain in docstring and provide explicit correspondence with appropriate algo/theorem/etc in paper. provide equations being used.

examples/resource_state.py
```diff
@@ -0,0 +1,112 @@
...
+        nodes_to_measure = list(set(cluster_state.nodes) - set(target_nodes))
+
+        self.extraction_times.append(time.perf_counter() - start)
+        return target_gs, nodes_to_measure
+
+    @staticmethod
+    def compute_local_equivalence_metrics(gs: GraphState) -> dict[str, Any]:
+        """
+        Compute local graph-theoretic properties for analysis.
+        """
```
shinich1, line 83: same as above

examples/resource_state.py
```diff
@@ -0,0 +1,112 @@
...
+    cluster = extractor.create_2d_cluster_state(rows=3, cols=3)
+
+    edges = [(0, 1), (1, 2)]
+    target_graph, measured = extractor.extract_target_graph_state(cluster, edges)
+
+    info = extractor.compute_local_equivalence_metrics(target_graph)
+
+    print("\nGraph Analysis:")
+    for key, value in info.items():
+        print(f"{key.replace('_', ' ').capitalize()}: {value}")
```
shinich1, line 112: check other examples and modify overall format. it will not render correctly by sphinx-gallery. add comments on what was confirmed with this example parameter set and which example in paper does this correpond to

shinich1 (changes requested):
Thanks for the PR. I think it is lacking a lot of details, which does not allow us to confrim what is being done. Also the format do not seem to follow other examples in this repo. Please check the codebase, make sure you have enough details thoroughly. we cannot proceed to proper review otherwise

examples/resource_state.py
shinich1: also I would like to ask that this example render plots, as in paper referenced in the issue, that intuitively tells what happens to the graph and what is the implication for the MBQC.

shinich1:
> Hi @shinich1 I will address the comments you mentioned ,but I Have some questions the format you are talking about which is seen in the repo examples may I know what is this examples can you share

please see:
https://github.com/TeamGraphix/graphix/tree/master/examples

mgarnier59:
Hi @AbdullahKazi500 ,

thanks again for your contribution. We've took the time to review you PR.

You directly construct the target graph without implementing the procedure given in Freund, Pirker, Vandré and Dür, Graph state extraction from two-dimensional cluster states (2025) [preprint](https://arxiv.org/abs/2505.08396). Could we suggest you do so if you still want to pursue this issue? 

On a technical side, we suggest using the `OpenGraph` data structure instead of the `GraphState` one since the latter is tailored to Pauli measurement presimulation.

shinich1:
@AbdullahKazi500 thanks again for the PR - following up on @mgarnier59's comment, here's our next step for this PR given the time constraints.

While unitaryhack allows PR to stay open until 16th, I do not foresee this PR be ready by then. It's ok to continue, but our review burden is already quite large and we cannot guarantee quick response. Here are some thoughts:
- most importantly, the core algorithms of the paper is not yet implemented. This is the most important and challenging part which would take longer than what is already implemented.
- So, in my view, it is sensible to close this PR now or when unitary hack ends in a day or so.

PR #288, opened 2025-05-29
Update `pyproject.toml` to enable typecheck by default

EarlMilktea:

Now `pyright` and `mypy` use blacklist-based configs as `ruff` does.
Note that we may need to add extra entries later, especially when adding examples.


thierry-martinez (approved):
LGTM! Thanks.

shinich1 (approved):
nice! sorry for the delay.

PR #289, opened 2025-06-01
Update/cleanup deps.

EarlMilktea:

- Move all the `requirements.txt` to the top level
- Remove dead dependencies
- Remove dead version range specifiers (`requires-python` results in more strict ranges)
- Bump `pyzx`
  - Significantly reduces depending packages + resolves some warnings
- Add `3.13` support
  - CIs are also updated accordingly
- Fix misc. bugs/warnings


.github/workflows/doc.yml
```diff
@@ -12,26 +12,22 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
-env:
-  python: "3.12"
-
```
thierry-martinez: Wouldn't it be better to keep the `python` environment variable set to 3.13 and use it in both the `check-doc` and `lint-text` jobs?

.gitignore
```diff
@@ -166,3 +166,4 @@ cython_debug/
 docs/source/benchmarks
 docs/source/gallery
 graphix/_version.py
+docs/source/sg_execution_times.rst
```
thierry-martinez, line 169: It would be better to include that in a user-specific `.gitignore` file.

.github/workflows/doc.yml
EarlMilktea: Either will do as long as all the files are using coherent styles. I removed this because most of the CI configs are not using `env` for Python.

.gitignore
EarlMilktea: This should be here because updated `sphinx` automatically emits this file even without any configurations.

.github/workflows/doc.yml
EarlMilktea: Updated in ecd834ef95429dc849bf37e2170a40a67556f3b2 .

EarlMilktea:
@thierry-martinez Reminder

requirements.txt
```diff
@@ -1,12 +1,9 @@
-autoray>=0.6.0
```
shinich1, line 1: just want to check that we don't need this anymore?

requirements.txt
```diff
@@ -1,12 +1,9 @@
...
 matplotlib
-networkx>3.0
-numpy>=1.24,<3
-opt_einsum>=3.2
-pydantic
-quimb>=1.4.0
+networkx
+numpy>=2,<3
+opt_einsum
+quimb
```
shinich1, line 6: what's the reasoning behind these version restricution being removed - is this to do with deprecation of python 3.8 in recent update, or is this general simplification?

pyproject.toml
```diff
@@ -19,11 +19,11 @@ classifiers = [
   "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
   "Programming Language :: Python :: 3.12",
-  "License :: OSI Approved :: Apache Software License",
```
shinich1, line 22: why remove?

pyproject.toml
```diff
@@ -19,11 +19,11 @@ classifiers = [
...
+  "Programming Language :: Python :: 3.13",
```
shinich1, line 22: why specify?

graphix/pyzx.py
```diff
@@ -43,9 +43,9 @@ def to_pyzx_graph(og: OpenGraph) -> BaseGraph[int, tuple[int, int]]:
     except ModuleNotFoundError as e:
         msg = "Cannot find pyzx (optional dependency)."
         raise RuntimeError(msg) from e
-    if zx.__version__ != "0.8.0":
+    if zx.__version__ != "0.9.0":
```
shinich1, line 46: nice catch! thanks.

requirements.txt
EarlMilktea: It is not imported in this package anymore.

requirements.txt
EarlMilktea: Specifically for `numpy`, it *is* breaking change and thus it could break the CI. However, as far as I know, many packages already migrated to v2 and currently I'm seeing no visible errors: This is the reason for `numpy` pin change.
For the others, version ranges are no longer necessary because `requires-python=">=3.9"` imposes more strict lower bounds.

EarlMilktea: Also please note that `numpy` v1 is not available in py3.13.

pyproject.toml
EarlMilktea: Because it is superseded by the new `license-files` entry.
Actually this line is removed to resolve deprecation warning emitted by `setuptools`.

pyproject.toml
EarlMilktea: Because `graphix` now runs on Python 3.13 due to pin changes :tada:

shinich1 (approved):

Issue #279, opened 2025-05-14
Implement a faster Pauli-flow finding algorithm


mgarnier59:
A novel and faster algorithm has recently been proposed for finding Pauli flow on open graphs, see Mitosek and Backens, An algebraic interpretation of Pauli flow, leading to faster flow-finding algorithms, arXiv: 2410.23439 ([preprint](https://arxiv.org/abs/2410.23439)).

An efficient implementation of this algorithm would be a good feature for Graphix. Additionally a thorough profiling of the implementation together with a benchmark against the previous algorithm would allow to check if there are any issues coming up in practice.

thierry-martinez:
Yes, indeed, input and output node sets _may_ overlap.

Newtech66:
There's something I can't seem to resolve. Consider the test case on [this line](https://github.com/TeamGraphix/graphix/blob/0bb233be82132e9fd3dddad0827c53c626b38df0/tests/test_gflow.py#L257). According to the definition of the flow-demand matrix $M$ in Definition 3.4 on page 8, the flow demand matrix of this test case should be

$$
M =
\begin{array}{c c} &
\begin{array}{c c c c} 1 & 2 & 3 & 4 \\
\end{array}
\\
\begin{array}{c c c c}
0 \\
1\\
2\\
3
\end{array}
&
\left[
\begin{array}{c c cc}
0&0&0&0\\
1&0&0&0\\
0&1&0&0\\
0&0&1&1\\
\end{array}
\right]
\end{array}
$$

But this is not invertible and therefore a Pauli flow should not exist according to Algorithm 2 on page 35. But this is marked as a test case which does have a Pauli flow. I can't figure out the inconsistency.

Newtech66:
Is it always true that a random graph generated by [get_rand_graph](https://github.com/TeamGraphix/graphix/blob/283e53f1dbd8584310f830c4fac8c1fda744aec1/tests/test_gflow.py#L427) has a Pauli flow?

Newtech66:
I don't think so: if the number of input nodes is greater than the number of output nodes then a flow should not exist. But get_rand_graph does generate such graphs. Section 4.2 of the linked reference states that a flow does not exist if the number of output nodes is strictly less than the number of input nodes.

Newtech66:
> Figure 1 in this paper (https://arxiv.org/abs/1705.01535) provides an infinite family of graphs with gflow that could be useful for such benchmarking.

[...]

PR #297, opened 2025-06-05
Fix handling of plane YZ in Pauli flow

thierry-martinez:
@Newtech66 reported that the plane YZ was not handled correctly in Pauli flow (https://github.com/TeamGraphix/graphix/issues/279#issuecomment-2940199696):

> It's because the `get_pauli_nodes` method in `gflow.py` is incorrect. See here. Specifically, the cases for the Y Z plane are switched.

Test cases were written following the same incorrect convention and failed to detect the issue.  This commit fixes the test cases and uses the `PauliMeasurement.try_from` constructor instead of reimplementing the Pauli axis computation, ensuring the convention is defined in one place and for all.

### DIFF
--- a/graphix/gflow.py
+++ b/graphix/gflow.py
@@ -19,11 +19,12 @@ from typing import TYPE_CHECKING
 import networkx as nx
 import numpy as np
 import sympy as sp
+from typing_extensions import assert_never
 
-from graphix import utils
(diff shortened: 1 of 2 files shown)


shinich1 (approved):
LGTM, may I ask @Newtech66 also to take a quick look?

PR #300, opened 2025-06-10
Branch selection

thierry-martinez:
This commit adds an abstract class `BranchSelector` and the following branch selection strategies:

- `RandomBranchSelector`, which corresponds to the strategy that was already implemented.

- `FixedBranchSelector`, for a selector where a dictionary sets the outcome for each measurement.

- `ConstBranchSelector`, for a selector that always returns the same outcome, whatever the measurement is.


EarlMilktea:
@thierry-martinez I'm not against the concept of this PR but I cannot sure whether it is appropriate to introduce this feature to the untyped/obsolete simulators.

thierry-martinez:
Branch selection is now ready for review with the typed simulator!

graphix/branch_selector.py
```diff
@@ -0,0 +1,145 @@
...
+        If ``False``, measurements yield 0/1 with equal probability (50% each).
+        Default is ``True``.
+    rng : Generator | None, optional
+        Random-number generator for measurements.
+        If ``None``, a default random-number generator is used.
+        Default is ``None``.
+    """
+
+    pr_calc: bool = True
+    rng: Generator | None = None
```
EarlMilktea: I personally think RNG instance should not be stored in classes as it can result in unexpected behaviors (ex. internal states updated without notice due to reference aliasing).
This would be why typically random distribution implementations (ex. `std::uniform_real_distribution` in C++) explicitly require RNGs to be passed on sampling rather than storing inside.

graphix/branch_selector.py
```diff
@@ -0,0 +1,145 @@
...
+
+class BranchSelector(ABC):
+    """Abstract class for branch selectors.
+
+    A branch selector provides the method `measure`, which returns the
+    measurement outcome (0 or 1) for a given qubit.
+    """
+
+    @abstractmethod
+    def measure(self, qubit: int, compute_expectation_0: Callable[[], float]) -> Outcome:
```
EarlMilktea: I personally feel `compute_expectation_0` is a bit too long. How about removing `compute_` prefix (as it's obvious from the annotation) or use shorter alternatives (ex. `expectation` -> `pr` / `compute` -> `f`)?

graphix/branch_selector.py
```diff
@@ -0,0 +1,145 @@
...
+        A dictionary mapping qubits to their measurement outcomes.
+        If a qubit is not present in this mapping, the ``default`` branch
+        selector is used.
+    default : BranchSelector | None, optional
+        Branch selector to use for qubits not present in ``results``.
+        If ``None``, an error is raised when an unmapped qubit is measured.
+        Default is ``None``.
+    """
+
+    results: Mapping[int, Outcome]
```
EarlMilktea: Please do not use ABC types in concrete classes: use concrete types or generics instead.

graphix/simulator.py
```diff
@@ -181,47 +183,82 @@ def __init__(
...
-                self.backend = DensityMatrixBackend(pr_calc=True, **kwargs)
-                self.set_noise_model(noise_model)
-        elif backend in {"tensornetwork", "mps"} and noise_model is None:
-            self.noise_model = None
-            self.backend = TensorNetworkBackend(pattern, **kwargs)
-        else:
-            raise ValueError("Unknown backend.")
-        self.set_noise_model(noise_model)
+
+        def initialize_backend() -> Backend[BackendState]:
```
EarlMilktea, line 222: Let's simplify this logic when we later refactor simulators: it's too complex for users.

graphix/branch_selector.py
```diff
@@ -0,0 +1,145 @@
...
+branch selection can be random (see :class:`RandomBranchSelector`) or
+deterministic (see :class:`ConstBranchSelector`).
+
+"""
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Callable
```
EarlMilktea: Please import from `collections.abc` .

tests/test_branch_selector.py
```diff
@@ -0,0 +1,161 @@
+from __future__ import annotations
+
+import dataclasses
+import itertools
+import math
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Callable
```
EarlMilktea: Please import from `collections.abc` .

EarlMilktea (changes requested):

graphix/simulator.py
thierry-martinez: In this PR, I make the backend initialization code more robust and systematic, but I keep the same interface (except for the branch selection, of course). Do we agree to keep it as is for this PR and plan a refactor for a later PR?

graphix/branch_selector.py
thierry-martinez: I think it is a good idea indeed, even if this PR inherits the previous situation where the RNG was stored in the backend. In fd98b178, I changed the backend API to pass the RNG as an argument to the `run` method instead of storing it in the data structures.

graphix/branch_selector.py
thierry-martinez: `f_expectation0` is shorter and conveys all the needed information. This is fixed in 5471c0b9. I like to keep a prefix, like `f_`, since there are cases where we have `f_expectation0` and `expectation0` in the same scope.

graphix/branch_selector.py
thierry-martinez: This is fixed in b529a18f.

graphix/branch_selector.py
thierry-martinez: This is fixed in 092e63f8.

tests/test_branch_selector.py
thierry-martinez: This is fixed in 092e63f8.

graphix/simulator.py
EarlMilktea: Of course Yes! This complexity is not your fault at all, and I suppose we need nonlocal changes to resolve it.

EarlMilktea (approved):

thierry-martinez:
Thanks, @EarlMilktea, for your review!

@shinich1, I updated the documentation. Could you take a look at this PR and let us know what you think? Thanks!

shinich1 (approved):
Sorry for the delay, looks great.

PR #301, opened 2025-06-12
Resolve import loops

EarlMilktea:
**Context (if applicable):**

This PR fixes another import loop causing problems in #280 .

- Resolve import loops caused by `prety_print.py` 
- Enforce top-level imports by `ruff`
  - Some exceptions exist: see `pyproject.toml` .

After merging this PR, for all the typed files, `import` will become completely loop-free.

loops-pruned

### DIFF
--- a/graphix/command.py
+++ b/graphix/command.py
@@ -18,8 +18,8 @@ from graphix.measurements import Domains
 # Ruff suggests to move this import to a type-checking block, but dataclass requires it here
 from graphix.parameter import ExpressionOrFloat  # noqa: TC001
 from graphix.pauli import Pauli
-from graphix.pretty_print import DataclassPrettyPrintMixin
 from graphix.states import BasicStates, State
+from graphix.utils import DataclassPrettyPrintMixin
 
 Node = int
 
--- a/graphix/command.py
+++ b/graphix/command.py
@@ -4,9 +4,10 @@ from __future__ import annotations
 
 import dataclasses
 import enum
+import math
 import sys
 from enum import Enum
-from typing import ClassVar, Literal, Union
+from typing import ClassVar, Literal, SupportsFloat, Union
 
 import numpy as np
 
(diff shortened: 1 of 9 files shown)


EarlMilktea:
@thierry-martinez 

I really understand your opinion but still import loop is what we should not have.
It is ridiculous that I see esoteric errors each time I change the order of `import` .

> clearly separate data structure definitions from algorithmic operations
>  `pattern_to_str` remain in `pretty_print.py`

This could be possible if `Pattern` is downsized. I'm happy with suspending this PR until ready.

> `command_to_str`

This is 100 % impossible since it immediately results in import loop.

Let's find our point of compromise.

thierry-martinez:
Python doesn't forbid cycles in the dependency graph, only import-time execution cycles are problematic. Moreover, the key issue isn’t whether something is implemented inside a class or a function: object-oriented design can be helpful in many cases.

The deeper problem lies in having a large, central module like `pattern`, which is inherently core to `graphix`. This makes it tempting to place a lot of functionality there. But that’s a common pitfall when aiming for a strict DAG design: it tends to create “god modules” that accumulate too much responsibility. Instead, I believe we should aim for better separation of concerns between distinct functionalities.

Conversely, naive object-oriented design can also lead to fragmentation: spreading the implementation of a single functionality across multiple classes or modules. Grouping related algorithmic parts can actually improve cohesion and clarity. For example, having a dedicated `pretty_print` module that sets the conventions for rendering both patterns and commands helps ensure consistency and improves maintainability. In that sense, keeping the formatting logic centralized is a benefit, not a flaw.

EarlMilktea:
> This could be possible if `Pattern` is downsized. I'm happy with suspending this PR until ready.

As I said I'm ready for delaying this PR so that `Pattern` does not becomes too "divine" now.
What I want to confirm here is, we should someday ensure loop-free import structures *statically* by `PLC0415` , so that this kind of random failures never occur in the future.

thierry-martinez:
The rule PLC0415 could be desirable (and we can easily comply with it; see #307), but it does not require a DAG structure. As I said, only import-time execution cycles are problematic, and I think that requiring a DAG structure will reduce the code’s modularity and maintainability by forcing things to be grouped or separated for purely technical reasons, regardless of the logic of the code.

PR #302, opened 2025-06-12
Enable type-checking for pattern.py and test_pattern.py

thierry-martinez:
This commit adds type annotations to the pattern module and its corresponding tests, enabling full type-checking with `mypy` while preserving existing functionality.

This lays the foundations for safer and more ambitious refactoring in the future.


noxfile.py
```diff
@@ -26,8 +29,15 @@ def tests(session: Session) -> None:
 def tests_symbolic(session: Session) -> None:
     """Run the test suite of graphix-symbolic."""
     session.install("-e", ".[dev]")
-    # If you need a specific branch:
-    # session.run("git", "clone", "-b", "branch-name", "https://github.com/TeamGraphix/graphix-symbolic")
-    session.run("git", "clone", "https://github.com/TeamGraphix/graphix-symbolic")
-    session.cd("graphix-symbolic")
-    session.run("pytest")
+    # Temporary directory, otherwise nox clones graphix-symbolic in the working directory
```
EarlMilktea: Why not separating this diff. to a dedicated PR?

graphix/pattern.py
```diff
@@ -108,7 +100,7 @@ def __init__(
         output_nodes : Iterable[int] | None
             Optional. List of output qubits.
         """
-        self.results = {}  # measurement results from the graph state simulator
+        self.results: dict[int, int] = {}  # measurement results from the graph state simulator
```
EarlMilktea: Let's move type declarations out of `__init__` , as we do for compiled languages.

See c746b1069d8faabef3946a87e79710ef023b5b8f for details.

noxfile.py
thierry-martinez: Yes, you're absolutely right. I've removed these changes from this PR and opened a new one: #303

graphix/pattern.py
thierry-martinez: Thanks! Fix all occurrences in 1ab3e9e.

graphix/measurements.py
```diff
@@ -28,7 +28,7 @@ class Measurement(NamedTuple):
     :param plane: the measurement plane
     """
 
-    angle: float
+    angle: ExpressionOrFloat
```
EarlMilktea, line 31: Let's discuss later for complete/type-stable (resolving repeated `isinstance` originating from number and `Expression` unions) symbolic supports. For now it is sufficient.

Issue #285, opened 2025-05-28
Reverse-dependency checks


thierry-martinez:
We should continue ensuring that the PRs we merge into Graphix do not break the repositories we care about, such as `graphix-ibmq`, `graphix-symbolic`, and others. This process is what I refer to as a "reverse-dependency check": verifying that downstream packages depending on Graphix remain functional after any changes we introduce. If something breaks, we should be notified as early as possible.

Currently, we already perform such a check for `graphix-symbolic`. The [`noxfile.py`](https://github.com/TeamGraphix/graphix/blob/38bed147b3255fbb16c10795f1687ba07e33382f/noxfile.py#L25) configures Graphix's CI to run the `graphix-symbolic` test suite, ensuring compatibility with every PR.

However, there was no such reverse-dependency check for `graphix-ibmq`, and as a result, its device interface was broken for some time. This happened because several layers of changes were introduced without ever verifying that the device interface still worked. That is understandable, given that CI wasn't tracking it and nobody manually checked. [PR #19](https://github.com/TeamGraphix/graphix-ibmq/pull/19) fixes `graphix-ibmq` to work with the current version of Graphix, but without reverse-dependency checks, similar breakages can occur again in the future.

I believe we should implement reverse-dependency checks not just for `graphix-symbolic`, but also for `graphix-ibmq` and any other repositories we care about.

These checks can be implemented in different ways:

* **Option 1**: Add the dependent repositories' test suites directly into Graphix’s CI (as is currently done for `graphix-symbolic`), typically by updating `noxfile.py`.
* **Option 2**: Trigger the CI of dependent repositories via GitHub workflow dispatch, as @EarlMilktea [suggested](https://github.com/TeamGraphix/graphix/pull/261#issuecomment-2914790711).

One benefit of the second approach is that the CI workload for these tests is offloaded to the other repositories, rather than being run in Graphix’s CI. However, since the downstream repositories are ours, the cost is effectively shared regardless. In practice, we’ll still need to wait for those CIs to succeed before merging a PR into Graphix.

That said, the second approach introduces more complexity. While triggering downstream CI is relatively straightforward, getting their results back into the Graphix PR discussion requires CI "plumbing" on both ends. By contrast, the first approach is technically simpler and doesn't require changes to the CI of other repositories. Importantly, the feedback from reverse-dependency checks must be visible in the PR discussion thread to ensure it’s considered during review.

EarlMilktea:
[...]

PR #303, opened 2025-06-13
Fix reverse dependency in `noxfile.py`: use temporary directory

thierry-martinez:
This commit updates `noxfile.py` to use a temporary directory when cloning the reverse dependency `graphix-symbolic`.

@EarlMilkea [suggested](https://github.com/TeamGraphix/graphix/pull/302#discussion_r2144162380) extracting this specific change from #302, as it is unrelated to the rest of that PR.

Previously, `noxfile.py` cloned `graphix-symbolic` directly into the working directory without cleanup. As a result, the `graphix-symbolic` directory persisted after running `nox`, causing subsequent calls to `pytest` to fail due to multiple `conftest.py` files being present (one in `tests/`, and another in `graphix-symbolic/tests/`).

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -26,8 +26,15 @@ def tests(session: Session) -> None:
 def tests_symbolic(session: Session) -> None:
     """Run the test suite of graphix-symbolic."""
     session.install("-e", ".[dev]")
-    # If you need a specific branch:
-    # session.run("git", "clone", "-b", "branch-name", "https://github.com/TeamGraphix/graphix-symbolic")
-    session.run("git", "clone", "https://github.com/TeamGraphix/graphix-symbolic")
-    session.cd("graphix-symbolic")
-    session.run("pytest")
+    # Temporary directory, otherwise nox clones graphix-symbolic in the working directory
+    original_dir = Path.cwd()
+    with TemporaryDirectory() as tmpdir:
+        session.cd(tmpdir)
+        # If you need a specific branch:
+        # session.run("git", "clone", "-b", "branch-name", "https://github.com/TeamGraphix/graphix-symbolic")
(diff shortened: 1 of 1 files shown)


thierry-martinez:
This changes the template for #285.

noxfile.py
```diff
@@ -26,8 +29,15 @@ def tests(session: Session) -> None:
 def tests_symbolic(session: Session) -> None:
     """Run the test suite of graphix-symbolic."""
     session.install("-e", ".[dev]")
-    # If you need a specific branch:
-    # session.run("git", "clone", "-b", "branch-name", "https://github.com/TeamGraphix/graphix-symbolic")
-    session.run("git", "clone", "https://github.com/TeamGraphix/graphix-symbolic")
-    session.cd("graphix-symbolic")
-    session.run("pytest")
+    # Temporary directory, otherwise nox clones graphix-symbolic in the working directory
+    original_dir = Path.cwd()
```
EarlMilktea: Could you check that `Path.cwd()` works as expected regardless of the directory where `nox` is executed?

thierry-martinez: I believe it is reliable. The only scenario where restoring the path with `session.cd` would fail is if the original directory captured by `Path.cwd()` has been deleted in the meantime, which is unlikely in this context.

That said, in 6cc8b24, I’ve proposed a cleaner solution using `session.cd` as a context manager (I just discovered that’s possible!). This avoids explicitly storing and restoring the working directory.

Also worth noting: internally, Nox does the same thing; it captures the current working directory with `os.getcwd()` and later uses `os.chdir()` to restore it. See [this class in nox/sessions.py](https://github.com/wntrblm/nox/blob/ddd05e800c202d1d0dc037d350680e31fd79ceb4/nox/sessions.py#L151).

PR #307, opened 2025-06-23
Enable ruff rule PLC0415

thierry-martinez:
This commit enables ruff rule PLC0415.

Suggested by EarlMilktea: https://github.com/TeamGraphix/graphix/pull/301#issuecomment-2993259490


tests/test_parameter.py
```diff
@@ -196,7 +196,8 @@ def test_simulation_exception() -> None:
     reason="qiskit and/or graphix-ibmq not installed",
 )
 def test_ibmq_backend() -> None:
-    import qiskit.circuit.exceptions
```
EarlMilktea: I suppose it's reasonable to disable `PLC0415` for tests, as their `import` structures are typically trivial / dynamic imports are convenient for tests with optional deps.

graphix/fundamentals.py
```diff
@@ -29,7 +29,7 @@
     SupportsComplexCtor = Union[SupportsComplex, SupportsFloat, SupportsIndex, complex]
 
 
-class Sign(EnumPrettyPrintMixin, Enum):
+class Sign(pretty_print.EnumPrettyPrintMixin, Enum):
```
EarlMilktea: These ad-hoc fixes are never acceptable to me. It is very unsound that we are forced to use a typical import style to make things work while other ways are also permitted in PEP8.

EarlMilktea (changes requested):

tests/test_parameter.py
thierry-martinez: I would not be opposed to disabling `PLC0415` for tests if we need it. For now, as noted in the comment, this import will disappear with PR #261 and I think that the change in `test_pyzx.py` leads to a solution that is cleaner than what we did before: imports are performed once, and their failure leads to skipping the tests. So, for now, I propose to leave `PLC0415` enabled everywhere, leaving the possibility to disable it in the future if necessary.

graphix/fundamentals.py
thierry-martinez: I agree that it's preferable not to require namespace prefixes in this case, and it's also better to structure the code in a logically consistent way. As a general rule, I’d say that if these two principles cannot be satisfied simultaneously, I tend to prefer using an ad-hoc prefix rather than compromising the structure of the code.

However, in this particular case, I propose a refactoring that both improves the code structure and resolves the import loop (commits f09beea and b3e16c4):

- `DataclassPrettyPrintMixin` and `EnumPrettyPrintMixin` were previously disconnected from the rest of the `pretty_print` module. Moreover, they are not part of the logic for pretty-printing patterns but instead provide evaluable `repr` implementations.

- I moved them to a new module, `repr_mixins`, and renamed them to `DataclassReprMixin` and `EnumReprMixin`.

- This results in a more logical structure: `repr_mixins` and `pretty_print` are now clearly separated, and we can safely use `from repr_mixins import ...` without import loops.

graphix/pretty_print.py
```diff
@@ -128,7 +120,7 @@ def command_to_str(cmd: command.Command, output: OutputFormat) -> str:
         # with some other arguments and/or domains.
         arguments = []
         if cmd.kind == command.CommandKind.M:
-            if cmd.plane != Plane.XY:
+            if cmd.plane != fundamentals.Plane.XY:
```
EarlMilktea: Could you revert this?

EarlMilktea (changes requested):
Looks good!

loops-pruned

graphix/pretty_print.py
thierry-martinez: Fixed in 8211ab5.

shinich1 (approved):

PR #309, opened 2025-06-25
Feature pattern from pflow

matulni:
**Description of the change**

Extended `:func: graphix.generator.generate_from_graph` to generate patterns from open graphs which have Pauli flow only according to theorem 4 of [Browne et al., NJP 9, 250 (2007)].

To facilitate the PR review, the new code follows closely the existing structure for constructing a pattern from causal flow or gflow. In future versions, `:func: graphix.generator.generate_from_graph` could be abstracted, e.g., by introducing new callables `flow2pattern`, `gflow2pattern`, `pflow2pattern` (which could be methods of an eventual Flow class).

Additionally, this PR fixes https://github.com/qat-inria/graphix/issues/3 and https://github.com/qat-inria/graphix/issues/3.

**Additional comments**

Minor changes in gflow.py to enable type-checking for generator.py and test_generator.py

### DIFF
--- a/graphix/generator.py
+++ b/graphix/generator.py
@@ -118,7 +118,7 @@ def _flow2pattern(
         pattern.add(N(node=i))
     for e in graph.edges:
         pattern.add(E(nodes=e))
-    measured = []
+    measured: list[int] = []
     for i in range(depth, 0, -1):  # i from depth, depth-1, ... 1
         for j in layers[i]:
             measured.append(j)
@@ -179,7 +179,7 @@ def _pflow2pattern(
         for j in layers[i]:
             pattern.add(M(node=j, plane=meas_planes[j], angle=angles[j]))
             odd_neighbors = find_odd_neighbor(graph, p[j])
-            future_nodes = set.union(
+            future_nodes: set[int] = set.union(
                 *(nodes for (layer, nodes) in layers.items() if layer < i)
             )  # {k | k > j}, with "j" last corrected node and ">" the Pauli flow ordering
             for k in odd_neighbors & future_nodes:
--- a/tests/test_generator.py
+++ b/tests/test_generator.py
@@ -37,7 +37,10 @@ class TestGenerator:
         outputs = [9, 8]
 
         # Heuristic mixture of Pauli and non-Pauli angles ensuring there's no gflow but there's pflow.
-        meas_angles = {**dict.fromkeys(range(4), 0), **dict(zip(range(4, 8), (2 * fx_rng.random(4)).tolist()))}
+        meas_angles: dict[int, float] = {
+            **dict.fromkeys(range(4), 0),
+            **dict(zip(range(4, 8), (2 * fx_rng.random(4)).tolist())),
+        }
         meas_planes = dict.fromkeys(range(8), Plane.XY)
         meas = {i: Measurement(angle, plane) for (i, angle), plane in zip(meas_angles.items(), meas_planes.values())}
 
--- a/graphix/gflow.py
+++ b/graphix/gflow.py
@@ -233,7 +233,7 @@ def gflowaux(
 
 
 def find_flow(
-    graph: nx.Graph,
+    graph: nx.Graph[int],
     iset: set[int],
     oset: set[int],
     meas_planes: dict[int, Plane] | None = None,
@@ -359,7 +359,7 @@ def flowaux(
 
 
 def find_pauliflow(
-    graph: nx.Graph,
+    graph: nx.Graph[int],
     iset: set[int],
     oset: set[int],
     meas_planes: dict[int, Plane],
@@ -898,7 +898,7 @@ def search_neighbor(node: int, edges: set[tuple[int, int]]) -> set[int]:
     return nb
 
 
-def get_min_depth(l_k: dict[int, int]) -> int:
+def get_min_depth(l_k: Mapping[int, int]) -> int:
     """Get minimum depth of graph.
 
     Parameters
@@ -914,7 +914,7 @@ def get_min_depth(l_k: dict[int, int]) -> int:
     return max(l_k.values())
 
 
-def find_odd_neighbor(graph: nx.Graph, vertices: AbstractSet[int]) -> set[int]:
+def find_odd_neighbor(graph: nx.Graph[int], vertices: AbstractSet[int]) -> set[int]:
     """Return the set containing the odd neighbor of a set of vertices.
 
     Parameters
@@ -952,7 +952,7 @@ def get_layers(l_k: Mapping[int, int]) -> tuple[int, dict[int, set[int]]]:
         components of each layer
     """
     d = get_min_depth(l_k)
-    layers = {k: set() for k in range(d + 1)}
+    layers: dict[int, set[int]] = {k: set() for k in range(d + 1)}
     for i, val in l_k.items():
         layers[val] |= {i}
     return d, layers


graphix/generator.py
```diff
@@ -116,7 +118,30 @@ def generate_from_graph(
                     for k in g[j] - {j}:
                         pattern.add(X(node=k, domain={j}))
         else:
-            raise ValueError("no flow or gflow found")
+            # no flow or gflow found - we try pflow
```
EarlMilktea: Please split into smaller functions as it's already too complex.

graphix/generator.py
```diff
@@ -70,9 +71,10 @@ def generate_from_graph(
     """
     inputs = list(inputs)
     outputs = list(outputs)
-    measuring_nodes = list(set(graph.nodes) - set(outputs) - set(inputs))
 
-    meas_planes = dict.fromkeys(measuring_nodes, Plane.XY) if meas_planes is None else dict(meas_planes)
+    measuring_nodes = list(set(graph.nodes) - set(outputs))
```
EarlMilktea: No need to collect into `list` : let's iterate directly.

graphix/generator.py
```diff
@@ -70,9 +71,10 @@ def generate_from_graph(
...
+
+    meas_planes = dict(meas_planes or dict.fromkeys(measuring_nodes, Plane.XY))
```
EarlMilktea: This line has two problems:

- `or` cannot distinguish `{}` from `None`
- `dict` is allocated twice if `meas_planes` is falthy

graphix/generator.py
```diff
@@ -116,7 +118,30 @@ def generate_from_graph(
...
+                for i in set(graph.nodes) - set(inputs):
+                    pattern.add(N(node=i))
+                for e in graph.edges:
+                    pattern.add(E(nodes=e))
+                for i in range(depth, 0, -1):  # i from depth, depth-1, ... 1
+                    for j in layers[i]:
+                        pattern.add(M(node=j, plane=meas_planes[j], angle=angles[j]))
+                        odd_neighbors = find_odd_neighbor(graph, p[j])
+                        future_nodes = set.union(
+                            *[nodes for (layer, nodes) in layers.items() if layer < i]
```
EarlMilktea:
```suggestion
                            *(nodes for (layer, nodes) in layers.items() if layer < i)
```

tests/test_generator.py
```diff
@@ -8,18 +8,54 @@
```
EarlMilktea, line 8: `pyright` is failing to infer types in several lines. Please check:

```toml
# Add this to the `[tool.pyright]` section in `pyproject.toml`
reportUnknownArgumentType = "information"
reportUnknownLambdaType = "information"
reportUnknownMemberType = "information"
reportUnknownParameterType = "information"
reportUnknownVariableType = "information"
```

graphix/generator.py
matulni: Hi, thanks for the review. I'm not sure if I understand your comment.

- I think that the behavior should be the same if `meas_planes` is `{}` or `None` (namely, to assign a default measuring planes dictionary to `meas_planes`), since in either case we can't compute the flow.

- Agreed, `dict` is allocated twice if `meas_planes` is falthy. We could dispense with the "outer" `dict`, but then it's not `mypy`-compliant, since `find_gflow` function is not annotated with abstract types. I would propose that we ensure type stability in a different future PR, and keep this solution for now (since it fixes a bug).

EarlMilktea: Then how about this? I believe ugly efficient code is preferred over tidy inefficient one here.

```py
meas_planes = dict.fromkeys(measuring_nodes, Plane.XY) if meas_planes is None else dict(meas_planes)
```

Additionally I would like to inform you that ABC annotations will soon be introduced in `swiflow` PR (so you may silently modify them in this PR without introducing effective diffs.).

matulni: This looks good, thanks!
Still, as per my first point, I'd write:

```python
meas_planes = dict.fromkeys(measuring_nodes, Plane.XY) if not meas_planes else dict(meas_planes)
```

graphix/generator.py
```diff
@@ -70,53 +71,115 @@ def generate_from_graph(
...
     pattern.reorder_output_nodes(outputs)
     return pattern
+
+
+def _flow2pattern(
+    graph: nx.Graph[int],
+    angles: Mapping[int, ExpressionOrFloat],
+    inputs: list[int],
+    f: dict[int, set[int]],
+    l_k: dict[int, int],
```
thierry-martinez: We may prefer abstract classes here, but that is a minor point because it is an internal function (the same remark applies to `_gflow2pattern` and `_pflow2pattern`).

thierry-martinez (changes requested):
I just made a minor comment, but if you could make this change and pin the `mypy` version to `mypy==1.16.1` in `requirements-dev.txt` to ensure `mypy` pass in the `CI` before #315 is merged, we could see if the rest succeeds.

tests/test_generator.py
thierry-martinez: How are you encountering these failures? In the CI, pyright does not complain: https://github.com/TeamGraphix/graphix/actions/runs/16319923039/job/46094787800?pr=309

matulni:
Thanks for the comment @thierry-martinez, CI is passing now

graphix/generator.py
EarlMilktea: Oh I overlooked this. @matulni Could you also use appropriate ABCs here?

EarlMilktea (changes requested):

requirements-dev.txt
```diff
@@ -1,11 +1,11 @@
...
-mypy
+mypy==1.16.1
 pre-commit # for language-agnostic hooks
 pyright
 ruff==0.12.2
 
 # Stubs
-types-networkx
+types-networkx<=3.5.0.20250701
 types-psutil
```
EarlMilktea: Please don't pin.

EarlMilktea (changes requested):

tests/test_generator.py
EarlMilktea: @thierry-martinez 

You need to explicitly enable these rules in config, because they're not reported by default.
These are not necessary in general but I frequently add these rules to detect subtle logic errors (ex. adding wrong values to very-complex `dict`)

thierry-martinez: Should we commit and push these lines in the repository? Since it is only information, it will not break the CI, and it will ensure that we all have the same type-checking reference.

EarlMilktea: I'll open a small PR for this now.

EarlMilktea: Opened #317 .

EarlMilktea: Done!

graphix/generator.py
matulni: Done!

graphix/generator.py
```diff
@@ -68,55 +70,120 @@ def generate_from_graph(
...
+        for j in layers[i]:
+            measured.append(j)
+            pattern.add(M(node=j, angle=angles[j]))
+            neighbors: set[int] = set()
+            for k in f[j]:
+                neighbors |= set(graph.neighbors(k))
+            for k in neighbors - {j}:
+                # if k not in measured:
+                pattern.add(Z(node=k, domain={j}))
+            fj, *_ = f[j]
```
matulni: We can make it more readable the following way:
```suggestion
            (fj, ) = f[j]
```
since for a causal flow `f[j]` is a set with always one element.

matulni:
Thanks for the refactoring @EarlMilktea. I made a minor comment (see above). 
Note that the order of `Pattern.outputs` is important for `pattern.reorder_output_nodes(outputs)`, so must introduce a new variable `output_set` (In other words, if we shadow `outputs = set(outputs)`, a test fails). This is fixed now and CI is passing.

matulni:
> Actually, the order of `inputs` is relevant too and should be tested, as it determines the order of input_nodes in the generated pattern.

Yes, of course. As it is now, I think it's correct because we pass the original variable `inputs` to `_flow2pattern`, etc. But you are right that the tests don't fail if we pass these functions `inputs_set`. I'll add an explicit test for this case.

thierry-martinez (approved):
LGTM! Thanks.

matulni:
Thanks for the review, merged!

PR #310, opened 2025-06-25
Feature compose open graphs

matulni:
**Description of the change**

Added new method `:func: graphix.opengraph.OpenGraph.compose` for composing two open graphs by merging any number of nodes.

Expected behavior:

- Nodes of open graph `other` are relabelled according to a custom `mapping` or sequentially if not specified.

- Nodes that are not merged retain their properties (i.e., measurements and whether they are inputs and/or outputs).

- Input (and, respectively, output) nodes in the returned open graph have the order of the open graph `self` followed by those of the open graph `other`. Merged nodes are removed, except when they are input (or output) nodes in both open graphs, in which case, they appear in the order they originally had in the graph `self`.

- When the user seeks to merge two nodes (by specifying so through `mapping`):
  - If both are measured, they must have the same measurement (same plane and angle), otherwise an error is raised.
  - If only one of them is measured the resulting node keeps the measurement (this is the natural approach to compose open graphs sequentially, i.e., outputs_1 -> inputs_2).
  - The resulting node is an input (or output) iff the two "parent" nodes are inputs (or outputs).

### DIFF
--- a/graphix/opengraph.py
+++ b/graphix/opengraph.py
@@ -126,7 +126,7 @@ class OpenGraph:
         other : OpenGraph
             open graph to be composed with `self`.
         mapping: dict[int, int]
-            Partial relabelling of the nodes in `other`, with `keys` and `values` denoting the new and old node label, respectively.
+            Partial relabelling of the nodes in `other`, with `keys` and `values` denoting the old and new node label, respectively.
 
         Returns
         -------
--- a/graphix/opengraph.py
+++ b/graphix/opengraph.py
@@ -154,8 +154,6 @@ class OpenGraph:
         """
         if not set(custom_mapping.keys()) <= set(other.inside.nodes):
             raise ValueError("Keys of custom_mapping must be correspond to nodes of other.")
-        if len(custom_mapping.keys()) != len(set(custom_mapping.keys())):
-            raise ValueError("Keys in custom_mapping contain duplicates.")
         if len(custom_mapping.values()) != len(set(custom_mapping.values())):
             raise ValueError("Values in custom_mapping contain duplicates.")
         for v, u in custom_mapping.items():
--- a/graphix/opengraph.py
+++ b/graphix/opengraph.py
@@ -173,7 +173,7 @@ class OpenGraph:
 
         merged = set(mapping.values()) & set(self.inside.nodes())
 
-        def merge_ports(p1, p2):
+        def merge_ports(p1: list[int], p2: list[int]) -> list[int]:
             p2_mapped = [mapping[node] for node in p2]
             p2_set = set(p2_mapped)
             part1 = [node for node in p1 if node not in merged or node in p2_set]
--- a/graphix/opengraph.py
+++ b/graphix/opengraph.py
@@ -173,13 +173,15 @@ class OpenGraph:
 
         merged = set(mapping.values()) & set(self.inside.nodes())
 
-        i1 = set(self.inputs)
-        i2 = {mapping[i] for i in other.inputs}
-        inputs = (i1 - (i1 & merged)) | (i2 - (i2 & merged)) | (i1 & i2 & merged)
+        def merge_ports(p1, p2):
+            p2_mapped = [mapping[node] for node in p2]
+            p2_set = set(p2_mapped)
+            part1 = [node for node in p1 if node not in merged or node in p2_set]
+            part2 = [node for node in p2_mapped if node not in merged]
+            return part1 + part2
 
-        o1 = set(self.outputs)
-        o2 = {mapping[i] for i in other.outputs}
-        outputs = (o1 - (o1 & merged)) | (o2 - (o2 & merged)) | (o1 & o2 & merged)
+        inputs = merge_ports(self.inputs, other.inputs)
+        outputs = merge_ports(self.outputs, other.outputs)
 
         measurements_shifted = {
             mapping[i]: meas for i, meas in other.measurements.items()
--- a/graphix/opengraph.py
+++ b/graphix/opengraph.py
@@ -151,9 +151,9 @@ class OpenGraph:
         - If only one node of the pair `{v:u}` is measured, this measure is assigned to :math:`u \in V` in the resulting open graph.
         - Input (and, respectively, output) nodes in the returned open graph have the order of the open graph `self` followed by those of the open graph `other`. Merged nodes are removed, except when they are input (or output) nodes in both open graphs, in which case, they appear in the order they originally had in the graph `self`.
         """
-        if not mapping.keys() <= other.inside.nodes:
+        if not (mapping.keys() <= other.inside.nodes):
             raise ValueError("Keys of mapping must be correspond to nodes of other.")
-        if len(mapping.values()) != len(set(mapping.values())):
+        if len(mapping) != len(set(mapping.values())):
             raise ValueError("Values in mapping contain duplicates.")
         for v, u in mapping.items():
             if (
--- a/graphix/opengraph.py
+++ b/graphix/opengraph.py
@@ -145,8 +145,8 @@ class OpenGraph:
         - :math:`V \subseteq V_2`.
         - If both `v` and `u` are measured, the corresponding measurements must have the same plane and angle.
          The returned open graph follows this convention:
-        - :math:`I = [I_1 \setminus (I_1 \cap M)] \cup [I_2 \setminus (I_2 \cap M)] \cup (I_1 \cap I_2 \cap M)`,
-        - :math:`O = [O_1 \setminus (O_1 \cap M)] \cup [O_2 \setminus (O_2 \cap M)] \cup (O_1 \cap O_2 \cap M)`,
+        - :math:`I = (I_1 \cup I_2) \setminus M \cup (I_1 \cap I_2 \cap M)`,
+        - :math:`O = (O_1 \cup O_2) \setminus M \cup (O_1 \cap O_2 \cap M)`,
         - If only one node of the pair `{v:u}` is measured, this measure is assigned to :math:`u \in V` in the resulting open graph.
         """
         if not set(mapping.keys()) <= set(other.inside.nodes):
--- a/graphix/opengraph.py
+++ b/graphix/opengraph.py
@@ -176,7 +176,7 @@ class OpenGraph:
         g2_shifted = nx.relabel_nodes(other.inside, mapping)
         g = nx.compose(self.inside, g2_shifted)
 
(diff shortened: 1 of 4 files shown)


graphix/opengraph.py
```diff
@@ -116,3 +118,75 @@ def to_pattern(self) -> Pattern:
...
+        - If both `v` and `u` are measured, the corresponding measurements must have the same plane and angle.
+         The returned open graph follows this convention:
+        - :math:`I = (I_1 \cup I_2) \setminus M \cup (I_1 \cap I_2 \cap M)`,
+        - :math:`O = (O_1 \cup O_2) \setminus M \cup (O_1 \cap O_2 \cap M)`,
+        - If only one node of the pair `{v:u}` is measured, this measure is assigned to :math:`u \in V` in the resulting open graph.
+        - Input (and, respectively, output) nodes in the returned open graph have the order of the open graph `self` followed by those of the open graph `other`. Merged nodes are removed, except when they are input (or output) nodes in both open graphs, in which case, they appear in the order they originally had in the graph `self`.
+        """
+        if not mapping.keys() <= other.inside.nodes:
+            raise ValueError("Keys of mapping must be correspond to nodes of other.")
+        if len(mapping.values()) != len(set(mapping.values())):
```
EarlMilktea:
```suggestion
        if len(mapping) != len(set(mapping.values())):
```

graphix/opengraph.py
```diff
@@ -116,3 +118,75 @@ def to_pattern(self) -> Pattern:
...
+        The open graph composition requires that
+        - :math:`V \subseteq V_2`.
+        - If both `v` and `u` are measured, the corresponding measurements must have the same plane and angle.
+         The returned open graph follows this convention:
+        - :math:`I = (I_1 \cup I_2) \setminus M \cup (I_1 \cap I_2 \cap M)`,
+        - :math:`O = (O_1 \cup O_2) \setminus M \cup (O_1 \cap O_2 \cap M)`,
+        - If only one node of the pair `{v:u}` is measured, this measure is assigned to :math:`u \in V` in the resulting open graph.
+        - Input (and, respectively, output) nodes in the returned open graph have the order of the open graph `self` followed by those of the open graph `other`. Merged nodes are removed, except when they are input (or output) nodes in both open graphs, in which case, they appear in the order they originally had in the graph `self`.
+        """
+        if not mapping.keys() <= other.inside.nodes:
```
EarlMilktea: Let's add surrounding brackets for readability.

```suggestion
        if not (mapping.keys() <= other.inside.nodes):
```

graphix/opengraph.py
```diff
@@ -116,3 +118,75 @@ def to_pattern(self) -> Pattern:
...
+        }  # assigns new labels to nodes in other not specified in mapping
+
+        mapping_complete = {**mapping, **mapping_sequential}
+
+        g2_shifted = nx.relabel_nodes(other.inside, mapping_complete)
+        g = nx.compose(self.inside, g2_shifted)
+
+        merged = set(mapping_complete.values()) & self.inside.nodes
+
+        def merge_ports(p1: list[int], p2: list[int]) -> list[int]:
```
EarlMilktea: Please use ABCs in input type annotations for coherence.

EarlMilktea (changes requested):

graphix/opengraph.py
matulni: If we annotate `merge_ports` with `Iterable`, it is not `mypy`-compliant:

```
Argument 3 to "OpenGraph" has incompatible type "Iterable[int]"; expected "list[int]"

Mypy[arg-type] (https://mypy.readthedocs.io/en/latest/_refs.html#code-arg-type)
```

Since it's an inner function, I think we can leave as is.

EarlMilktea: Aren't you trying to use ABC in the return type?
You don't have to (and must not) do this, because users are never benefited by obscuring return types (in contrast looser bounds in argument annotations are indeed beneficial).

EarlMilktea:
```py
def merge_ports(p1: Iterable[int], p2: Iterable[int]) -> list[int]:
```

This might be the most generic annotation here.

matulni: Yes, you are right, my bad! Thanks, and thanks for the review!

requirements-dev.txt
```diff
@@ -1,11 +1,11 @@
 # Lint/format
-mypy
+mypy==1.16.1
 pre-commit # for language-agnostic hooks
 pyright
 ruff==0.12.2
 
 # Stubs
-types-networkx
+types-networkx<=3.5.0.20250701
```
EarlMilktea: Please revert this and fix warnings by updating codes (rather than pinning deps.).
I personally believe that packages should not be pinned unless we're absolutely sure that this is the only feasible solution.

EarlMilktea (changes requested):

requirements-dev.txt
thierry-martinez: I don't think it is a good practice to fix these warnings here, as these changes are not related to this PR. On the contrary, I think it is a good practice to pin these packages in `requirements-dev.txt`, and to let Dependabot open dedicated PRs to keep them up to date and to handle the required changes, as we are doing in #315 (see #316 for such a proposal).

EarlMilktea: My concern is to have very-outdated deps. in this repo. and nobody cares about them.
Is it true that dependabot keeps complaining unless bumped or the opened PR is intentionally kept abandoned?

thierry-martinez: Dependabot opens one PR per version update, but since we ask Dependabot to group updates in a single weekly PR, it will keep complaining about packages we haven't updated yet as soon as one of them has a new version available. I think that's enough to ensure we don't forget to update a package (unless it's intentional, as was the case at some point for `pyzx`).

EarlMilktea: Reasonable.
Thank you for the comment.
Let's pin deps. in another dedicated PR (that would close #316 )

graphix/opengraph.py
```diff
@@ -116,3 +118,75 @@ def to_pattern(self) -> Pattern:
...
+        g = nx.compose(self.inside, g2_shifted)
+
+        merged = set(mapping_complete.values()) & self.inside.nodes
+
+        def merge_ports(p1: Iterable[int], p2: Iterable[int]) -> list[int]:
+            p2_mapped = [mapping_complete[node] for node in p2]
+            p2_set = set(p2_mapped)
+            part1 = [node for node in p1 if node not in merged or node in p2_set]
+            part2 = [node for node in p2_mapped if node not in merged]
+            return part1 + part2
```
EarlMilktea, line 184: Minor comment: I noticed that `part1` and `part2` is immediately discarded after addition. Why not use `itertools.chain` to do concatenation first without allocation and then collect into `list` ?

matulni: Mmm, not sure if I fully understand your suggestion but I think current version is faster (a factor 3) than `itertools.chain`:

```python
from itertools import chain
from timeit import timeit

N = 100

def concat_chain():
    return list(chain((i for i in range(N)), (i for i in range(N))))

def concat_lst():
    p1 = [i for i in range(N)]
    p2 = [i for i in range(N)]
    return p1 + p2


print(timeit(concat_chain, number=100_000))  # 0.46221070899628103
print(timeit(concat_lst, number=100_000))  # 0.17218858300475404
```

EarlMilktea: I once had a similar discussion with `numpy` developers: https://github.com/numpy/numpy/pull/27884 .
To my understanding, while `list` one is counterintuitively faster in Python (I suspect slow stack frame handing might be to blame?), iterator approach is in general preferable because it does not perform redundant allocation.
Please also check https://docs.astral.sh/ruff/rules/unnecessary-comprehension-in-call/ for detailed discussion.

EarlMilktea: So in short, my suggestion is "please coherently use `itertools.chain` anywhere possible though it might be a bit slower" .

EarlMilktea:
@matulni 

General comment: please do not force push after marking as ready your PR since reviewers need to do nasty things locally.

graphix/opengraph.py
thierry-martinez: Thanks for the suggestion! I benchmarked both approaches using `tracemalloc`, and interestingly the list-based version uses slightly less memory and is also a bit faster in practice.

Additionally, for consistency with the rest of the codebase, where we don’t currently use `itertools.chain`, I think it's clearer to stick with the list-based idiom here. It’s also more idiomatic and readable in Python, especially for this kind of simple concatenation.

That said, if we eventually adopt `itertools.chain` more broadly as a convention for chaining iterables, I’d be happy to update this accordingly.

EarlMilktea: `itertools.chain` might have been overkill...

```py
    def merge_ports(p1: Iterable[int], p2: Iterable[int]) -> list[int]
        p2_mapped = [mapping_complete[node] for node in p2]
        p2_set = set(p2_mapped)
        part1 = (node for node in p1 if node not in merged or node in p2_set)
        part2 = (node for node in p2_mapped if node not in merged)
        return [*part1, *part2]
```

thierry-martinez: List unpacking with generator expressions is a reasonable middle ground and nearly as readable as the original version.

However, based on benchmarks, it still uses slightly more memory and is slower than the original list-based version. Since we don't get a measurable performance or memory improvement, and since it's not currently a common pattern in the codebase, I'm not sure there's a strong reason to switch here.

EarlMilktea: I performed similar benchmarks on my end and confirmed the similar phenomena though it's also likely that sometimes (literally *some*) genexp outperforms listcomp.
It's very unfortunate that our code becomes *slower* each time we try to improve laziness, I'm agree with your idea: keep it as-is for now.

tests/test_opengraph.py
```diff
@@ -41,3 +41,214 @@ def test_open_graph_to_pattern() -> None:
...
+
+    og, mapping_complete = og_1.compose(og_1, mapping)
+
+    expected_graph: nx.Graph[int]
+    expected_graph = nx.Graph([(1, 2), (100, 200)])
+    assert nx.is_isomorphic(og.inside, expected_graph)
+    assert og.inputs == [1, 100]
+    assert og.outputs == [2, 200]
+
+    outputs_c = {i for i in og.inside.nodes if i not in og.outputs}
```
EarlMilktea, line 80: MEMO: please don't perform `in` check against `list` in production codes (would be acceptable in tests as long as it's not too slow)

EarlMilktea (approved):

Issue #266, opened 2025-04-16
Resolve union types of `float` / `complex` and symbolic stuff


EarlMilktea:
**Describe the feature you'd like**

Refactor simulators or/and update annotations to accurately keep track of types and get rid of unions such as `npt.NDArray[np.complex128 | np.object_]`.

**Additional context**

Union types in return type annotations are in general not good, as users/developers need to further narrow types frequently.

EarlMilktea:
@thierry-martinez 

I have a few questions and suggestions:

- API design of `Expression`
- 
Why not change them to always return `Expression` by removing automatic coercion of `Expression` to numbers?
This will remove union return types and lead to type stability.

```py
@abstractmethod
# def __mul__(self, other: object) -> ExpressionOrFloat:
def __mul__(self, other: object) -> Expression:
```

emlynsg:
Following our discussion, suggest the following breakdown:
- Make patterns generic on parameterised angles `Pattern[Generic[PA]]`
- Create subclasses `NumericPattern` and `ParameterizedPattern` and specialise the methods relevant to those classes (e.g. some simulation will not be in `ParameterizedPattern`, substitution methods won't be in `NumericPattern`).

Potential issues:
- Combining patterns/adding commands to `NumericPattern` could be a bit verbose to protect against.

Issue #167, opened 2024-06-18
Support for arbitrary input states in TensorNetworkBackend


thierry-martinez:
**Is your feature request related to a problem? Please describe.**
`TensorNetworkBackend` currently only supports `|+>` input states.

**Describe the feature you'd like**
We should be able to pass arbitrary states as `input_states` to `TensorNetworkBackend`.

**Additional context**
Support for input states has been added for other back-ends in https://github.com/TeamGraphix/graphix/pull/135.

masa10-f:
Hi! Thanks for you feature request
Let me ask, which do you assume a highly entangled state or a sparsely entangled state as a `input_state`? Do you want to initialize with an arbitrary state vector?

mgarnier59:
Hi Masato,

that will be one of the topics of our discussion in July with @shinich1 and @thierry-martinez. We want to have a better grasp of this backend (why it works, when it works, its relation to more traditional TN backends, how it can be extended, etc...). 

Indeed, we are currently working on refactoring the backends to make the API simpler ans safer. Additionally, since this backend is very specific, the latest features are not implemented there (some of them most likely can't be, like truly arbitrary input states) so it needs special treatment.

masa10-f:
Hi Maxime! Thank you for your response.

> that will be one of the topics of our discussion in July with @shinich1 and @thierry-martinez. We want to have a better grasp of this backend (why it works, when it works, its relation to more traditional TN backends, how it can be extended, etc...).

I'm sorry for not  being able to join the discussion. 
- why it works
  - TN is just a generalization of usual statevec sim. So the result must match
- when it works
  - TN

masa10-f:
Hi Maxime! Thank you for your response.

> that will be one of the topics of our discussion in July with @shinich1 and @thierry-martinez. We want to have a better grasp of this backend (why it works, when it works, its relation to more traditional TN backends, how it can be extended, etc...).

I'm sorry for not  being able to join the discussion. Here are quick answers to your wondering
- why it works
  - TN is just a generalization of usual statevec sim from the perspective of matrix multiplication. The only difference is in the contraction order.
- when it works
  - When we calculate Unitary, because we can skip middle measurement probability calculation and change the contraction order from the SV backend. This is strict especially when a pattern is strongly deterministic because the probability of 0/1 for all the measurement planes must be 50:50. The important point is whether it's permitted to skip probability calculations
- its relation to more traditional TN backends
[...]

PR #312, opened 2025-07-06
Enable type-checking for simulator and backends

thierry-martinez:
This commit adds type annotations to the simulator and backend modules, along with their corresponding tests, enabling full type-checking with `mypy` while preserving existing functionality.

The separation between `TensorNetworkBackend` and backends that operate on a full-state representation, such as `StatevecBackend` and `DensityMatrixBackend`, is now clearer with the introduction of the abstract classes `FullStateBackend` and `DenseState`, which derive from `Backend` and `BackendState`, respectively. `StatevecBackend` and `DensityMatrixBackend` inherit from `FullStateBackend`, while `Statevec` and `DensityMatrix` inherit from `DenseState`. Note that the class hierarchy of `BackendState` mirrors that of `Backend`.

The `State` class in `base_backend` has been renamed to `BackendState` to avoid a name conflict with the `State` class defined in `graphix.states`, which represents the state of a single qubit.

The classes `Backend[StateT_co]` and `FullStateBackend[DenseStateT_co]` are now parameterized by covariant type variables, allowing subclasses to narrow the type of the `state` field to match their specific state representation. Covariance is sound in this context because the classes are frozen, and it ensures that `Backend[BackendState]` is a supertype of all backend classes.

The type `Data` is now declared in a dedicated module, `sim.data`, to break the import cycle between `base_backend`, `statevec`, and `density_matrix`. The definition of `Data` requires importing the `statevec` and `density_matrix` modules, both of which import `base_backend`; `data` is thus only imported within a type-checking block in `base_backend`.

An update of `graphix-symbolic` is required to implement `__abs__` in `Expression`, which is used in the definition of `get_statevec_norm`. This avoids warnings caused by casting to `float`, which would discard the imaginary part. See https://github.com/TeamGraphix/graphix-symbolic/pull/3.

A type alias `Outcome`, defined as `Literal[0, 1]`, has been introduced to replace the mixed use of `int` and `bool`.

**Related issue:**
This PR continues the work started in #302 and #308, and will unblock #300 by ensuring type safety for the proposed changes.


mgarnier59:
Thanks @thierry-martinez for this huge PR! I can't claim I understand all the technical details but it is good that typing has been implemented for the backends. Plus the structure of the backends and simulators is now clearer and will allow easier extension in the future. 
And thanks for the work on the documentation!

mgarnier59 (approved):

EarlMilktea:
@thierry-martinez and all:

General comment:

As you might expect, my comment is simple: please eliminate `typing.cast` and related unsound stuff (ex. repeated `isinstance` check, `TypeGuard`, etc.).
That said, I admit it is very very cumbersome, or even impossible mainly due to symbolic stuff so we might need to discuss about how we deal with it.
Here are possible directions I came up with:

- do the hard work

I don't recommend this: the major problem is that `numpy` only accepts `np.object_` as non-primitive types.
I cannot imagine simple/safe annotations with only `np.object_` ≒ `object` in my hand.

- move symbolic stuff to `graphix-symbolics` and leave only tiny typed subsets in this repo.

I believe this is someday necessary as long as we need static typing, because symbolics are inherently difficult to annotate.

- keep going for now and resolve it in a new major version (`v0.4`)

I personally prefer this plan.

EarlMilktea:
@thierry-martinez 

My concern could be summarized by this minimal example.

```py

import random

# Type correspondence
# int: Expression
# str: float

def bad_func(x: int) -> int | str:
    """Type-unstable function."""
    return str(x) if random.random() > 0.5 else x
```

The problem is the union type in the return type annotation: each time you call `bad_func`, type diverges and thus we repeatedly need to add some kind of branches or casts to identify/harness types.
This is what I wanted to say by "unsound stuff" leading to maintainability problems.

### My comments on symbolics (not directly related to this PR)

The origin of the problem is, I think, the design using `Expression` anywhere in the package.

To my understanding, if we were to bravely remove all the simulation components from `graphix`, measurement angles could be plain numbers or strings (symbol names) with NO special operations defined on them.
As there is no operation, no type instability as well.

As for numeric stuff (ex. `ops.py`) and simulators, my understanding is that we are suffering from type instability just because we are wrongly devoting efforts to reuse current code base at the same time with typing support, even though it is more or less impossible due to type-unstable nature of symbolic stuff (if possible we must have type-stable `sympy` or similar now, excellent).
Defining wrapper methods (ex. `.cos_sin`) also does not solve this problem because `@typing.overload` hell is inevitable as long as we pursue type stability.

Based on the above observations, my suggestion is the following:

1. Completely separate simulators from other components (mostly done in current code base)
2. Use small classes in the non-simulator components instead of huge monolithic `Expression`, for example something like this (or directly use `Fraction` / `str` ?)

```py
@dataclasses.dataclass(frozen=True)
class Fixed:
    """Fixed angle represented as a fraction."""

    value: Fraction = Fraction()


@dataclasses.dataclass(frozen=True)
class Variable:
    """Variable angle."""

    symbol: str

# Carefully-chosen method implementations (substitution etc.)
```

This change will greatly improve type instabilities in the non-simulator section because all the measurement angles are limited to the above two classes and no complicated operations are permitted.

3. In the simulator components, clearly separate symbolic/numeric control flows since we encounter type instabilities immediately after mixing them.
4. Finally, move all the isolated symbolic stuff to `graphix-symbolic` .

thierry-martinez:
I believe your concerns are related to the open issue #266, which this PR does not intend to address. The goal of this PR series is to make most of the codebase well-typed while preserving existing functionalities. It does not aim to make the API more user-friendly or more powerful, but rather to establish a safer foundation for future refactoring. You can think of this as a preliminary step toward a more ambitious refactoring that will address #266 and other contributions. The idea is to resolve the current blockage, where contributions to fundamental parts of Graphix are on hold due to a lack of confidence in the safety of the codebase.

shinich1:
@thierry-martinez thanks for this! I believe docs need updates. 

@EarlMilktea 's concerns are noted. But as @EarlMilktea acknowledges these should be at a separate place and it's good to move forward with this PR.

graphix/graphsim.py
```diff
@@ -455,7 +457,7 @@ def measure_z(self, node: int, choice: int = 0) -> int:
         if choice:
             for i in self.neighbors(node):
                 self.flip_sign(i)
-        result = choice if not isolated else int(self.nodes[node]["sign"])
+        result = choice if not isolated else 1 if self.nodes[node]["sign"] else 0
```
EarlMilktea: Chaining `if` twice is not readable.

graphix/sim/__init__.py
```diff
@@ -1 +1,10 @@
 """Simulation backends."""
+
+from __future__ import annotations
+
+from graphix.sim.base_backend import Backend, BackendState, StateT_co
+from graphix.sim.data import Data
+from graphix.sim.density_matrix import DensityMatrix
+from graphix.sim.statevec import Statevec
+
+__all__ = ["Backend", "BackendState", "Data", "DensityMatrix", "StateT_co", "Statevec"]
```
EarlMilktea: Please do not export `ShapeT_co` as users are unlikely to use it by default.

graphix/sim/base_backend.py
```diff
@@ -14,18 +22,234 @@
...
+
+    Raises
+    ------
+    TypeError
+        If `mat` has an unsupported dtype.
+    """
+    if mat.dtype == np.object_:
+        mat_o = mat.astype(np.object_, copy=False)
+        # mypy doesn't accept object dtype here
+        return np.linalg.eig(mat_o)  # type: ignore[arg-type]
```
EarlMilktea: Could you explain why it's safe to call `np.linalg.eig` here?

graphix/sim/base_backend.py
```diff
@@ -127,15 +354,169 @@ def swap(self, i: int, j: int) -> None:
...
+    --------
+    :class:`DenseState`, :class:`MBQCTensorNet`, :class:`Statevec`, :class:`DensityMatrix`
+    """
+
+    def apply_channel(self, channel: KrausChannel, qargs: Sequence[int]) -> None:
+        """Apply channel to the state."""
+        _ = self  # silence PLR6301
+        _ = channel  # silence ARG002
+        _ = qargs
+        raise NoiseNotSupportedError
```
EarlMilktea: Please do not define dummy method without `abc.abstractmethods` decolator.

graphix/sim/base_backend.py
```diff
@@ -2,10 +2,18 @@
```
EarlMilktea, line 2: General remark:

Please do not use `type: ignore` unless the source of the problem is *outside* this package (ex. typeshed stub is broken).

graphix/sim/base_backend.py
```diff
@@ -152,25 +533,41 @@ def _op_mat_from_result(vec: tuple[float, float, float], result: bool, symbolic:
...
+    if symbolic:
+        op_mat_symbolic: npt.NDArray[np.object_] = np.eye(2, dtype=np.object_) / 2
+        for i, t in enumerate(vec):
+            op_mat_symbolic += sign * t * Clifford(i + 1).matrix / 2
+        return op_mat_symbolic
+    op_mat_complex: npt.NDArray[np.complex128] = np.eye(2, dtype=np.complex128) / 2
+    x = vec[0]
+    y = vec[1]
+    z = vec[2]
+    if not isinstance(x, SupportsFloat) or not isinstance(y, SupportsFloat) or not isinstance(z, SupportsFloat):
```
EarlMilktea, line 515: Let's use `not all(...)` or `any(...)` against iterator.

graphix/sim/base_backend.py
```diff
@@ -152,25 +533,41 @@ def _op_mat_from_result(vec: tuple[float, float, float], result: bool, symbolic:
...
-    return op_mat
+    if symbolic:
+        op_mat_symbolic: npt.NDArray[np.object_] = np.eye(2, dtype=np.object_) / 2
+        for i, t in enumerate(vec):
+            op_mat_symbolic += sign * t * Clifford(i + 1).matrix / 2
+        return op_mat_symbolic
+    op_mat_complex: npt.NDArray[np.complex128] = np.eye(2, dtype=np.complex128) / 2
+    x = vec[0]
+    y = vec[1]
+    z = vec[2]
```
EarlMilktea: Let's use iterator unpack if the number of elements expected is statically known.

graphix/sim/base_backend.py
```diff
@@ -152,25 +533,41 @@ def _op_mat_from_result(vec: tuple[float, float, float], result: bool, symbolic:
...
+    symbolic: bool = False,
+) -> Outcome:
     """Perform measurement of a qubit."""
     vec = plane.polar(angle)
     if pr_calc:
-        op_mat = _op_mat_from_result(vec, False, symbolic=symbolic)
+        op_mat = _op_mat_from_result(vec, 0, symbolic=symbolic)
         prob_0 = state.expectation_single(op_mat, qubit)
-        result = rng.random() > prob_0
+        result: Outcome = 1 if rng.random() > abs(prob_0) else 0
```
EarlMilktea: Can't we remove `: Outcome` ?

graphix/sim/base_backend.py
```diff
@@ -179,80 +576,220 @@ def perform_measure(
     return result
 
 
-class Backend:
-    """Base class for backends."""
+StateT_co = TypeVar("StateT_co", bound="BackendState", covariant=True)
+
+
+@dataclass(frozen=True)
+class Backend(Generic[StateT_co]):
```
EarlMilktea: We need to carefully check if this `TypeVar` should be covariant or not.
I suspect it shouldn't (unless this class is read-only), as typically done in Python STL annotations.

graphix/sim/base_backend.py
```diff
@@ -179,80 +576,220 @@ def perform_measure(
...
+        node: int
+        measurement: Measurement
+        """
+
+
+DenseStateT_co = TypeVar("DenseStateT_co", bound="DenseState", covariant=True)
+
+
+@dataclass(frozen=True)
+class FullStateBackend(Backend[DenseStateT_co], Generic[DenseStateT_co]):
```
EarlMilktea: Same as above.

graphix/sim/density_matrix.py
```diff
@@ -188,20 +221,22 @@ def expectation_single(self, op: npt.NDArray, i: int) -> complex:
...
 
         st1 = copy.copy(self)
         st1.normalize()
 
-        rho_tensor = st1.rho.reshape((2,) * st1.nqubit * 2)
-        rho_tensor = np.tensordot(op, rho_tensor, axes=[1, i])
-        rho_tensor = np.moveaxis(rho_tensor, 0, i)
+        nqubit = self.nqubit
+        rho_tensor: Matrix = st1.rho.reshape((2,) * nqubit * 2)
+        rho_tensor = tensordot(op, rho_tensor, axes=([1], [loc]))
```
EarlMilktea: Let's use `tuple` for static small arrays as it's type signature gives us/type checkers more information.

graphix/sim/density_matrix.py
```diff
@@ -280,27 +326,31 @@ def ptrace(self, qargs: Collection[int] | int) -> None:
         rho_res = self.rho.reshape((2,) * n * 2)
         # ket, bra indices to trace out
         trace_axes = list(qargs) + [n + qarg for qarg in qargs]
-        rho_res = np.tensordot(
-            np.eye(2**qargs_num).reshape((2,) * qargs_num * 2), rho_res, axes=(list(range(2 * qargs_num)), trace_axes)
-        )
+        op: Matrix = np.eye(2**qargs_num).reshape((2,) * qargs_num * 2).astype(np.complex128)
+        rho_res = tensordot(op, rho_res, axes=(list(range(2 * qargs_num)), trace_axes))
```
EarlMilktea: `list` might be redundant as `range` itself is `Sequence` .

graphix/simulator.py
```diff
@@ -94,12 +95,12 @@ def set_measure_result(self, node: int, result: bool) -> None:
 class DefaultMeasureMethod(MeasureMethod):
     """Default measurement method implementing standard measurement plane/angle update for MBQC."""
 
-    def __init__(self, results=None):
+    def __init__(self, results: dict[int, Outcome] | None = None):
```
EarlMilktea: `Mapping` ?

graphix/sim/statevec.py
```diff
@@ -243,15 +286,17 @@ def remove_qubit(self, qarg: int) -> None:
         norm = _get_statevec_norm(self.psi)
         if isinstance(norm, SupportsFloat):
             assert not np.isclose(norm, 0)
-        psi = self.psi.take(indices=0, axis=qarg)
+        index: list[slice[int] | int] = [slice(None)] * self.psi.ndim
+        index[qarg] = 0
+        psi = self.psi[tuple(index)]
```
EarlMilktea, line 267: `index` must be initialized as `tuple` in the first place for better performance.

graphix/sim/statevec.py
```diff
@@ -191,13 +226,21 @@ def ptrace(self, qargs) -> None:
...
         # NOTE works since only one 1 in the eigenvalues corresponding to the state
         # TODO use np.eigh since rho is Hermitian?
         self.psi = np.reshape(evecs[:, np.argmax(evals)], (2,) * nqubit_after)
 
+    # Note that `@property` must appear before `@override` for pyright
+    @property
+    @override
+    def nqubit(self) -> int:
+        """Return the number of qubits."""
+        return self.psi.shape[0].bit_length() - 1
```
EarlMilktea: `.size`

graphix/sim/statevec.py
```diff
@@ -100,18 +104,25 @@ def __init__(
...
-            tmp_psi = functools.reduce(np.kron, list_of_sv)
+
+            def get_statevector(
+                s: states.State | ExpressionOrSupportsComplex | Iterable[ExpressionOrSupportsComplex],
+            ) -> npt.NDArray[np.complex128]:
+                if not isinstance(s, states.State):
+                    raise TypeError("Data should be an homogeneous sequence of states.")
+                return s.get_statevector()
+
+            list_of_sv: list[npt.NDArray[np.complex128]] = [get_statevector(s) for s in input_list]
```
EarlMilktea: Is it necessary to explicitly annotate this?

graphix/sim/tensornet.py
```diff
@@ -2,225 +2,66 @@
...
+from collections.abc import Iterable, Sequence
 from copy import deepcopy
-from typing import TYPE_CHECKING
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, SupportsComplex
 
 import numpy as np
+import numpy.typing as npt
 import quimb.tensor as qtn
 from quimb.tensor import Tensor, TensorNetwork
```
EarlMilktea, line 16: Seems like `quimb` is untyped? 
If so we could (or should) skip forcibly annotating this file as we cannot avoid plenty many of `type: ignore`.
(Also this should be removed from dependencies in the future as well)

graphix/sim/base_backend.py
```diff
@@ -14,18 +22,234 @@
...
 
+if sys.version_info >= (3, 10):
+    Matrix: TypeAlias = npt.NDArray[np.object_ | np.complex128]
+else:
+    from typing import Union
+
+    Matrix: TypeAlias = npt.NDArray[Union[np.object_, np.complex128]]
+
+
+def tensordot(op: Matrix, psi: Matrix, axes: tuple[int | Sequence[int], int | Sequence[int]]) -> Matrix:
```
EarlMilktea, line 48: General remark: these kinds of wrapper functions must be elided over time, as they infinitely bloat each time we access to new `numpy` features.
I accept this change for now but keep in mind that here we *SURELY* require refactor in the future.

graphix/sim/base_backend.py
```diff
@@ -127,15 +354,169 @@ def swap(self, i: int, j: int) -> None:
...
-    """Base class for backend state."""
+class NoiseNotSupportedError(Exception):
+    """Exception raised when `apply_channel` is called on a backend that does not support noise."""
+
+    def __str__(self) -> str:
+        """Return the error message."""
+        return "This backend does not support noise."
+
+
+class BackendState(ABC):
```
EarlMilktea, line 335: General remark: I personally feel we need to carefully downsize abstract methods in the future so that they are guaranteed to be fast for all the sibling classes.
Currently some features seem to be implemented in an inefficient manner.

graphix/sim/base_backend.py
```diff
@@ -179,80 +576,220 @@ def perform_measure(
...
+        """Apply CZ gate to two connected nodes.
+
+        Parameters
+        ----------
+        edge : tuple (i, j)
+            a pair of node indices
+        """
+
+    @abstractmethod
+    def finalize(self, output_nodes: Iterable[int]) -> None:
```
EarlMilktea, line 701: Is it really necessary? At least it should not be in the abstract base class (mandatory for all the sibling classes).

graphix/sim/density_matrix.py
```diff
@@ -6,31 +6,35 @@
...
+    from graphix.sim.data import Data
 
 
-class DensityMatrix(State):
+class DensityMatrix(DenseState):
     """DensityMatrix object."""
 
+    rho: Matrix
+
     def __init__(
```
EarlMilktea, line 38: Missing return type annotation.
(I wonder why this is not caught by type checkers...)

graphix/sim/density_matrix.py
```diff
@@ -61,40 +65,43 @@ def __init__(
...
+            if len(input_list) != 0 and isinstance(input_list[0], Iterable):
+
+                def get_row(
+                    item: Iterable[ExpressionOrSupportsComplex] | State | Expression | SupportsComplex,
+                ) -> list[ExpressionOrSupportsComplex]:
+                    if isinstance(item, Iterable):
+                        return list(item)
+                    raise TypeError("Every row of a matrix should be iterable.")
+
+                input_matrix: list[list[ExpressionOrSupportsComplex]] = list(map(get_row, input_list))
```
EarlMilktea: Let's use listcomp rather than `map`.

graphix/sim/density_matrix.py
```diff
@@ -61,40 +65,43 @@ def __init__(
...
+                if isinstance(input_matrix[0][0], (Expression, SupportsComplex, SupportsFloat)):
```
EarlMilktea: Isn't it guaranteed by annotations?

graphix/sim/density_matrix.py
```diff
@@ -61,40 +65,43 @@ def __init__(
...
+                    self.rho = np.array(input_matrix)
+                    if not lv.is_qubitop(self.rho):
+                        raise ValueError("Cannot interpret the provided density matrix as a qubit operator.")
+                    check_size_consistency(self.rho)
+                    if self.rho.dtype != "O":
```
EarlMilktea: Let's use class style (`np.object_`).

graphix/sim/statevec.py
```diff
@@ -44,9 +40,11 @@ def __init__(self, **kwargs) -> None:
 )
 
 
-class Statevec(State):
+class Statevec(DenseState):
     """Statevector object."""
 
+    psi: Matrix
+
     def __init__(
```
EarlMilktea, line 48: Missing return type annotation.

graphix/sim/statevec.py
```diff
@@ -369,22 +426,27 @@ def xreplace(self, assignment: Mapping[Parameter, ExpressionOrSupportsFloat]) ->
...
+    """MBQC simulator with statevector method."""
+
+    state: Statevec = dataclasses.field(init=False, default_factory=lambda: Statevec(nqubit=0))
+
+
+def _get_statevec_norm_symbolic(psi: npt.NDArray[np.object_]) -> Expression:
     """Return norm of the state."""
-    return np.sqrt(np.sum(psi.flatten().conj() * psi.flatten()))
+    flat = psi.flatten()
+    return cast("Expression", np.abs(np.sqrt(np.sum(flat.conj() * flat))))
```
EarlMilktea: Please do not use `typing.cast`.

graphix/sim/statevec.py
```diff
@@ -369,22 +426,27 @@ def xreplace(self, assignment: Mapping[Parameter, ExpressionOrSupportsFloat]) ->
...
-
-    Data = Union[
-        states.State,
-        Statevec,
-        Iterable[states.State],
-        Iterable[ExpressionOrSupportsComplex],
-    ]
+def _get_statevec_norm_numeric(psi: npt.NDArray[np.complex128]) -> float:
+    flat = psi.flatten()
+    return float(np.abs(np.sqrt(np.sum(flat.conj() * flat))))
```
EarlMilktea: General remark:

`math.xxx` is preferable if arguments are guaranteed to be scalars, as `np.xxx` are full of `Any` in return types.

graphix/simulator.py
```diff
@@ -46,7 +47,7 @@ def measure(self, backend: Backend, cmd, noise_model=None) -> None:
         self.set_measure_result(cmd.node, result)
 
     @abc.abstractmethod
-    def get_measurement_description(self, cmd: BaseM) -> Measurement:
+    def get_measurement_description(self, cmd: command.BaseM) -> Measurement:
```
EarlMilktea: Please do not introduce unnecessary diffs.

graphix/graphsim.py
thierry-martinez: Fixed in e0e1a7c7 (introduced a dedicated coercion function `outcome`).

graphix/sim/__init__.py
thierry-martinez: Fixed in 558d0132.

graphix/sim/base_backend.py
thierry-martinez: I suggest that you test your suggestions before submitting them.
I added this comment in commit e1fa51eb:
```python
    # mypy requires each of x, y, and z to be tested explicitly for it to infer
    # that they are instances of `SupportsFloat`.
    # In particular, using a loop or comprehension like
    # `not all(isinstance(v, SupportsFloat) for v in (x, y, z))` is not supported.
```

graphix/sim/base_backend.py
thierry-martinez: Fixed in e1fa51e.

graphix/sim/base_backend.py
thierry-martinez: Fixed using the dedicated `outcome` coercion in e0e1a7c7.

graphix/sim/density_matrix.py
thierry-martinez: I'm sorry, but I don't understand what you want me to do here. Could you provide an example?

graphix/sim/density_matrix.py
thierry-martinez: Indeed, fixed in 7ba231f0. Note, however, that the use of `list(range(.))` was not a change introduced in this PR.

graphix/simulator.py
thierry-martinez: Fixed in a7acac2a, with this note in the docstring:
```python
        If a mapping is provided, it is treated as read-only. Measurements
        performed during simulation are stored in `self.results`, which is a copy
        of the given mapping. The original `results` mapping is not modified.
```

graphix/sim/base_backend.py
EarlMilktea: Understood why you did so but we still not appropriate.
Let's use `TypeGuard` here: it is not zero-cost but still better than performing redundant list unpack and re-initialization.

```py
def _narrow(lst: list[Any]) -> TypeGuard[list[SupportsFloat]]:
    return all(isinstance(x, SupportsFloat) for x in lst)
```

graphix/sim/statevec.py
thierry-martinez: Fixed in e131f5f1.

graphix/sim/statevec.py
thierry-martinez: Indeed not, thanks. Fixed in 84d8ab14.

graphix/sim/density_matrix.py
thierry-martinez: Fixed in 8aff7bf0.

graphix/sim/statevec.py
thierry-martinez: Fixed in 8aff7bf0.

graphix/sim/density_matrix.py
thierry-martinez: Fixed in dc0e2559.

graphix/sim/density_matrix.py
thierry-martinez: Indeed, yes, thanks. Fixed in c88f456d.

graphix/sim/density_matrix.py
thierry-martinez: Fixed in a454fd4e.

graphix/sim/statevec.py
thierry-martinez: Fixed in dbf70f47 (remove use of `abs` entirely).

graphix/sim/statevec.py
thierry-martinez: Fixed in dbf70f47.

graphix/simulator.py
thierry-martinez: Fixed in 5adb8bab.

graphix/sim/density_matrix.py
EarlMilktea: Oh my bad. Sorry...

graphix/sim/density_matrix.py
EarlMilktea: Like this!

```py
rho_tensor = tensordot(op, rho_tensor, axes=((1,), (loc,)))
```

graphix/sim/base_backend.py
thierry-martinez: This is not an abstract method, but a method that fails in the default case.
I added this note to the docstring in e458e6e2.

```python
        The default implementation of this method raises
        `NoiseNotSupportedError`, indicating that the backend does not
        support noise. Backends that support noise (e.g.,
        `DensityMatrixBackend`) override this method to implement
        the effect of noise.
```

graphix/sim/base_backend.py
thierry-martinez: This method is the only point in the interface for the simulator to convey the order of `output_nodes` in the pattern to the back-end. Currently, the `TensorNetworkBackend` ignores it, but this is a bug: https://github.com/TeamGraphix/graphix/issues/323. In the general case, backends should take the order of `output_nodes` into account when producing their resulting state.

graphix/sim/base_backend.py
thierry-martinez: I think that such an API change is out of the scope of this PR. I invite you to file a dedicated issue, where you could point precisely which methods you would like to remove. However, the current interface is precisely the methods that are called by the simulator to run a pattern. In the commit 19b3b9bb, I added the following comment in the docstring:
```python
    The interface expected from a backend includes the following methods:
    - `add_nodes`: which executes `N` commands.
    - `apply_channel`: used for noisy simulations.
      The class `Backend` provides a default implementation that
      raises `NoiseNotSupportedError`, indicating that the backend
      does not support noise. Backends that support noise (e.g.,
      `DensityMatrixBackend`) override this method to implement the
      effect of noise.
    - `apply_clifford`: executes `C` commands.
    - `correct_byproduct`: executes `X` and `Z` commands.
    - `entangle_nodes`: executes `E` commands.
    - `finalize`: called at the end of pattern simulation to convey
      the order of output nodes.
    - `measure`: executes `M` commands.
```

graphix/sim/base_backend.py
thierry-martinez: The covariance is explained in the message introducing this PR, and in the docstring of this class. In particular, the following note appears in the [docstring](https://github.com/thierry-martinez/graphix/blob/19b3b9bb8fe747fddabe2345f8162f37ec6422e6/graphix/sim/base_backend.py#L634):
`This covariance is sound because backends are frozen dataclasses; thus, the type of state cannot be changed after instantiation.`.

Note that `mypy` partly checks the validity of the covariance annotation: for instance, `mypy` will refuse to remove the following [annotation](https://github.com/thierry-martinez/graphix/blob/19b3b9bb8fe747fddabe2345f8162f37ec6422e6/graphix/sim/base_backend.py#L657) (the comments are part of the committed code):
```python
    # `init=False` is required because `state` cannot appear in a contravariant position
    # (specifically, as a parameter of `__init__`) since `StateT_co` is covariant.
    state: StateT_co = dataclasses.field(init=False)
```

graphix/sim/base_backend.py
thierry-martinez: Same answer as above.

graphix/sim/tensornet.py
thierry-martinez: I propose to keep this as it is, as ignoring type errors for calls to `quimb` doesn't prevent other kinds of errors from being caught in `tensornet.py`, and I think it is still beneficial to have such static type checking. In the long run, we should refactor the tensor network backend (there are at least the following related issues: #167, #323), but even if we have to use `quimb` because there would be no better alternatives, I think a good plan would be to write a type stub file ourselves.

graphix/sim/base_backend.py
thierry-martinez: I agree with this remark, and I invite you to point out the occurrences where you think this is not the case.

graphix/sim/statevec.py
thierry-martinez: Could you suggest an implementation that would initialize index as a tuple in the first place?
I tried my best and obtained the following benchmark results, which show that the tuple approach is twice as slow as using lists:

```python
from timeit import timeit

import numpy as np


ndim = 8
psi = np.array(range(1 << ndim)).reshape((2,) * ndim)


def with_list():
    index: list[slice[int] | int] = [slice(None)] * psi.ndim
    index[2] = 0
    x = psi[tuple(index)]
    index[2] = 1
    y = psi[tuple(index)]
    return x + y


def without_list():
    index = tuple(0 if i == 2 else slice(None) for i in range(psi.ndim))
    x = psi[index]
    index = tuple(1 if i == 2 else slice(None) for i in range(psi.ndim))
    y = psi[index]
    return x + y


print(timeit(with_list))  # 1.3093764169607311
print(timeit(without_list))  # 2.683633374981582
```

graphix/sim/base_backend.py
EarlMilktea: I have two comments on this thread.

## On default implementation

Do you know we can call `super()` even for abstract methods?

```py
class C(ABC):
    @abc.abstractmethod
    def f(self) -> None:
        print("f")


class D(C):
    @typing.override
    def f(self) -> None:
        super().f()


d = D()
d.f()  # Safe
```

If you want to keep this method in the base class, let's add `@abc.abstractmethod` decorator here and instruct sibling classes to call `super()` explicitly if default implementation is sufficient.
This is very common design frequently appearing in OOP (for example see effective C++).

## On design itself

I believe we should not define non-virtual throwing methods in base class.
This approach ultimately leads to something strange like this:

```py
class object:
    # Not all the Python objects are iterable, but just for sibling classes,
    # add this dummy method.
    def __iter__(self) -> Never:
        raise RuntimeError
```

Obviously, the correct way here is to leave `__iter__` undefined.

graphix/sim/base_backend.py
thierry-martinez: Nice catch, thanks! It wasn't safe. I also noticed that `eig` is no longer used in the code base, so I removed the function `eig` in commit 85a6131f.

graphix/sim/base_backend.py
EarlMilktea: OK I understood. Thanks for explanation!

graphix/simulator.py
EarlMilktea: I personally think `Mapping` variable cannot be modified in the first place unless copied, but anyway detailed documentation is always good. Thanks!

graphix/sim/statevec.py
EarlMilktea: I today suddenly realized we need to also care about max memory usages to compare listcomp with genexp in a fair way. Let me report the result later.

graphix/sim/tensornet.py
EarlMilktea: >  write a type stub file ourselves

Looks good. Anyway we should somehow deal with untyped dependencies.

graphix/sim/base_backend.py
EarlMilktea: I see. So the source of problem here is the inappropriate name `finalize` .

graphix/sim/statevec.py
EarlMilktea: (CC: @matulni )

I performed some tests and again confirmed that

- genexp is in general slower than listcomp (typically x2~3)
- genexp is in general memory-friendly (constant vs. linear order)

so we should give up either speed or memory consumption.
My preference is genexp, as the latter is `O(L)` vs. `O(1)` while the former is `O(L)` vs. `O(L)` .
How do you think about it?

thierry-martinez: Honestly, I think it is a typical case of [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality): we are discussing here and elsewhere the use of tuples instead of lists for very small data structures that just contain some indices in cold code paths, where our bottlenecks are on big matrix computations occurring just after in the same functions, so we can't even say that there will be use cases when these considerations will be relevant. I think if we could invest even a small fraction of the time we spend discussing these useless changes on relevant parts of the code, we could make Graphix even better.

The truth is that I could just make the change you suggest to stop the discussion, because even if it would make these operations twice as slow, it would not change anything for Graphix since these operations are negligible, as is their memory consumption. But there is a more general problem: it feels like we’re spending disproportionate time on a micro-optimization that likely has no real impact.

If we go back to the technical subject, you could share your benchmark. I completed the benchmark I posted in my previous comment to measure the memory consumption. The conclusion is that the version using lists is not only substantially faster, but also uses substantially less memory.

Perhaps we can agree to avoid micro-optimizing cold paths unless there's profiling data suggesting a real-world benefit.

```python
import tracemalloc
from timeit import timeit
from typing import Callable

import numpy as np


ndim = 8
psi = np.array(range(1 << ndim)).reshape((2,) * ndim)


def with_list() -> None:
    index: list[slice[int] | int] = [slice(None)] * psi.ndim
    index[2] = 0
    x = psi[tuple(index)]
    index[2] = 1
    y = psi[tuple(index)]
    return x + y


def without_list() -> None:
    index = tuple(0 if i == 2 else slice(None) for i in range(psi.ndim))
    x = psi[index]
    index = tuple(1 if i == 2 else slice(None) for i in range(psi.ndim))
    y = psi[index]
    return x + y


def benchmark(f: Callable[[], None]) -> None:
    tracemalloc.start()
    print(f"- timing: {timeit(f)}s")
    current, peak = tracemalloc.get_traced_memory()
    tracemalloc.stop()
    print(f"- memory: {current / 10**6:.3f} MB / {peak / 10**6:.3f} MB")


print("List:")
benchmark(with_list)
print("Without list:")
benchmark(without_list)
```

```
List:
- timing: 3.373721667099744s
- memory: 0.001 MB / 0.041 MB
Without list:
- timing: 10.382660415954888s
- memory: 0.208 MB / 0.214 MB
```

EarlMilktea: I'm agree with you that we don't have to waste our precious time on such non-critical parts.
Then let's do this:

- I'll approve your code here.
- Let's always use genexp whenever possible, solely because `ruff` sometimes suggests listcomp -> genexp transitions but not the opposite. This convention prevents us from wasting time on micro-optimizations.

graphix/sim/base_backend.py
```diff
@@ -2,30 +2,223 @@
...
+    """
+    if a.dtype == np.complex128 and b.dtype == np.complex128:
+        a_c = a.astype(np.complex128, copy=False)
+        b_c = b.astype(np.complex128, copy=False)
+        return complex(np.vdot(a_c, b_c))
+
+    if a.dtype == np.object_ and b.dtype == np.object_:
+        a_o = a.astype(np.object_, copy=False)
+        b_o = b.astype(np.object_, copy=False)
+        return np.vdot(a_o, b_o)  # type: ignore[no-any-return]
```
EarlMilktea: We could define a function that asserts

1. Given value is not an array
2. Given value is of type `Expression`

rather than using `type: ignore` .

graphix/sim/base_backend.py
```diff
@@ -179,80 +547,241 @@ def perform_measure(
...
 
-    @property
-    def node_index(self) -> NodeIndex:
-        """Return the node index table of the backend."""
-        return self.__node_index
+        Parameters
+        ----------
+            qargs : list of ints. Target qubits
+        """
+        _ = self  # silence PLC0105
```
EarlMilktea: Same as above (should be abstract)

graphix/sim/base_backend.py
```diff
@@ -179,80 +547,241 @@ def perform_measure(
...
+        """Perform measurement of a node and trace out the qubit.
+
+        Parameters
+        ----------
+        node: int
+        measurement: Measurement
+        """
+
+
+DenseStateT_co = TypeVar("DenseStateT_co", bound="DenseState", covariant=True)
```
EarlMilktea: Let's make it module-private: `_DenseStateT_co` .

graphix/sim/statevec.py
```diff
@@ -243,15 +262,17 @@ def remove_qubit(self, qarg: int) -> None:
         norm = _get_statevec_norm(self.psi)
         if isinstance(norm, SupportsFloat):
             assert not np.isclose(norm, 0)
-        psi = self.psi.take(indices=0, axis=qarg)
+        index: list[slice[int] | int] = [slice(None)] * self.psi.ndim
+        index[qarg] = 0
+        psi = self.psi[tuple(index)]
```
EarlMilktea, line 267:
```suggestion
        psi = self.psi[*index]
```

graphix/sim/statevec.py
```diff
@@ -243,15 +262,17 @@ def remove_qubit(self, qarg: int) -> None:
...
         norm = _get_statevec_norm(psi)
-        self.psi = (
-            psi
-            if not isinstance(norm, SupportsFloat) or not np.isclose(norm, 0)
-            else self.psi.take(indices=1, axis=qarg)
-        )
+        if isinstance(norm, SupportsFloat) and math.isclose(norm, 0):
+            index[qarg] = 1
+            psi = self.psi[tuple(index)]
```
EarlMilktea, line 271:
```suggestion
            psi = self.psi[*index]
```

graphix/sim/statevec.py
```diff
@@ -369,22 +402,29 @@ def xreplace(self, assignment: Mapping[Parameter, ExpressionOrSupportsFloat]) ->
...
+    """MBQC simulator with statevector method."""
+
+    state: Statevec = dataclasses.field(init=False, default_factory=lambda: Statevec(nqubit=0))
+
+
+def _get_statevec_norm_symbolic(psi: npt.NDArray[np.object_]) -> Expression:
     """Return norm of the state."""
-    return np.sqrt(np.sum(psi.flatten().conj() * psi.flatten()))
+    flat = psi.flatten()
+    return np.sqrt(np.sum(flat.conj() * flat))  # type: ignore[no-any-return]
```
EarlMilktea: Please perform type check explicitly.

graphix/sim/tensornet.py
```diff
@@ -268,11 +104,11 @@ def get_open_tensor_from_index(self, index):
             index = str(index)
         assert isinstance(index, str)
         tags = [index, "Open"]
-        tid = next(iter(self._get_tids_from_tags(tags, which="all")))
+        tid = next(iter(self._get_tids_from_tags(tags, which="all")))  # type: ignore[no-untyped-call]
         tensor = self.tensor_map[tid]
-        return tensor.data
+        return tensor.data  # type: ignore[no-any-return]
```
EarlMilktea: I typically call `np.asarray` where it should be `npt.NDArray` but type checker does not recognize it.

graphix/sim/tensornet.py
```diff
@@ -318,38 +158,50 @@ def evolve_single(self, index, arr, label="U"):
...
         """
-        if not isinstance(states, list):
-            states = [states] * len(indices)
-        for i, ind in enumerate(indices):
-            self.add_qubit(ind, state=states[i])
+        if isinstance(states, str):
+            states_iter: list[PrepareState] = [states] * len(indices)
+        else:
+            states_list = list(states)
+            if isinstance(states_list[0], SupportsComplex):
```
EarlMilktea: I suppose calling `isinstance` only against one element is not a good idea. Let's narrow the type of the sequence directly.

graphix/sim/tensornet.py
```diff
@@ -729,4 +798,22 @@ def outer_product(vectors):
...
+    raise ValueError("Unsupported states")
+
+
+def _check_state(
+    item: Iterable[Expression | SupportsComplex] | State | Expression | SupportsComplex,
+) -> State | Statevec | Iterable[SupportsComplex]:
+    if isinstance(item, (State, Statevec)):
+        return item
+    if isinstance(item, Iterable):
+        return list(map(_check_complex, item))
```
EarlMilktea: Let's use listcomp.

graphix/sim/tensornet.py
```diff
@@ -697,24 +557,233 @@ def _get_decomposed_cz():
...
+
+        Parameters
+        ----------
+        cmd : list
+            Byproduct command
+            i.e. ['X' or 'Z', node, signal_domain]
+        measure_method : MeasureMethod
+            The measure method to use
+        """
+        if np.mod(sum(measure_method.get_measure_result(j) for j in cmd.domain), 2) == 1:
```
EarlMilktea: Let's use `% 2` rather than `numpy` counterpart.

graphix/sim/tensornet.py
```diff
@@ -697,24 +557,233 @@ def _get_decomposed_cz():
...
+    _decomposed_cz: list[npt.NDArray[np.complex128]] = dataclasses.field(init=False)
+    _isolated_nodes: list[int] = dataclasses.field(init=False)
+
+    def __post_init__(self) -> None:
+        """Construct a tensor network backend."""
+        if self.input_state != BasicStates.PLUS:
+            msg = "TensorNetworkBackend currently only supports BasicStates.PLUS as input state."
+            raise NotImplementedError(msg)
+        # object.__setattr__ is used here to initialize frozen fields
+        object.__setattr__(self, "output_nodes", self.pattern.output_nodes)
```
EarlMilktea: Please don't do this. If you need lazy initialization you can never use `frozen=True` as it's not frozen at all.

graphix/sim/tensornet.py
```diff
@@ -697,24 +557,233 @@ def _get_decomposed_cz():
...
+        object.__setattr__(self, "results", deepcopy(self.pattern.results))
+        if self.graph_prep in {"parallel", "sequential"}:
+            graph_prep = self.graph_prep
+        elif self.graph_prep == "opt":
+            graph_prep = "parallel"
+            print(f"graph preparation strategy '{graph_prep}' is deprecated and will be replaced by 'parallel'")
```
EarlMilktea: Please do not call `print` in library code.

graphix/sim/tensornet.py
```diff
@@ -628,34 +488,34 @@ def evolve(self, operator, qubit_indices, decompose=True, **kwagrs):
...
             [str(index) for index in node_indices],
         )
         if decompose:  # decompose tensor into Matrix Product Operator(MPO)
             tensors = []
-            bond_inds = {0: None}
+            bond_inds: dict[int, str | None] = {0: None}
             for i in range(len(node_indices) - 1):
                 bond_inds[i + 1] = gen_str()
-                left_inds = [new_ind_list[i], old_ind_list[i]]
+                left_inds: list[str | None] = [new_ind_list[i], old_ind_list[i]]
```
EarlMilktea: Maybe tuple?

graphix/sim/tensornet.py
```diff
@@ -602,15 +462,15 @@ def expectation_value(self, op, qubit_indices, output_node_indices=None, **kwagr
...
             tn_cp_right.tensor_map[tid_right].retag({"Open": "Close"})
-        tn_cp_left.add([op_ts, tn_cp_right])
+        tn_cp_left.add([op_ts, tn_cp_right])  # type: ignore[no-untyped-call]
 
         # contraction
-        tn_cp_left = tn_cp_left.full_simplify("ADCR")
-        exp_val = tn_cp_left.contract(output_inds=[], **kwagrs)
-        norm = self.get_norm(**kwagrs)
+        tn_cp_left = tn_cp_left.full_simplify("ADCR")  # type: ignore[no-untyped-call]
+        exp_val: float = tn_cp_left.contract(output_inds=[], optimize=optimize)
```
EarlMilktea: I prefer calling `float` function to asserting by `: float` to ensure that the type is surely of type `float` .
(Many other similar occurrence)

EarlMilktea (changes requested):

graphix/sim/tensornet.py
```diff
@@ -484,22 +343,22 @@ def get_basis_coefficient(self, basis, normalize=True, indices=None, **kwagrs):
...
+            tid = next(iter(tn._get_tids_from_inds(old_ind)))  # type: ignore[no-untyped-call]
             tn.tensor_map[tid].retag({"Open": "Close"})
-            tn.add_tensor(tensor)
+            tn.add_tensor(tensor)  # type: ignore[no-untyped-call]
 
         # contraction
-        tn_simplified = tn.full_simplify("ADCR")
-        coef = tn_simplified.contract(output_inds=[], **kwagrs)
+        tn_simplified = tn.full_simplify("ADCR")  # type: ignore[no-untyped-call]
+        coef: complex = tn_simplified.contract(output_inds=[])
```
EarlMilktea: Let's call `complex` (many other occurrences)

graphix/sim/tensornet.py
```diff
@@ -628,34 +488,34 @@ def evolve(self, operator, qubit_indices, decompose=True, **kwagrs):
...
             self._dangling[str(node_indices[i])] = new_ind_list[i]
 
-        ts = Tensor(
+        ts: Tensor | TensorNetwork = Tensor(  # type: ignore[no-untyped-call]
             operator,
             new_ind_list + old_ind_list,
             [str(index) for index in node_indices],
         )
         if decompose:  # decompose tensor into Matrix Product Operator(MPO)
             tensors = []
```
EarlMilktea: Let's add a type annotation here for safety.

graphix/sim/tensornet.py
```diff
@@ -697,24 +557,233 @@ def _get_decomposed_cz():
...
+        """Make entanglement between nodes specified by edge.
+
+        Parameters
+        ----------
+        edge : tuple of int
+            edge specifies two target nodes of the CZ gate.
+        """
+        if self.graph_prep == "sequential":
+            old_inds = [self.state._dangling[str(node)] for node in edge]
+            tids = self.state._get_tids_from_inds(old_inds, which="any")  # type: ignore[no-untyped-call]
```
EarlMilktea: Isn't there a public counterpart of this protected method?

graphix/sim/tensornet.py
```diff
@@ -2,225 +2,66 @@
 
 from __future__ import annotations
```
EarlMilktea, line 3: I recommend you call `pyright` with the new configuration (added unknown type reporting) specifically for this file.

EarlMilktea (changes requested):

graphix/sim/density_matrix.py
thierry-martinez: Fixed in 80580c82.

graphix/sim/base_backend.py
thierry-martinez: I think the correct version of `_narrow` would be:
```python
def _narrow(
    lst: tuple[ExpressionOrFloat, ExpressionOrFloat, ExpressionOrFloat],
) -> TypeGuard[tuple[SupportsFloat, SupportsFloat, SupportsFloat]]:
    return all(isinstance(x, SupportsFloat) for x in lst)
```
because `vec` is of type `tuple[ExpressionOrFloat, ExpressionOrFloat, ExpressionOrFloat]`, which is not a subtype of `list[Any]`.

But why do you think that your proposed change improves the code?
- Will it improve readability? You propose to replace a straightforward piece of code by an intermediate function which rely on subtle typing rules of `TypeGuard` that are not obvious and are used nowhere else in the code base.
- Will it improve performance ? Let’s be honest, we are talking about a 3-length tuple here. But even if we did care, just "profile, don't speculate", as the motto says. Spoiler: I did the benchmark (see at the end of the comment), but again, performance is clearly not a concern in this context.
- Will it improve type safety? As you probably know, correctness of `TypeGuard` are not checked by `mypy`. That means the implementation of `_narrow` could be just wrong (for instance, simply `return True`), and `mypy` will not complain.

```python
from typing import SupportsFloat, TypeGuard
from timeit import timeit

ExpressionOrFloat = str | float


def _narrow_vec_float(
    lst: tuple[ExpressionOrFloat, ExpressionOrFloat, ExpressionOrFloat],
) -> TypeGuard[tuple[SupportsFloat, SupportsFloat, SupportsFloat]]:
    return all(isinstance(x, SupportsFloat) for x in lst)


vec: tuple[ExpressionOrFloat, ExpressionOrFloat, ExpressionOrFloat] = (1.0, 2.0, 3.0)


def using_narrow() -> float:
    assert _narrow_vec_float(vec)
    s = 0.0
    for u in vec:
        s += float(u)
    return s


def using_unpacking() -> float:
    x, y, z = vec
    if (
        not isinstance(x, SupportsFloat)
        or not isinstance(y, SupportsFloat)
        or not isinstance(z, SupportsFloat)
    ):
        assert False
    float_vec = x, y, z
    s = 0.0
    for u in float_vec:
        s += float(u)
    return s


print(f"{timeit(using_narrow)=}")
print(f"{timeit(using_unpacking)=}")
```
```
timeit(using_narrow)=0.8231092079986411
timeit(using_unpacking)=0.6818534999983967
```

graphix/sim/base_backend.py
thierry-martinez: Could you clarify the specific issue you're trying to address and what alternative design you'd suggest?

The current approach defines a method in the base class with a default implementation (raising `NoiseNotSupportedError`), which covers the common case of backends not supporting noise. Derived classes only override it if they do support noise.

This seems straightforward and avoids boilerplate in subclasses. Is there a particular drawback you're seeing with this pattern?

graphix/sim/base_backend.py
thierry-martinez: I answered here: https://github.com/TeamGraphix/graphix/pull/312#discussion_r2225297572

graphix/sim/base_backend.py
thierry-martinez: Why do you think that your proposed change improves the code? In particular, there are legitimate use of `StateT_co` if a user wants to write code that is generic over every backend (as is typically done in `simulator.py`), and it is not hard to imagine legitimate uses of `DenseStateT_co` is a user wants to write code that is generic over every dense backend.

graphix/sim/statevec.py
thierry-martinez: `ruff check` says:
```SyntaxError: Cannot use star expression in index on Python 3.9 (syntax was added in Python 3.11)```

File: graphix/sim/statevec.py (hunk 33 above)
Comment (thierry-martinez) [reply graphix/sim/statevec.py:271]: See my previous answer: https://github.com/TeamGraphix/graphix/pull/312#discussion_r2225363276

File: graphix/sim/tensornet.py (hunk 41 above)
Comment (thierry-martinez) [reply graphix/sim/tensornet.py]: I would prefer not to change too much of the current logic of the tensor network backend. This is a PR about type-checking, not intended to refactor the backend. I agree that backends in general, and this one in particular, would benefit from refactoring, but that is explicitly not the goal of this PR. In this particular case, two lines later, there is a call to `left_inds.append` in a section of the code I'd prefer to leave unchanged in this PR, because it is not necessary for type-checking.

File: graphix/sim/tensornet.py (hunk 45 above)
Comment (thierry-martinez) [reply graphix/sim/tensornet.py]: This question would be pertinent to refactoring the tensor network backend, which is not the scope of this PR.

File: graphix/sim/tensornet.py (hunk 46 above)
Comment (thierry-martinez) [reply graphix/sim/tensornet.py:3]: The new configuration has made this type of reporting non-blocking for the CI, on purpose.

File: graphix/sim/tensornet.py (hunk 45 above)
Comment (EarlMilktea) [reply graphix/sim/tensornet.py]: In a sense yes, but we still should not write bad codes. You're responsible for ensuring code quality of newly-added lines.

File: graphix/sim/tensornet.py (hunk 41 above)
Comment (EarlMilktea) [reply graphix/sim/tensornet.py]: > two lines later, there is a call to `left_inds.append`

Oops, sorry. You're right!

File: graphix/sim/base_backend.py (hunk 4 above)
Comment (EarlMilktea) [reply graphix/sim/base_backend.py]: > you're seeing with this pattern

When user see a non-virtual method defined in a class, it is very natural to expect that it returns cleanly but it's not the case for this method: this is why.
If you still want to define such a method return annotation must be `Never`, but this time you cannot extend the method in derived classes.

```py
class C:
    def f(self) -> Never:
        raise NotImplementedError


class D(C):
    # Return type "None" of "f" incompatible with return type "Never" in supertype "C"
    def f(self) -> None:
        print(3)
```

graphix/sim/tensornet.py
EarlMilktea: My question here is that, whether we can improve the newly-added line by substituting it by public APIs. It's not related to refactoring.

graphix/sim/base_backend.py
EarlMilktea: One strong reason for this suggestion is https://peps.python.org/pep-0695/ introduced in Python 3.12: we cannot share typevars anymore with the new syntax. This is why I suggest keeping it private if possible (for `StateT_co` you don't have to do it immediately as it's used in other places).

graphix/sim/statevec.py
EarlMilktea: I didn't know that. Thanks!

graphix/sim/base_backend.py
EarlMilktea: OK let's keep your code here. Either will do as long as you pick the correct approach depending on situations (ex. not doing the same thing for $10^9$-element tuples).

graphix/sim/tensornet.py
```diff
@@ -418,7 +270,7 @@ def set_graph_state(self, nodes, edges):
         .. seealso:: :meth:`~graphix.sim.tensornet.TensorNetworkBackend.__init__()`
         """
         ind_dict = {}
```
EarlMilktea:
```suggestion
        ind_dict: dict[int, list[str]] = {}
```

graphix/sim/base_backend.py
thierry-martinez: Fixed in e8f5a694.

graphix/sim/statevec.py
thierry-martinez: Fixed in cdde47ef.

graphix/sim/tensornet.py
thierry-martinez: I finally added a type stub for quimb here a77e1ece (that I had planned to do for the long term, but it solves a lot of your comments, so I motivated me to add it right now).

graphix/sim/tensornet.py
thierry-martinez: I added the following comment and a check to handle the empty case in 7b543609:
```python
            # `states` is of type `PrepareState`, a type alias for
            # `str | npt.NDArray[np.complex128]`. To distinguish
            # between the two cases, we just need to check whether
            # an element is a character or a complex number.
```
The type of `PrepareState` is probably something that will change when we will refactor the tensor network backend.

graphix/sim/tensornet.py
thierry-martinez: Fixed in c8e67beb.

graphix/sim/tensornet.py
EarlMilktea: Excellent. I'm also personally interested in how we can place stubs of external packages in our repo.

graphix/sim/tensornet.py
thierry-martinez: Fixed in 57dbb308. Note, however, that the use of `np.mod` was not a change introduced in this PR.

graphix/sim/tensornet.py
thierry-martinez: Replaced by a warning in a711ea77. Note, however, that the `print` statement was not introduced in this PR.

graphix/sim/tensornet.py
```diff
@@ -729,4 +807,22 @@ def outer_product(vectors):
...
+
+def _check_complex(
+    item: Iterable[Expression | SupportsComplex] | State | Expression | SupportsComplex,
+) -> SupportsComplex:
+    if isinstance(item, SupportsComplex):
+        return item
+    raise ValueError("Unsupported states")
+
+
+def _check_state(
```
EarlMilktea: Probably unused?

graphix/sim/tensornet.py
thierry-martinez: Solved with the addition of a type stub for `quimb` (see https://github.com/TeamGraphix/graphix/pull/312#discussion_r2225782787).

graphix/sim/tensornet.py
thierry-martinez: Solved with the addition of a type stub for `quimb` (see https://github.com/TeamGraphix/graphix/pull/312#discussion_r2225782787).

graphix/sim/tensornet.py
thierry-martinez: Solved with the addition of a type stub for `quimb` (see https://github.com/TeamGraphix/graphix/pull/312#discussion_r2225782787).

graphix/sim/tensornet.py
EarlMilktea: Sorry as usual...

graphix/sim/tensornet.py
thierry-martinez: It is not properly a newly-added line (see https://github.com/TeamGraphix/graphix/blame/73bac1ff2e91f3e7273d7fb603466f38f8cf49a8/graphix/sim/tensornet.py#L125): this is marked as a change because `diff` is not clever enough to follow the lines that have been preserved.

graphix/sim/tensornet.py
thierry-martinez: I consider it fairly standard to initialize fields of frozen dataclasses in `__post_init__` using `object.__setattr__`. However, as an alternative, I proposed initializing the fields in `_AbstractTensorNetworkBackend` as shown here: 81c62ccc.

graphix/sim/tensornet.py
thierry-martinez: Fixed in 4e33978b.

graphix/sim/tensornet.py
thierry-martinez: Fixed in be848ebb.

graphix/sim/base_backend.py
thierry-martinez: I don’t think PEP 695 deprecates the use of `TypeVar` itself, but I agree that the modern generic syntax introduced there encourages using a different approach for type variables. To future-proof the code and maintain symmetry, I made both `_StateT_co` and `_DenseStateT_co` private in commit d88ca878, and I redefine them locally when needed.

thierry-martinez:
Thank you, @EarlMilktea. I think I have addressed all the comments I could address. Would you check if the PR is ready for merging now?

graphix/sim/tensornet.py
EarlMilktea: Terribly sorry...

thierry-martinez:
It seems that all points have been addressed. Shall we proceed with the merge?

EarlMilktea:
No. You have still unresolved threads. Also give me a time to catch up with recent changes.

graphix/sim/base_backend.py
thierry-martinez: We are not in such a case here. As specified in the method’s documentation, the method succeeds if the backend supports noise, and raises `NoiseNotSupportedError` if it does not, which is the documented and expected behavior.

This is not a placeholder method nor a method that always raises; rather, raising is part of its normal control flow, depending on the backend’s capabilities.

This design allows the simulator to treat all backends uniformly: the method is defined in the base class, and backends that do not support noise simply raise `NoiseNotSupportedError` when the user attempts to run a noisy simulation. This makes explicit support detection unnecessary in most use cases.

EarlMilktea: OK let's go ahead with your design. Before resolving this thread please replace `_ = ...` with `noqa` comments since compiler explorer says the assignments are not optimized out:

```py
# Python source
def f(x: int) -> None:
    _ = x
```

```
# Disassemble
1           RESUME                   0

2           LOAD_FAST                0 (x)
            STORE_FAST               1 (_)
            RETURN_CONST             0 (None)
```

graphix/sim/base_backend.py
EarlMilktea: Checked. Looks good. `no-any-return` for `@` would be acceptable as `numpy` 's  `__matmul__` is truly a monster.

EarlMilktea:
@thierry-martinez Almost there. Could you finally resolve the remaining two comments for redundant assigments?

graphix/sim/base_backend.py
EarlMilktea: Please also use `noqa` here.

thierry-martinez: Fixed in c535d4e4.

graphix/sim/base_backend.py
thierry-martinez: Fixed in f30da166.

thierry-martinez:
> @thierry-martinez Almost there. Could you finally resolve the remaining two comments for redundant assigments?

Done. Thanks!

PR #317, opened 2025-07-16
Report unknown types

EarlMilktea:

This PR enables several optional `pyright` rules to report unknown types (ex. unannotated collection literals, complex expressions)
Note that these rules cannot be treated as errors as there are plenty many of false positives.


thierry-martinez (approved):
LGTM! Thanks!

PR #318, opened 2025-07-16
Fix #316: pin package version for `mypy` and `types-networkx`

### LINKED ISSUE #316: [Bug]: Set more precise package versions in `requirements-dev.txt` for cleaner PR reviews
To ensure smoother PR review process, we already fix the version of `ruff` in `requirements-dev.txt` and rely on Dependabot to keep it up to date.

However, as seen in #309 and #315, CI can still break due to other package updates. `requirements-dev.txt` doesn't need to pin all package versions immediately, but as #309 and #315 illustrate, we may want to specify an exact version for packages like `mypy` and `types-networkx`, and let Dependabot handle updates.

thierry-martinez:
This commit pins the package version for `mypy==1.17.0` and `types-networkx==3.5.0.20250701`. It's a continuation of PR #243 where we pinned the version of `ruff`.

Pinning these package should make the PR review process smoother.

- In #315, CI broke due to a regression introduced in `mypy==1.17`, even though the PR was only about updating `ruff`.

- In #309, CI broke due to stricter typing in newer versions of `types-networkx`, which required changes in parts of the code not touched by the PR.

By pinning these packages and relying on Dependabot to open dedicated PRs for updates:

- We avoid unpredicatable CI failures caused by unexpected version updates.

- We prevent unrelated changes from being added to PRs just because a package update occurred during the review process.

- We get dedicated PRs to manage and test updates.

- We are notified when updates break the CI, even if no new code has been pushed since the update was released.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -13,26 +13,22 @@ def install_pytest(session: Session) -> None:
     session.install("pytest", "pytest-mock", "psutil")
 
 
-def run_pytest(session: Session) -> None:
-    """Run pytest."""
-    session.run("pytest", "--doctest-modules")
-
-
 @nox.session(python=["3.9", "3.10", "3.11", "3.12", "3.13"])
 def tests_minimal(session: Session) -> None:
     """Run the test suite with minimal dependencies."""
     session.install("-e", ".")
     install_pytest(session)
-    run_pytest(session)
+    # We cannot run `pytest --doctest-modules` here, since some tests
+    # involve optional dependencies, like pyzx.
+    session.run("pytest")
 
-    # Note that recent types-networkx versions don't support Python 3.9
 
-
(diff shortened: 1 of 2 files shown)


EarlMilktea:
It seems that `types-networkx` should pinned as `==3.5.0.20250712` .

EarlMilktea:
@thierry-martinez 

> do not support Python 3.9

I suppose, we could just migrate type-checking python version to `3.10` (fix CI config) and use `<=` version pin for `types-newtorkx` for the time being.
This is not a clean solution especially for 3.9 users, but would be acceptable as `3.9` will soon be EoL and it's very unlikely that next release will be done before the day.
Don't you think it's a bit overkill to adjust our CI checks just for one package?

EarlMilktea (approved):

PR #320, opened 2025-07-21
New feature: `Pattern.compose

matulni:
Introduce new method of `:class: graphix.pattern.Pattern` allowing to combine two patterns. 

Heading:
```python
graphix.pattern.Pattern.compose(self, other: Pattern, mapping: Mapping[int, int], preserve_mapping: bool = False) -> tuple[Pattern, dict[int, int]]:
```

Let's denote $(I_j, O_j, V_j, S_j)$ the ordered set of inputs and outputs, the computational space and the sequence of commands of pattern $P_j$, respectively, with $j =1 (2)$ for pattern `self`(`other`). Let's denote $P$ the resulting pattern with ($I, O, V, S$).
Let's denote $K, U$ the sets of `keys` and `values`of `mapping`.

Expected behavior:

1. $K \subseteq V_2$.
2. For a pair $(k, v) \in (K, U)$ 
	1. $v$ can always satisfy $v \notin V_1$, thereby allowing a custom relabelling. 
	2. $U \cap V_1 \setminus O_1 = \emptyset$ (there can't be measured qubits in `values`).  If $v \in O_1$, then $k \in I_2$, otherwise an error is raised.
3. $i \in V_2 \cap K^c$ are relabelled in ascending order.
4. The sequence of the resulting pattern is simply $S = S_2 S_1$, where qubits in $S_2$ are relabelled according to `mapping` (commands of $P_2$ are added _after_ commands of $P_1$).   
5. $I = I_1 \cup (I_2 \setminus M_2)$ where $M_2 = \{k \in I_2 \cap K | k \rightarrow v, v \in M_1 \}$ is the set of merged inputs.
6. $O = (O_1 \setminus M_1) \cup O_2$ where $M_1 = O_1 \cap U$ is the set of merged outputs.
7. Input (and, respectively, output) nodes in the returned pattern have the order of the pattern `self` followed by those of the pattern `other`. Merged nodes are removed.
8. If `preserve_mapping = True` and $|M_1| = |I_2| = |O_2|$, then the outputs of the returned pattern are the outputs of pattern `self`, where the nth merged output is replaced by the output of pattern `other` corresponding to its nth input instead.
For example, let's consider:
```
G1 = 0--1--2--3
I1 = O1 = [0, 1, 2, 3]

G2 =
0  1
|  |
2--3
|  |
4  5

I2 = [0, 1]
O2 = [4, 5]
```
If `mapping = {0: 1, 1: 2, 4: 100, 5:101}`, then 

```
- O(preserve_mapping = False) = [0, 3, 100, 101]
- O(preserve_mapping = True) = [0, 100, 101, 3]
```
If `mapping = {0: 2, 1: 1, 4: 100, 5:101}`, then 

```
- O(preserve_mapping = False) = [0, 3, 100, 101]
- O(preserve_mapping = True) = [0, 101, 100, 3]
```

Notes:

- (2.2) could be made more general by allowing $k \in V_2$. If $k \notin I_2$, then we would remove the command $N_k$ from $S_2$ in the resulting pattern. However this behavior could be obtained by setting $k$ as in input in $P_2$, when defining the pattern. The design choice above is follows closer the theory (e.g., Danos et al., 2007).
- The feature `:func: graphix.opengraph.OpenGraph.compose`(see #310 ) is more general. Namely:
	- It allows for merging measured nodes provided that the measuring plane is the same. This is not possible here, because we would need to suppress the measurement command in $P_1$, which might result in commands belonging to $S_1$ depending on an outcome not yet measured.
	- It allows for merging $o \in O_2$ and $i \in I_1$. This behavior in the pattern composition is not consistent with (4.)

Additional modifications:

- Added `Pattern.results` comparison to `Pattern.__eq__` method.


tests/test_pattern.py
```diff
@@ -427,6 +427,253 @@ def test_standardize_domains_and_clifford(self, fx_bg: PCG64, jumps: int) -> Non
...
+        circuit_12.h(0)
+        circuit_12.rz(0, alpha)
+        p12 = circuit_12.transpile().pattern
+
+        assert p == p12
+
+    # Equivalence between pattern a circuit composition after standarization
+    # FAILING because of standardize
+    @pytest.mark.skip(reason="Pattern standarization does not support Clifford commands before entanglement commands.")
+    def test_compose_6(self, fx_rng: Generator) -> None:
```
thierry-martinez:
```suggestion
    # Equivalence between pattern a circuit composition after standardization
    # FAILING because of standardize
    @pytest.mark.skip(reason="Pattern standardization does not support Clifford commands before entanglement commands.")
    def test_compose_6(self, fx_rng: Generator) -> None:
```

tests/test_pattern.py
```diff
@@ -427,6 +427,253 @@ def test_standardize_domains_and_clifford(self, fx_bg: PCG64, jumps: int) -> Non
...
+        p2 = circuit_2.transpile().pattern  # inputs: [0]
+
+        p, _ = p1.compose(p2, mapping={0: 1, 1: 2, 2: 3})
+
+        circuit_12 = Circuit(1)
+        circuit_12.h(0)
+        circuit_12.rz(0, alpha)
+        p12 = circuit_12.transpile().pattern
+
+        assert p == p12
```
thierry-martinez, line 658: We could even test with random circuits:
```python
    def test_compose_random(self, fx_bg: PCG64, jumps: int) -> None:
        nqubits = 4
        depth = 2
        circuit_1 = rand_circuit(nqubits, depth, rng, use_ccx=True)
        circuit_2 = rand_circuit(nqubits, depth, rng, use_ccx=True)
        circuit = Circuit(width = nqubits, instr=circuit_1.instruction + circuit_2.instruction)
        p1 = circuit_1.transpile().pattern
        p2 = circuit_2.transpile().pattern
        p = circuit.transpile().pattern
        p_compose, _ = p1.compose(p2, mapping={i: o for i, o in zip(p2.input_nodes, p1.output_nodes)}, preserve_mapping=True)
        p.minimize_space()
        p_compose.minimize_space() 
        s = p.simulate_pattern()
        s_compose = p_compose.simulate_pattern()
        assert np.abs(np.dot(s.flatten().conjugate(), s_compose.flatten())) == pytest.approx(1)
```

thierry-martinez (changes requested):
Good! Just a few comments.

tests/test_pattern.py
matulni: Thanks for catching the typo!

tests/test_pattern.py
matulni: Beautiful!

graphix/pattern.py
```diff
@@ -165,6 +165,119 @@ def replace(self, cmds: list[Command], input_nodes: list[int] | None = None) ->
...
+                )
+
+        shift = max(*nodes_p1, *mapping.values()) + 1
+        mapping_sequential = {
+            node: i for i, node in enumerate(sorted(nodes_p2 - mapping.keys()), start=shift)
+        }  # assigns new labels to nodes in other not specified in mapping
+
+        mapping_complete = {**mapping, **mapping_sequential}
+
+        mapped_inputs = [mapping_complete[n] for n in other.input_nodes]
```
EarlMilktea, line 251: Please do not perform `in` check against `list` (there are many other affected lines) as it's slow.

graphix/pattern.py
```diff
@@ -165,6 +165,119 @@ def replace(self, cmds: list[Command], input_nodes: list[int] | None = None) ->
...
+        nodes_p2_lst, _ = other.get_graph()
+        nodes_p2: set[int] = set(nodes_p2_lst) | other.results.keys()
+
+        if not mapping.keys() <= nodes_p2:
+            raise ValueError("Keys of `mapping` must correspond to the nodes of `other`.")
+
+        if len(mapping) != len(set(mapping.values())):
+            raise ValueError("Values of `mapping` contain duplicates.")
+
+        if set(mapping.values()) & nodes_p1 - set(self.__output_nodes):
```
EarlMilktea: General remark: 

`set` has several methods that accept `Iterable` as operands, which enable us to perform set operations without redundant `set` calls.

```py
x: list[int]
y: list[int]

# set allocation x2
set(x) | set(y)

# set allocation x1
set(x).union(y)
```

tests/test_pattern.py
```diff
@@ -427,6 +427,272 @@ def test_standardize_domains_and_clifford(self, fx_bg: PCG64, jumps: int) -> Non
...
+        pc.standardize()  # Problematic because pc contains Cliffords before entanglement operators
+        s_compose_std = pc.simulate_pattern(backend="statevector")
+
+        circuit_3 = Circuit(1)
+        circuit_3.h(0)
+        circuit_3.rz(0, 2 * alpha)
+        p3 = circuit_3.transpile().pattern
+        s_circuit = p3.simulate_pattern(backend="statevector")
+
+        for s in [s_compose, s_compose_std]:
```
EarlMilktea:
```suggestion
        for s in (s_compose, s_compose_std):
```

EarlMilktea (changes requested):

tests/test_pattern.py
matulni: Thanks for the review. Out of curiosity, why using a tuple instead of a list here?

graphix/pattern.py
matulni: Hi thanks for the comment.
I think that here 
```python
(nodes_p1 - set(self.__output_nodes)).intersection(mapping.values())
```
has worse readability while it doesn't provide any significant performance gain, so I'd rather leave as is.

I implemented this suggestion in https://github.com/TeamGraphix/graphix/pull/320/commits/bca159445e2e5707966806e55ceb41752f53de18 for `merged`

graphix/pattern.py
matulni: Note that the order of inputs and outputs matters, so we can't do
```python
mapped_inputs = [mapping_complete[n] for n in set(other.input_nodes)]
```

I implemented this where possible, see https://github.com/TeamGraphix/graphix/pull/320/commits/bca159445e2e5707966806e55ceb41752f53de18

EarlMilktea: Oops sorry, I added this thread to a wrong place...

EarlMilktea: Anyway please keep in mind that `in` for membership test (not iteration) should always be done against `set` or `dict.keys()` .

graphix/pattern.py
EarlMilktea: Yeah exactly (this is why I added `General remark:`). Please use them appropriately when possible :)

tests/test_pattern.py
EarlMilktea: The most significant reason is, `tuple` can be more specifically annotated compared to `list` : has length information, can be inhomogeneous, etc.
There might be performance advantages as well (not sure for Python, but for other languages there is due to stack allocation).

shinich1:
Thanks! I see `test_compose_7` seems to be problematic and needs to be resolved. I believe there was a discussion of turning Clifford commands into other commands (since they're just single-qubit rotations) if cannot be absorbed into measurements. Isn't this good time to make that change (perhaps another PR)?

tests/test_pattern.py
matulni: I see, thanks for the reply!

Note that in this specific case:

1) It's a collection of two elements, so performance differences are minimal.
2) It's not part of production code, so micro-optimizations are less critical.
3) It's homogenous.

I'm not sure if here it makes much sense to change it, but I don't have any strong preference for one or the other, I leave it to your discretion :)

EarlMilktea (approved):

graphix/pattern.py
```diff
@@ -165,6 +165,130 @@ def replace(self, cmds: list[Command], input_nodes: list[int] | None = None) ->
...
```
EarlMilktea: General remark: please don't call `set` (or pure functions in general) against same variables multiple times as it's costly.

graphix/pattern.py
```diff
@@ -165,6 +165,130 @@ def replace(self, cmds: list[Command], input_nodes: list[int] | None = None) ->
...
+            preserve_mapping = False
+
+        if preserve_mapping:
+            io_mapping = {mapping[i]: mapping_complete[o] for i, o in zip(other.input_nodes, other.output_nodes)}
+            outputs = [io_mapping[n] if n in merged else n for n in self.__output_nodes]
+        else:
+            outputs = [n for n in self.__output_nodes if n not in merged] + mapped_outputs
+
+        def update_command(cmd: Command) -> Command:
+            cmd_new = copy.copy(cmd)
```
EarlMilktea, line 274: Please make sure that `copy.copy` is what you really wanted here (rather than `copy.deepcopy`) .

EarlMilktea (changes requested):

thierry-martinez (approved):
LGTM! Approved. Thanks.

graphix/pattern.py
matulni: Shallow copy is enough since the mutable attributes of `cmd_new` are reassigned.
Only exception is `cmd_new.state` if `cmd_new.kind == CommandKind.N" which should not be modified anywhere in the code.
I added a comment to clarify.

graphix/pattern.py
```diff
@@ -169,6 +169,133 @@ def replace(self, cmds: list[Command], input_nodes: list[int] | None = None) ->
...
+        mapping_values_set = set(mapping.values())
+
+        if len(mapping) != len(mapping_values_set):
+            raise ValueError("Values of `mapping` contain duplicates.")
+
+        if mapping_values_set & nodes_p1 - set(self.__output_nodes):
+            raise ValueError("Values of `mapping` must not contain measured nodes of pattern `self`.")
+
+        # Cast to set for improved performance in membership test
+        o1_set = set(self.__output_nodes)
```
EarlMilktea: Please reuse this in the `if` branch in line 225.

PR #321, opened 2025-07-21
Optimal HSZ decompositions

thierry-martinez:
This commit ensures that the HSZ decompositions in the `_db.py` module are optimal with respect to decomposition length. That is, for each Clifford gate, there is no shorter decomposition into H, S, and Z gates than the one provided.

The previous decompositions were suboptimal. For example, the decomposition `(6, 6, 3, 6, 3)` applies two Hadamard gates (denoted by `6`) in succession, which can be simplified.

For reference, here is a function that computes an optimal decomposition for each Clifford gate:

```python
def compute_optimal_hsz_decompositions() -> dict[Clifford, list[Clifford]]:
    """Compute an optimal HSZ decomposition for each Clifford gate."""
    basis = [Clifford.H, Clifford.S, Clifford.Z]
    result = {gate: [gate] for gate in [Clifford.I, *basis]}
    gates_to_expand = basis
    while gates_to_expand:
        new_gates = []
        for gate in gates_to_expand:
            for basic_gate in basis:
                new_gate = gate @ basic_gate
                if new_gate not in result:
                    result[new_gate] = result[gate] + [basic_gate]
                    new_gates.append(new_gate)
        gates_to_expand = new_gates
    return result
```


matulni (approved):
Thanks for posting the function, very elegant!

EarlMilktea (approved):

PR #322, opened 2025-07-21
Fix #314: Add `optimization` module with correct standardization

### LINKED ISSUE #314: [Bug]: `standardize(method="mc")` is broken in presence of `C` commands
The following code demonstrates that `standardize(method="mc")` does not preserve semantics when the pattern contains C commands:

```python
import numpy as np
import math

from graphix import Pattern, command
from graphix.clifford import Clifford

seed = 0
pattern = Pattern(input_nodes=[0, 1, 2])
pattern.add(command.E((0, 2)))
pattern.add(command.E((1, 2)))
pattern.add(command.M(0, Plane.XY, 0.2))
pattern.add(command.M(1, Plane.XY, 0.2))
pattern.add(command.X(2, {0}))
pattern.add(command.Z(2, {1}))
pattern.add(command.C(2, Clifford.S))
state = pattern.simulate_pattern(rng=np.random.default_rng(seed))

pattern.standardize(method="direct")
state2 = pattern.simulate_pattern(rng=np.random.default_rng(seed))
assert math.isclose(np.abs(np.dot(state.psi.conjugate(), state2.psi)), 1)

pattern.standardize(method="mc")
state2 = pattern.simulate_pattern(rng=np.random.default_rng(seed))
assert math.isclose(np.abs(np.dot(state.psi.conjugate(), state2.psi)), 1)  # fails here
```

The final assertion fails, showing that the `"mc"` standardization method produces a pattern that is not equivalent to the original in the presence of C commands. In contrast, the `"direct"` method works correctly.

thierry-martinez:
This commit introduces a new module `optimization.py` containing:

- A functional version of `standardize` that returns a standardized pattern as a new object.  This function now correctly handles patterns with C commands, returning an equivalent standardized pattern or raising an error if no such form exists, thus fixing #314.

- A new function `incorporate_pauli_results` that returns an equivalent pattern where `results` are incorporated into measurement and correction domains. The resulting pattern is suitable for flow analysis. In particular, if a pattern has a flow, it is preserved by `perform_pauli_measurements` after applying `standardize` and `incorporate_pauli_results`.

Standardized patterns now place C commands after X and Z commands. The `flow_from_pattern` functions now fail if the input pattern is not strictly standardized, as checked by
`Pattern.is_standard(strict=True)` (which requires C commands to be at the end).

The `method` parameter of `standardize` is now deprecated.  The method `mc` has been removed, as it did not guarantee the equivalence of the resulting pattern.

These algorithms are placed in a new module to reduce the algorithmic contents of `pattern.py`. For backward compatibility and convenience, the `Pattern` class retains an in-place `standardize` method, which internally calls `optimization.standardize`.

As a result, there is now an intentional import loop between `pattern.py` and `optimization.py`: `Pattern` calls `optimization.standardize`, and `optimization.standardize` instantiates `graphix.Pattern`. To manage this safely, we avoid `from ... import` statements between these modules, except in type-checking blocks, thereby guaranteeing that there is no import-time execution cycle.

### DIFF
--- a/graphix/optimization.py
+++ b/graphix/optimization.py
@@ -1,0 +1,204 @@
+"""Optimization procedures for patterns."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import graphix
+from graphix import command
+from graphix.clifford import Clifford
+from graphix.command import CommandKind
+from graphix.measurements import Domains
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
+    from collections.abc import Set as AbstractSet
+
+    from graphix import Pattern
+
+
+def standardize(pattern: Pattern) -> Pattern:
+    """Return a standardized form to the given pattern.
+
+    A standardized form is an equivalent pattern where the commands
+    appear in the following order: `N`, `E`, `M`, `Z`, `X`, `C`.
+
+    Note that a standardized form does not always exist in presence of
+    `C` commands. For instance, there is no standardized form for the
+    following pattern (written in the right-to-left convention):
+    `E(0, 1) C(0, H) N(1) N(0)`.
+
+    The function raises `NotImplementedError` if there is no
+    standardized form. This behavior can change in the future.
+
+
+    Parameters
+    ----------
+    pattern : Pattern
+        The original pattern.
+
+    Returns
+    -------
+    standardized : Pattern
+        The standardized pattern, if it exists.
+    """
+    n_list = []
+    e_list = []
+    m_list = []
+    c_dict: dict[int, Clifford] = {}
+    z_dict: dict[int, set[command.Node]] = {}
+    x_dict: dict[int, set[command.Node]] = {}
+
+    def add_correction_domain(
+        domain_dict: dict[command.Node, set[command.Node]], node: command.Node, domain: set[command.Node]
+    ) -> None:
+        """Merge a correction domain into ``domain_dict`` for ``node``.
+
+        Parameters
+        ----------
+        domain_dict : dict[int, Command]
+            Mapping from node index to accumulated domain.
+        node : int
+            Target node whose domain should be updated.
+        domain : set[int]
+            Domain to merge with the existing one.
+        """
+        if previous_domain := domain_dict.get(node):
+            previous_domain ^= domain
+        else:
+            domain_dict[node] = domain.copy()
+
+    def commute_clifford(clifford_gate: Clifford, c_dict: dict[int, Clifford], i: int, j: int) -> None:
+        """Commute a Clifford with an entanglement command.
+
+        Parameters
+        ----------
+        clifford_gate : Clifford
+            Clifford gate before the entanglement command
+        c_dict : dict[int, Clifford]
+            Mapping from the node index to accumulated Clifford commands.
+        i : int
+            First node of the entanglement command where the Clifford is applied.
+        j : int
+            Second node of the entanglement command where the Clifford is applied.
+        """
+        if clifford_gate in {Clifford.I, Clifford.Z, Clifford.S, Clifford.SDG}:
+            # Clifford gate commutes with the entanglement command.
+            pass
+        elif clifford_gate in {Clifford.X, Clifford.Y, Clifford(9), Clifford(10)}:
+            # Clifford gate commutes with the entanglement command up to a Z Clifford on the other index.
+            c_dict[j] = Clifford.Z @ c_dict.get(j, Clifford.I)
+        else:
+            # Clifford gate commutes with the entanglement command up to a two-qubit Clifford
+            raise NotImplementedError(
+                f"Pattern contains a Clifford followed by an E command on qubit {i} which only commute up to a two-qubit Clifford. Standarization is not supported."
+            )
+
+    for cmd in pattern:
+        if cmd.kind == CommandKind.N:
+            n_list.append(cmd)
+        elif cmd.kind == CommandKind.E:
+            for side in (0, 1):
+                i, j = cmd.nodes[side], cmd.nodes[1 - side]
+                if clifford_gate := c_dict.get(i):
+                    commute_clifford(clifford_gate, c_dict, i, j)
+                if s_domain := x_dict.get(i):
+                    add_correction_domain(z_dict, j, s_domain)
+            e_list.append(cmd)
+        elif cmd.kind == CommandKind.M:
+            new_cmd = cmd
+            if clifford_gate := c_dict.pop(cmd.node, None):
+                new_cmd = new_cmd.clifford(clifford_gate)
+            if t_domain := z_dict.pop(cmd.node, None):
+                # The original domain should not be mutated
+                new_cmd.t_domain = new_cmd.t_domain ^ t_domain  # noqa: PLR6104
+            if s_domain := x_dict.pop(cmd.node, None):
+                # The original domain should not be mutated
+                new_cmd.s_domain = new_cmd.s_domain ^ s_domain  # noqa: PLR6104
+            m_list.append(new_cmd)
+        # Use of `==` here for mypy
+        elif cmd.kind == CommandKind.X or cmd.kind == CommandKind.Z:  # noqa: PLR1714
+            if cmd.kind == CommandKind.X:
+                s_domain = cmd.domain
+                t_domain = set()
+            else:
+                s_domain = set()
+                t_domain = cmd.domain
+            domains = c_dict.get(cmd.node, Clifford.I).commute_domains(Domains(s_domain, t_domain))
+            if domains.t_domain:
+                add_correction_domain(z_dict, cmd.node, domains.t_domain)
+            if domains.s_domain:
+                add_correction_domain(x_dict, cmd.node, domains.s_domain)
+        elif cmd.kind == CommandKind.C:
+            # Each pattern command is applied by left multiplication: if a clifford `C`
+            # has been already applied to a node, applying a clifford `C'` to the same
+            # node is equivalent to apply `C'C` to a fresh node.
+            c_dict[cmd.node] = cmd.clifford @ c_dict.get(cmd.node, Clifford.I)
+    result = graphix.Pattern(input_nodes=pattern.input_nodes)
+    result.results = pattern.results
+    result.extend(
+        [
+            *n_list,
+            *e_list,
+            *m_list,
+            *(command.Z(node=node, domain=domain) for node, domain in z_dict.items()),
+            *(command.X(node=node, domain=domain) for node, domain in x_dict.items()),
+            *(command.C(node=node, clifford=clifford_gate) for node, clifford_gate in c_dict.items()),
+        ]
+    )
+    return result
+
+
+def _incorporate_pauli_results_in_domain(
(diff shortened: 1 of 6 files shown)


graphix/pattern.py
```diff
@@ -279,111 +279,30 @@ def print_pattern(self, lim: int = 40, target: Container[CommandKind] | None = N
...
-        `N`, `E`, `M`, `C`, `Z`, `X`.
-        """
-        n_list = []
-        e_list = []
-        m_list = []
-        c_dict: dict[int, Clifford] = {}
-        z_dict: dict[int, set[command.Node]] = {}
-        x_dict: dict[int, set[command.Node]] = {}
+        if method is not None:
+            warnings.warn(
```
EarlMilktea: I suppose we can silently eliminate `method` argument here because we've already introduced tons of breaking changes. Let's prioritize simplicity.

graphix/optimization.py
```diff
@@ -0,0 +1,204 @@
...
+        The original pattern.
+
+    Returns
+    -------
+    standardized : Pattern
+        The standardized pattern, if it exists.
+    """
+    n_list = []
+    e_list = []
+    m_list = []
```
EarlMilktea: Please add annotations here (`pyright` is complaining)

graphix/optimization.py
```diff
@@ -0,0 +1,204 @@
...
+    c_dict: dict[int, Clifford] = {}
+    z_dict: dict[int, set[command.Node]] = {}
```
EarlMilktea: Comment: Is there specific reason for prefixing `Node` with `command` while it's not done for other pascal-case symbols?

graphix/optimization.py
```diff
@@ -0,0 +1,204 @@
...
+    result.results = pattern.results
+    result.extend(
+        [
+            *n_list,
+            *e_list,
+            *m_list,
+            *(command.Z(node=node, domain=domain) for node, domain in z_dict.items()),
+            *(command.X(node=node, domain=domain) for node, domain in x_dict.items()),
+            *(command.C(node=node, clifford=clifford_gate) for node, clifford_gate in c_dict.items()),
+        ]
```
EarlMilktea, line 147: I'm again very strongly tempted to use iterator chains here.

graphix/optimization.py
```diff
@@ -0,0 +1,204 @@
...
+                new_cmd = command.M(cmd.node, cmd.plane, cmd.angle, new_s_domain, new_t_domain)
+                if apply_x:
+                    new_cmd = new_cmd.clifford(Clifford.X)
+                if apply_z:
+                    new_cmd = new_cmd.clifford(Clifford.Z)
+                result.add(new_cmd)
+            else:
+                result.add(cmd)
+        # Use == for mypy
+        elif cmd.kind == CommandKind.X or cmd.kind == CommandKind.Z:  # noqa: PLR1714
```
EarlMilktea, line 190: Should we entirely ignore `PLR1714` ? I feel it's not contributing to better DX.

EarlMilktea (changes requested):

graphix/optimization.py
EarlMilktea: Let's call `.extend` 6 times for each operands. It would be the most efficient and concise way.

graphix/optimization.py
```diff
@@ -0,0 +1,204 @@
...
+    result = graphix.Pattern(input_nodes=pattern.input_nodes)
+    result.results = pattern.results
+    result.extend(
+        [
+            *n_list,
+            *e_list,
+            *m_list,
+            *(command.Z(node=node, domain=domain) for node, domain in z_dict.items()),
+            *(command.X(node=node, domain=domain) for node, domain in x_dict.items()),
+            *(command.C(node=node, clifford=clifford_gate) for node, clifford_gate in c_dict.items()),
```
EarlMilktea: (Writing the same content again as I accidentally deleted, sorry!)

Let's call `.extend` 6 times to eliminate unnecessary list collection.

graphix/pattern.py
thierry-martinez: Which breaking changes do you have in mind? I feel that most of the API has been preserved by the currently merged and open PRs.

graphix/optimization.py
thierry-martinez: Do you think such a change would improve code readability or overall efficiency?

graphix/optimization.py
thierry-martinez: Fixed in 0b094525.

graphix/optimization.py
thierry-martinez: Fixed in 0b094525.

graphix/optimization.py
thierry-martinez: Filed a new issue here: https://github.com/TeamGraphix/graphix/issues/324. I think changing `ruff` rules is outside the scope of this PR.

graphix/optimization.py
EarlMilktea: https://docs.astral.sh/ruff/rules/unnecessary-list-cast/

I feel this is nearly against this rule while it's not caught by `ruff` . Let's use genexp coherently, especially here the lengths of iterables can be large.

graphix/pattern.py
EarlMilktea: For example module structures are aggressively changed and no deprecation warnings are emitted at all: it's allowed as `graphix` is still v0. Of course we should pay great attentions after reaching v1.

graphix/optimization.py
EarlMilktea: Actually this rule is the chief reason I'm sticking to genexp: I don't want to waste my time anymore on arguing whether we should use listcomp (micro optimization effective if input length is small) or genexp (`O(1)` memory consumption guaranteed even for long arrays, it's `ruff` 's choice). Please always use genexp if you want faster review.

thierry-martinez: The rule [unnecessary-list-cast (PERF101)](https://docs.astral.sh/ruff/rules/unnecessary-list-cast/#unnecessary-list-cast-perf101) is about unnecessary list casts. It is normal that it is not triggered here, since no list cast is present.

> Please always use genexp if you want faster review.

Why require using a generator expression when it is not appropriate? I think if someone had used a generator expression in code like this, it would have been more appropriate to suggest they change it for the sake of readability.

EarlMilktea: I forgot to mention to another rule: https://docs.astral.sh/ruff/rules/unnecessary-comprehension-in-call/ .

> Many builtin functions accept any iterable, including a generator.
> Constructing a temporary list via list comprehension is unnecessary and wastes memory for large iterables.

While `.extend` is not covered by `ruff` for now, it trivially applies to it.
Your code obviously allocates a temporary list.

Of course I never suggest writing terrible codes just for playing with genexp.
Don't you think it is readable enough?

```py
    result.extend(n_list)
    result.extend(e_list)
    result.extend(m_list)
    result.extend(command.Z(node=node, domain=domain) for node, domain in z_dict.items())
    result.extend(command.X(node=node, domain=domain) for node, domain in x_dict.items())
    result.extend(command.C(node=node, clifford=clifford_gate) for node, clifford_gate in c_dict.items())
```

thierry-martinez: > While `.extend` is not covered by `ruff` for now, it trivially applies to it.

The rule `unnecessary-comprehension-in-call (C419)` applies to list comprehensions, and there is no list comprehension here.

More generally, there are no ruff rules against constructs like `sum([1])` either, but I agree that we can still aim to write cleaner code. See below for a suggestion.

> Your code obviously allocates a temporary list.

Could you clarify why that’s a concern in this context? Allocating a temporary list is not necessarily an issue, and in this case, I don't think it significantly affects performance or memory usage.

As a general note, it’s often more productive to benchmark before making performance-based decisions. Intuition alone can be misleading: Profile, don't speculate! The space complexity of this function is O(n) regardless of whether we use a generator expression or a list, so any performance difference would need to be measured.

> Of course I never suggest writing terrible codes just for playing with genexp. Don't you think it is readable enough?

That’s a fair point! Personally, I find the original version slightly more readable, but that’s subjective. To improve clarity and avoid the intermediate list allocation, I’ve changed the code in commit 5671c507 to:
```python
    result.extend(
        n_list,
        e_list,
        m_list,
        (command.Z(node=node, domain=domain) for node, domain in z_dict.items()),
        (command.X(node=node, domain=domain) for node, domain in x_dict.items()),
        (command.C(node=node, clifford=clifford_gate) for node, clifford_gate in c_dict.items()),
    )
```
To make this work, I adapted the implementation of `extend` so that it behaves more like a built-in function; your mention of the ruff rule regarding built-ins gave me that idea.

EarlMilktea: > a temporary list is not necessarily an issue

Would be so if you consider 200 % memory usage acceptable. Unfortunately I don't.

Basically, I would like to emphasize that I'm not against redundant list collection when it greatly simplifies code / improves UX, for example when we need to store `Iterator` directly in class unless collected.
I don't think this is the case for this thread: actually, my suggestion has exactly the same cyclomatic complexity (20 specifically) with yours while eliding one redundant list collection.

> `O(n)` regardless of whether we use a generator expression or a list

No. For example my suggestion is `O(1)` if newly-added elements fit in preallocated buffers (I'm not sure of the implementation detail of Python `list`, but typically capacity is doubled on reallocation) while yours is always `O(n)` .

> changed the code in commit

Please note that the new API is not compatible with `collections.abc.MutableSequence` : we may need to change the signature once we decide to rely on it.
Aside from STL compatibility I'm content with your code!

```py
class MutableSequence(Sequence[_T]):
    # snipped
    def extend(self, values: Iterable[_T]) -> None: ...
```

thierry-martinez: > No. For example my suggestion is `O(1)` if newly-added elements fit in preallocated buffers (I'm not sure of the implementation detail of Python `list`, but typically capacity is doubled on reallocation) while yours is always `O(n)` .

This is not what I said. It doesn't affect the space complexity of the `standardize` function itself: the loop just above is `O(n)`, and `extend` is also `O(n)`. Improving a single part from `O(n)` to `O(1)` doesn’t change the overall complexity when the function still does linear work elsewhere. Asymptotic complexity needs to be evaluated in context.

graphix/pattern.py
thierry-martinez: Fixed in c2d10e09.

graphix/optimization.py
EarlMilktea: Let's keep going with your updated design for now, as our remarks are both correct (the only difference is what we prioritize) and thus we cannot conclude immediately.
Anyway your new zero-copy design completely looks good to me!

tests/test_optimization.py
```diff
@@ -0,0 +1,85 @@
...
+    nqubits = 3
+    depth = 3
+    circuit = rand_circuit(nqubits, depth, rng)
+    pattern = circuit.transpile().pattern
+    pattern.standardize()
+    pattern.shift_signals()
+    pattern.perform_pauli_measurements()
+    pattern2 = incorporate_pauli_results(pattern)
+    pattern2.standardize()
+    f, _l = gflow_from_pattern(pattern2)
```
EarlMilktea, line 84:
```suggestion
    f, _ = gflow_from_pattern(pattern2)
```

EarlMilktea (approved):

Issue #220, opened 2024-10-10
Avoid `get_` / `set_`


EarlMilktea:
Functions named `get_` / `set_` should be renamed due to following observations:

1. If it is not getter/setter (ex. has extra arguments)

We should not use `get_` / `set_` as it is not a getter/setter.

2. If it is lightweight getter/setter

We should consider `@property` for more user-friendly APIs.

3. If it is costly getter, but can be cached

We should consider `@functools.cached_property`.

4. If it is costly getter/setter that cannot be cached

We should avoid `get_` / `set_` as attribute getter/setter are considered to be lightweight.

PR #332, opened 2025-08-13
Fix #173: introduce `StandardizedPattern` class

### LINKED ISSUE #173: [Bug]: `get_measurement_commands` should not alter the pattern
**Describe the bug**
`get_` method should not normally change itself.
remove the following lines: https://github.com/TeamGraphix/graphix/blob/cef611d3b9555867aac8dfbbbdb21b484de01a45/graphix/pattern.py#L954C2-L956C2

thierry-martinez:
This commit introduces the `StandardizedPattern` class in the optimization module. The code from the `standardize` function has been moved into this class, allowing access to the decomposed parts of the standardized pattern through class attributes.

This class enables Pauli preprocessing to avoid relying on `Pattern.get_measurements_commands` and `Pattern.get_graph`, since the relevant commands are now extracted during the standardization process.

Previously, the method checked `Pattern.get_measurements_commands` to determine whether the pattern was standardized (which was always true in the context of Pauli preprocessing), and standardized it if necessary. It then extracted the portion of the pattern consisting of consecutive M commands. Although this optimization stopped just after the last M command, checking for standardization still required iterating over the entire pattern.

This method has now been replaced by `extract_measurement_commands`. This renaming follows the recommendation in #220 to avoid `get_` and `set_` method names. The new method produces an iterator over the M commands, regardless of whether the pattern is standardized.

### DIFF
--- a/graphix/optimization.py
+++ b/graphix/optimization.py
@@ -42,112 +42,175 @@ def standardize(pattern: Pattern) -> Pattern:
     standardized : Pattern
         The standardized pattern, if it exists.
     """
-    n_list: list[command.N] = []
-    e_list: list[command.E] = []
-    m_list: list[command.M] = []
-    c_dict: dict[int, Clifford] = {}
-    z_dict: dict[int, set[Node]] = {}
-    x_dict: dict[int, set[Node]] = {}
+    return StandardizedPattern(pattern).to_pattern()
 
-    def add_correction_domain(domain_dict: dict[Node, set[Node]], node: Node, domain: set[Node]) -> None:
-        """Merge a correction domain into ``domain_dict`` for ``node``.
 
-        Parameters
-        ----------
-        domain_dict : dict[int, Command]
-            Mapping from node index to accumulated domain.
-        node : int
-            Target node whose domain should be updated.
-        domain : set[int]
-            Domain to merge with the existing one.
+class StandardizedPattern:
+    """Pattern in standardized form.
+
+    Use the method :meth:`to_pattern()` to get the standardized pattern.
+    Note that the attribute :attr:`pattern` contains the original pattern.
+
+    Attributes
+    ----------
+    pattern: Pattern
+        The original pattern.
+    n_list: list[command.N]
+        The N commands.
+    e_list: list[command.E]
+        The E commands.
+    m_list: list[command.M]
+        The M commands.
+    c_dict: dict[int, Clifford]
+        Mapping associating Clifford corrections to some nodes.
+    z_dict: dict[int, set[Node]]
+        Mapping associating Z-domains to some nodes.
+    x_dict: dict[int, set[Node]]
+        Mapping associating X-domains to some nodes.
+    """
+
+    pattern: Pattern
+    n_list: list[command.N]
+    e_list: list[command.E]
+    m_list: list[command.M]
+    c_dict: dict[int, Clifford]
+    z_dict: dict[int, set[Node]]
+    x_dict: dict[int, set[Node]]
+
+    def __init__(self, pattern: Pattern) -> None:
+        """Compute the standardized form of the given pattern."""
+        s_domain: set[Node]
+        t_domain: set[Node]
+        s_domain_opt: set[Node] | None
+        t_domain_opt: set[Node] | None
+
+        self.pattern = pattern
+        self.n_list = []
+        self.e_list = []
+        self.m_list = []
+        self.c_dict = {}
+        self.z_dict = {}
+        self.x_dict = {}
+
+        for cmd in pattern:
+            if cmd.kind == CommandKind.N:
+                self.n_list.append(cmd)
+            elif cmd.kind == CommandKind.E:
+                for side in (0, 1):
+                    i, j = cmd.nodes[side], cmd.nodes[1 - side]
+                    if clifford_gate := self.c_dict.get(i):
+                        _commute_clifford(clifford_gate, self.c_dict, i, j)
+                    if s_domain_opt := self.x_dict.get(i):
+                        _add_correction_domain(self.z_dict, j, s_domain_opt)
+                self.e_list.append(cmd)
+            elif cmd.kind == CommandKind.M:
+                new_cmd = cmd
+                if clifford_gate := self.c_dict.pop(cmd.node, None):
+                    new_cmd = new_cmd.clifford(clifford_gate)
+                if t_domain_opt := self.z_dict.pop(cmd.node, None):
+                    # The original domain should not be mutated
+                    new_cmd.t_domain = new_cmd.t_domain ^ t_domain_opt  # noqa: PLR6104
+                if s_domain_opt := self.x_dict.pop(cmd.node, None):
+                    # The original domain should not be mutated
+                    new_cmd.s_domain = new_cmd.s_domain ^ s_domain_opt  # noqa: PLR6104
+                self.m_list.append(new_cmd)
+            # Use of `==` here for mypy
+            elif cmd.kind == CommandKind.X or cmd.kind == CommandKind.Z:  # noqa: PLR1714
+                if cmd.kind == CommandKind.X:
+                    s_domain = cmd.domain
+                    t_domain = set()
+                else:
+                    s_domain = set()
+                    t_domain = cmd.domain
+                domains = self.c_dict.get(cmd.node, Clifford.I).commute_domains(Domains(s_domain, t_domain))
+                if domains.t_domain:
+                    _add_correction_domain(self.z_dict, cmd.node, domains.t_domain)
+                if domains.s_domain:
+                    _add_correction_domain(self.x_dict, cmd.node, domains.s_domain)
+            elif cmd.kind == CommandKind.C:
+                # Each pattern command is applied by left multiplication: if a clifford `C`
+                # has been already applied to a node, applying a clifford `C'` to the same
+                # node is equivalent to apply `C'C` to a fresh node.
+                self.c_dict[cmd.node] = cmd.clifford @ self.c_dict.get(cmd.node, Clifford.I)
+
+    def extract_graph(self) -> tuple[list[int], list[tuple[int, int]]]:
+        """Return the list of nodes and edges from the command sequence, extracted from 'N' and 'E' commands.
+
+        Returns
+        -------
+        node_list : list
+            list of node indices.
+        edge_list : list
+            list of tuples (i,j) specifying edges
         """
-        if previous_domain := domain_dict.get(node):
-            previous_domain ^= domain
-        else:
-            domain_dict[node] = domain.copy()
+        # We rely on the fact that pattern.input_nodes returns a copy
+        node_list: list[int]
+        edge_list: list[tuple[int, int]]
+        node_list, edge_list = self.pattern.input_nodes, []
+        for cmd_n in self.n_list:
+            node_list.append(cmd_n.node)
+        for cmd_e in self.e_list:
+            edge_list.append(cmd_e.nodes)
+        return node_list, edge_list
 
-    def commute_clifford(clifford_gate: Clifford, c_dict: dict[int, Clifford], i: int, j: int) -> None:
-        """Commute a Clifford with an entanglement command.
+    def to_pattern(self) -> Pattern:
+        """Return the standardized pattern."""
+        result = graphix.pattern.Pattern(input_nodes=self.pattern.input_nodes)
+        result.results = self.pattern.results
+        result.extend(
+            self.n_list,
+            self.e_list,
+            self.m_list,
+            (command.Z(node=node, domain=domain) for node, domain in self.z_dict.items()),
+            (command.X(node=node, domain=domain) for node, domain in self.x_dict.items()),
+            (command.C(node=node, clifford=clifford_gate) for node, clifford_gate in self.c_dict.items()),
+        )
+        return result
 
-        Parameters
-        ----------
-        clifford_gate : Clifford
-            Clifford gate before the entanglement command
-        c_dict : dict[int, Clifford]
-            Mapping from the node index to accumulated Clifford commands.
-        i : int
-            First node of the entanglement command where the Clifford is applied.
-        j : int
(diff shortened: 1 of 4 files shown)


graphix/pattern.py
```diff
@@ -1734,7 +1717,9 @@ def measure_pauli(pattern: Pattern, leave_input: bool, copy: bool = False) -> Pa
     return pat
 
 
-def pauli_nodes(pattern: Pattern, leave_input: bool) -> tuple[list[tuple[command.M, PauliMeasurement]], set[int]]:
+def pauli_nodes(
+    pattern: optimization.StandardizedPattern, leave_input: bool
+) -> tuple[list[tuple[command.M, PauliMeasurement]], set[int]]:
     """Return the list of measurement commands that are in Pauli bases and that are not dependent on any non-Pauli measurements.
```
matulni, line 1710: `pattern : graphix.Pattern object` -> `pattern : optimization.StandardizedPattern` in docstring
(GitHub doesn't allow me to comment that specific line).

graphix/optimization.py
```diff
@@ -42,112 +42,175 @@ def standardize(pattern: Pattern) -> Pattern:
...
-            )
+        # We rely on the fact that pattern.input_nodes returns a copy
+        node_list: list[int]
+        edge_list: list[tuple[int, int]]
+        node_list, edge_list = self.pattern.input_nodes, []
+        for cmd_n in self.n_list:
+            node_list.append(cmd_n.node)
+        for cmd_e in self.e_list:
+            edge_list.append(cmd_e.nodes)
+        return node_list, edge_list
```
matulni: Would it make sense to return an `nx.Graph` object ? (Since we express graphs in this form in other parts of the code, e.g., in `graphix.OpenGraph`).

There's one subtle difference, however:

If we consider the pattern: `p = Pattern(input_nodes=[], output_nodes=[0,1], cmds=[N(0), N(1), E((0, 1)), E((0, 1)), E((0, 1))])`  (a bit silly, but valid), we have that

`p.get_graph()` -> `([0, 1], [(0, 1), (0, 1), (0, 1)])`, even after standarization.

However, `list(nx.Graph( [(0, 1), (0, 1), (0, 1)]).edges))` returns `[(0, 1)]`, so behavior is not entirely equivalent.

By the way, our standarization routine does not contemplate that $E_{i,j} E_{i,j} = 1$ , although I think that, strictly speaking, this is not a rewrite rule in the Measurement Calculus.

thierry-martinez: Indeed, the handling of double-entanglement is a long-standing bug that has not been properly reported yet. I just filled this issue: https://github.com/TeamGraphix/graphix/issues/339

I made some changes in the PR so that `extract_graph` now returns an `nx.Graph` object.
The implementation of `Pattern.extract_graph` computes the graph without standardizing the pattern first (but it correctly handles that $CZ ∘ CZ = I$). There is a new method `Pattern.extract_nodes` if we don't care about edges.

graphix/optimization.py
```diff
@@ -42,112 +45,178 @@ def standardize(pattern: Pattern) -> Pattern:
...
+        result.results = self.pattern.results
+        result.extend(
+            self.n_list,
+            (command.E((u, v)) for u, v in self.e_set),
+            self.m_list,
+            (command.Z(node=node, domain=domain) for node, domain in self.z_dict.items()),
+            (command.X(node=node, domain=domain) for node, domain in self.x_dict.items()),
+            (command.C(node=node, clifford=clifford_gate) for node, clifford_gate in self.c_dict.items()),
+        )
+        return result
```
matulni: I find `result` name a bit confusing, since `Pattern` has a `results` attribute. How about `pattern_std` ?
Also, we have to initialize the output nodes as well, right?

```suggestion
        pattern_std = graphix.pattern.Pattern(input_nodes=self.pattern.input_nodes, output_nodes=self.pattern.output_nodes)
        pattern_std.results = self.pattern.results
        pattern_std.extend(
            self.n_list,
            (command.E((u, v)) for u, v in self.e_set),
            self.m_list,
            (command.Z(node=node, domain=domain) for node, domain in self.z_dict.items()),
            (command.X(node=node, domain=domain) for node, domain in self.x_dict.items()),
            (command.C(node=node, clifford=clifford_gate) for node, clifford_gate in self.c_dict.items()),
        )
        return pattern_std
```

graphix/pattern.py
```diff
@@ -212,9 +212,9 @@ def compose(
         - Input (and, respectively, output) nodes in the returned pattern have the order of the pattern `self` followed by those of the pattern `other`. Merged nodes are removed.
         - If `preserve_mapping = True` and :math:`|M_1| = |I_2| = |O_2|`, then the outputs of the returned pattern are the outputs of pattern `self`, where the nth merged output is replaced by the output of pattern `other` corresponding to its nth input instead.
         """
-        nodes_p1_lst, _ = self.get_graph()
+        nodes_p1_lst = self.extract_nodes()
```
matulni: `:func: self.extract_nodes` returns a set, right?

```suggestion
        nodes_p1 = self.extract_nodes()
```

graphix/pattern.py
```diff
@@ -212,9 +212,9 @@ def compose(
...
         nodes_p1: set[int] = set(nodes_p1_lst) | self.results.keys()  # Results contain preprocessed Pauli nodes
```
```suggestion
        nodes_p1: set[int] = nodes_p1 | self.results.keys()  # Results contain preprocessed Pauli nodes
```

graphix/pattern.py
```diff
@@ -212,9 +212,9 @@ def compose(
...
-        nodes_p2_lst, _ = other.get_graph()
+        nodes_p2_lst = other.extract_nodes()
```
```suggestion
        nodes_p2 = other.extract_nodes()
```

graphix/pattern.py
```diff
@@ -212,9 +212,9 @@ def compose(
...
         nodes_p2: set[int] = set(nodes_p2_lst) | other.results.keys()
```
```suggestion
        nodes_p2: set[int] = nodes_p2 | other.results.keys()
```

graphix/pattern.py
```diff
@@ -1121,42 +1097,46 @@ def get_angles(self) -> dict[int, ExpressionOrFloat]:
...
         """
-        nodes, edges = self.get_graph()
-        g: nx.Graph[int] = nx.Graph()
-        g.add_nodes_from(nodes)
-        g.add_edges_from(edges)
-        degree = g.degree()
+        graph = self.extract_graph()
+        degree = graph.degree()
         assert isinstance(degree, nx.classes.reportviews.DiDegreeView)
         return int(max(list(dict(degree).values())))
```
```suggestion
        return int(max(dict(degree).values()))
```

graphix/pattern.py
```diff
@@ -1166,10 +1146,10 @@ def get_isolated_nodes(self) -> set[int]:
         isolated_nodes : set of int
```
```suggestion
        isolated_nodes : set[int]
```

graphix/pattern.py
```diff
@@ -1166,10 +1146,10 @@ def get_isolated_nodes(self) -> set[int]:
...
         """
-        nodes, edges = self.get_graph()
-        node_set = set(nodes)
+        graph = self.extract_graph()
+        node_set = set(graph.nodes)
         connected_node_set = set()
-        for edge in edges:
+        for edge in graph.edges:
             connected_node_set |= set(edge)
         return node_set - connected_node_set
```
matulni: Maybe slightly more succinct?
```suggestion
        return {node for node, d in graph.degree if d == 0}
```

graphix/pattern.py
```diff
@@ -1166,10 +1146,10 @@ def get_isolated_nodes(self) -> set[int]:
         isolated_nodes : set of int
             set of the isolated nodes
         """
-        nodes, edges = self.get_graph()
-        node_set = set(nodes)
+        graph = self.extract_graph()
+        node_set = set(graph.nodes)
         connected_node_set = set()
```
```suggestion
```
(See previous comment)

tests/test_generator.py
```diff
@@ -145,14 +142,15 @@ def test_pattern_generation_no_internal_nodes(self) -> None:
         g: nx.Graph[int] = nx.Graph()
         g.add_edges_from([(0, 1), (1, 2)])
         pattern = generate_from_graph(g, {}, {0, 1, 2}, {0, 1, 2}, {})
-        assert pattern.get_graph() == ([0, 1, 2], [(0, 1), (1, 2)])
+        graph = pattern.extract_graph()
+        assert set(graph.nodes) == {0, 1, 2}
+        assert set(map(frozenset, graph.edges)) == {frozenset((0, 1)), frozenset((1, 2))}
```
matulni:
```suggestion
        assert nx.is_isomorphic(g, pattern.extract_graph())
```

tests/test_pattern.py
```diff
@@ -769,35 +769,8 @@ def test_get_graph(self) -> None:
...
-        for edge in edges_ref:
-            edge_match = False
-            for edge_ref in edges:
-                edge_match |= assert_equal_edge(edge, edge_ref)
-            if not edge_match:
-                edges_check2 = False
-        assert edges_check1
-        assert edges_check2
+        assert set(nodes_ref) == set(graph.nodes)
+        assert set(map(frozenset, edges_ref)) == set(map(frozenset, graph.edges))
```
matulni:
```suggestion
        graph_ref: nx.Graph[int] = nx.Graph(edges_ref)
        graph_ref.add_nodes(nodes_ref)
        assert nx.is_isomorphic(graph_ref, graph)
```

benchmarks/statevec.py
```diff
@@ -84,8 +84,8 @@ def simple_random_circuit(nqubit, depth):
     pattern = circuit.transpile()
     pattern.standardize()
     pattern.minimize_space()
-    nodes, edges = pattern.get_graph()
-    nqubit = len(nodes)
+    graph = pattern.extract_graph()
+    nqubit = len(graph.nodes)
```
matulni:
```suggestion
    nqubit = len(pattern.extract_nodes())
```

tests/test_pattern.py
matulni: maybe better with `nx.graphs_equal(graph_ref, graph)`

graphix/optimization.py
thierry-martinez: Renamed as `pattern` in 58d5fdd.

graphix/pattern.py
thierry-martinez: Fixed in 58d5fdd.

graphix/pattern.py
thierry-martinez: Fixed in 58d5fdd.

graphix/pattern.py
thierry-martinez: Fixed in 58d5fdd.

graphix/pattern.py
thierry-martinez: Fixed in 58d5fdd.

graphix/pattern.py
thierry-martinez: Fixed in 58d5fdd.

graphix/pattern.py
thierry-martinez: Nice! This looks indeed much more efficient than the original code. Thanks.

```python

import networkx as nx
from timeit import timeit

def isolated1(graph: nx.Graph[int]) -> set[int]:
    return { node for node, d in graph.degree if d == 0 }

def isolated2(graph: nx.Graph[int]) -> set[int]:
    node_set = set(graph.nodes)
    for edge in graph.edges:
        node_set -= set(edge)
    return node_set

graphs = [nx.gnp_random_graph(100, 0.5) for _ in range(100)]

print(f"{timeit(lambda: list(map(isolated1, graphs)), number=100)=}")
print(f"{timeit(lambda: list(map(isolated2, graphs)), number=100)=}")
```

Fixed in 58d5fdd.

graphix/pattern.py
thierry-martinez: Fixed in 58d5fdd.

tests/test_generator.py
thierry-martinez: Fixed in 58d5fdd (with `nx.utils.graphs_equal`).

tests/test_pattern.py
thierry-martinez: Fixed in 58d5fdd (with `nx.utils.graphs_equal`).

benchmarks/statevec.py
thierry-martinez: Fixed in 58d5fdd.

graphix/optimization.py
matulni: Thanks! We still have to initialize the output nodes as well, right? Or I'm missing something?

graphix/pattern.py
matulni: @thierry-martinez

graphix/optimization.py
thierry-martinez: Oops, sorry, I forgot about the order of the output nodes! Fixed in 9718275.

graphix/pattern.py
thierry-martinez: Oops. Fixed in 2c0b243.

matulni:
Thanks a lot for the PR @thierry-martinez , for me LGTM!

matulni (approved):

graphix/pattern.py
```diff
@@ -1734,12 +1703,14 @@ def measure_pauli(pattern: Pattern, leave_input: bool, copy: bool = False) -> Pa
     return pat
 
 
-def pauli_nodes(pattern: Pattern, leave_input: bool) -> tuple[list[tuple[command.M, PauliMeasurement]], set[int]]:
+def pauli_nodes(
+    pattern: optimization.StandardizedPattern, leave_input: bool
+) -> tuple[list[tuple[command.M, PauliMeasurement]], set[int]]:
     """Return the list of measurement commands that are in Pauli bases and that are not dependent on any non-Pauli measurements.
```
shinich1, line 1709:
```suggestion
    """Return the list of measurement commands that are in Pauli bases and that are not dependent on any non-Pauli measurements. 
    Requires :class:`graphix.optimization.StandardizedPattern` for correct evaluation of dependency.
```

PR #343, opened 2025-10-01
Add OpenQASM 3 exporter for circuits

thierry-martinez:
This commit introduces an OpenQASM 3 exporter for Graphix circuits. The functionality was originally proposed in #245 but has not yet been merged.

The added tests verify that a round-trip through the OpenQASM 3 representation preserves Graphix circuits, using the graphix-qasm3-parser plugin.  This plugin is therefore added as a `requirements-dev.txt` dependency.

CI is updated so that `pip install .` can detect the current version number of Graphix, instead of the default `0.1`: to do so, the whole history and the tags should be available.  We removed the "-e" option from CI because it is useless in the CI context.

In the long term the `qasm3_exporter` module will also host the pattern exporter, but that feature is intentionally omitted from this PR to keep the change focused.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,7 +2,7 @@
 mypy==1.18.2
 pre-commit # for language-agnostic hooks
 pyright
-ruff==0.14.0
+ruff==0.14.1
 
 # Stubs
 types-networkx==3.5.0.20250712
@@ -11,7 +11,7 @@ types-setuptools
 
 # Tests
 # Keep in sync with CI
-nox==2025.5.1
+nox==2025.10.16
 psutil
 pytest
 pytest-benchmark


graphix/qasm3_exporter.py
```diff
@@ -0,0 +1,95 @@
...
+        or instruction.kind == InstructionKind.RY
+        or instruction.kind == InstructionKind.RZ
+    ):
+        if not isinstance(instruction.angle, float):
+            raise ValueError("QASM export of symbolic pattern is not supported")
+        rad_over_pi = instruction.angle / pi
+        tol = 1e-9
+        frac = Fraction(rad_over_pi).limit_denominator(1000)
+        if abs(rad_over_pi - float(frac)) > tol:
+            angle = f"{rad_over_pi}*pi"
```
matulni: I'm a bit confused by this assignment. Isn't `angle` always rewritten in the conditional block just before the `return` statement ?

graphix/qasm3_exporter.py
```diff
@@ -0,0 +1,95 @@
...
+        or instruction.kind == InstructionKind.Y
+        or instruction.kind == InstructionKind.Z
+    ):
+        return f"{instruction.kind.name.lower()} q[{instruction.target}]"
+    if instruction.kind == InstructionKind.CNOT:
+        return f"cx q[{instruction.control}], q[{instruction.target}]"
+    if instruction.kind == InstructionKind.SWAP:
+        return f"swap q[{instruction.targets[0]}], q[{instruction.targets[1]}]"
+    if instruction.kind == InstructionKind.RZZ:
+        return f"rzz q[{instruction.control}], q[{instruction.target}]"
```
matulni: I could not find the `rzz` instruction in the [OpenQASM 3.0 API](https://openqasm.com/language/standard_library.html#single-qubit-gates).

I would say that the best we can do is to rewrite it as three instructions, `CNOT(control, target), Rz(target, angle), CNOT(control, target)`. In this case, we have to ensure that the `angle` parameter of the `RZZ` instruction is not an `Expression` (as we already do for the rotations).

graphix/qasm3_exporter.py
```diff
@@ -0,0 +1,95 @@
...
+    if any(instr.kind == InstructionKind.M for instr in circuit.instruction):
+        yield f"bit[{circuit.width}] b;"
+    for instr in circuit.instruction:
+        yield f"{instruction_to_qasm3(instr)};"
+
+
+def instruction_to_qasm3(instruction: Instruction) -> str:
+    """Get the qasm3 representation of a single circuit instruction."""
+    if instruction.kind == InstructionKind.M:
+        return f"b[{instruction.target}] = measure q[{instruction.target}]"
```
matulni, line 78: If I understood correctly, the `measure` primitive projects in the $Z$-basis (see [doc](https://openqasm.com/language/insts.html#Measurement)). However, `graphix.instruction.M` has attributes `plane` and `angle`. Shouldn't we rotate the measurement basis (i.e., cast measurement instructions into `M[plane=Plane.XZ, angle=0]` commands) before calling the qasm exporter ?

tests/test_qasm3_exporter.py
```diff
@@ -0,0 +1,36 @@
...
+@pytest.mark.parametrize("jumps", range(1, 11))
+def test_circuit_to_qasm3(fx_bg: PCG64, jumps: int) -> None:
+    rng = Generator(fx_bg.jumped(jumps))
+    nqubits = 5
+    depth = 4
+    circuit = rand_circuit(nqubits, depth, rng)
+    qasm = circuit_to_qasm3(circuit)
+    parser = OpenQASMParser()
+    parsed_circuit = parser.parse_str(qasm)
+    assert parsed_circuit.instruction == circuit.instruction
```
matulni: If you agree, I'd add a test on a circuit generated by hand which has all the gates supported by graphix.

graphix/qasm3_exporter.py
thierry-martinez: Nice catch. It seems to be `crz`: https://openqasm.com/language/standard_library.html#crz
(Actually, `crz` are translated into `rzz` in the OpenQASM parser: https://github.com/TeamGraphix/graphix-qasm-parser/blob/3e4dbc48662b9b13568e0bd17cad983cc902fe63/graphix_qasm_parser/parser.py#L295 )

tests/test_qasm3_exporter.py
emlynsg: Makes good sense to me.

graphix/qasm3_exporter.py
emlynsg: Agree here also. It appears you have to define alternative measurement bases as subroutines in openQASM (see https://openqasm.com/language/subroutines.html) since they aren't included by default.

thierry-martinez: You're right, the emitted instruction only works for `M[plane=Plane.XZ, angle=0]`. In a076142, I added a check that rejects any other measurements. I think we should provide functions to rewrite circuits to a particular set of instructions (in this case, measurements in a fixed form), and the OpenQASM exporter should not try to rewrite the circuit itself.

tests/test_qasm3_exporter.py
thierry-martinez: Done in a076142.

graphix/qasm3_exporter.py
thierry-martinez: Fixed in a076142.

graphix/qasm3_exporter.py
emlynsg: Agree. Should lines 65-71 (inclusive) be under an `else` going with the `if` in line 63?

thierry-martinez: @emlynsg, you're right! Actually, the logic that was implemented here was very close to the logic of `pretty_print.angle_to_str`, so I refactored the code a bit in a076142 to remove the code duplication.

graphix/qasm3_exporter.py
matulni: It seems that the workaround in @emlynsg's link amounts to adding extra instructions, which can be done at the level of the circuit (and the pattern), so I agree that it's best to separate the rewrite from the OpenQASM exporter (so that it's a mere translator which doesn't change the depth of the object).
I propose that we open an issue once this PR is merged to we don't forget

thierry-martinez: I just opened #352 to discuss about this.

PR #344, opened 2025-10-02
Revert `BaseM` removed in #205

thierry-martinez:
This reverts the removal of `BaseM`, which was deleted in commit a07197ae2097403ae0b2d62108b46ef7659a4331.

This commit subsumes #237, which restores `BaseM` but lacks the associated documentation

Currently, `BaseM` is a type alias for `M`, which is not particularly useful. However, `BaseM` is used in Veriphix in blind patterns: the server only knows which node is measured, and the measurement itself is specified by the client's `MeasureMethod`.

This commit is a step towards upstreaming Graphix's features needed for Veriphix (and for UBQC in general).

- Format added code by `ruff`

Then, please fill in below:

**Context (if applicable):**

**Related issue:**

### DIFF
--- a/graphix/command.py
+++ b/graphix/command.py
@@ -67,10 +67,10 @@ class N(_KindChecker, DataclassReprMixin):
 class BaseM(DataclassReprMixin):
     """Base measurement command.
 
-    Represent a measurement of a node. In `graphix`, a measurement is an instance of 
-    class `M`, with given plane, angles, and domains. The base class `BaseM` allows users to define 
-    new class of measurements with different abstractions. For example, in the context 
-    of blind computations, the server only knows which node is measured, and the parameters 
+    Represent a measurement of a node. In `graphix`, a measurement is an instance of
+    class `M`, with given plane, angles, and domains. The base class `BaseM` allows users to define
+    new class of measurements with different abstractions. For example, in the context
+    of blind computations, the server only knows which node is measured, and the parameters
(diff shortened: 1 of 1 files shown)


graphix/command.py
```diff
@@ -64,7 +64,21 @@ class N(_KindChecker, DataclassReprMixin):
...
 
 @dataclasses.dataclass(repr=False)
-class M(_KindChecker, DataclassReprMixin):
+class BaseM(DataclassReprMixin):
+    """Base measurement command.
+
+    Represent a measurement of a node. In MBQC, a measure is an instance of `M`,
+    with given plane, angles, and domains. In the context of blind computations,
+    the server only knows which node is measured, and the parameters are given
+    by the :class:`graphix.simulator.MeasureMethod` provided by the client.
```
shinich1:
```suggestion
    Represent a measurement of a node. In `graphix`, a measurement is an instance of 
    class `M`, with given plane, angles, and domains. The base class `BaseM` allows users to define 
    new class of measurements with different abstractions. For example, in the context 
    of blind computations, the server only knows which node is measured, and the parameters 
    are given by the :class:`graphix.simulator.MeasureMethod` provided by the client.
```

shinich1 (approved):
Please consider changing the docstring with more context (suggestion above; anything along the line of above is good to me). Looks good otherwise!

PR #337, opened 2025-09-05
$O(N^3)$ Pauli-flow finding algorithm

matulni:
This PR introduces the $O(N^3)$ Pauli flow-finding algorithm presented in [1] which improves the previous version with $O(N^5)$ complexity [2]. Since a Pauli flow in an open graph without Pauli measurements defines a gflow, this algorithm also improves the previous implementation of the gflow-finding algorithm with $O(N^4)$ complexity [3].

The implementation proposed here considers the suggestion in [1] to eliminate the rows that are identically 0 in the order-demand matrix to accelerate the flow-finding routine in open graphs with $n_I \neq n_O$. 

Finally, the new algorithm solves the issue discussed in qat-inria/graphix#17 , whereby the previous flow-finding functions could return a correction function containing input nodes in the codomain, which is not allowed by the definition of gflow or Pauli flow. 

In summary:

- The core functions of the algorithm are encapsulated in the new module 
`graphix.find_pflow`, while `graphix.gflow` and related modules are adapted accordingly to avoid modifying the current API in this PR.
- The module `linalg.py` has been updated with more efficient functions for gaussian elimination, back substitution, kernel-finding and right-inverse. New implementation is $\mathbb{F}_2$-specialized and, where appropriate, just-in-time compiled with `numba`. The dependence on `simpy` and `galois` is dropped. Tests are simplified and updated accordingly, and module is now typed. 
- `pytest-benchmark` is incorporated to allow for easier benchmarking between different commits (courtesy of @thierry-martinez). 

## Additional comments

- This PR addresses the same issue as #294, but is otherwise unrelated.
- The module `graphix.gflow` is currently functioning but rather messy. We leave its restructuration for a future PR to simply the review process. 

## Benchmarking

The implementation exhibits the cubic scaling as described in [1].  The plot below was generated with [this code](https://gitlab.inria.fr/muldemol/pflow-on3-scaling.git).
scaling-20250930_1759

## References

[1] Mitosek and Backens, 2024 (arXiv:2410.23439)
[2] Simmons, 2021 (arXiv:2109.05654)
[3] Backens et al., Quantum 5, 421 (2021)

Issue #168, opened 2024-06-18
Support for arbitrary inputs in Pauli preprocessing


thierry-martinez:
**Is your feature request related to a problem? Please describe.**
Pauli preprocessing is currently only available when inputs are initialized in |+> state.

**Describe the feature you'd like**
Pauli preprocessing should support arbitrary inputs.

**Additional context**
Support for arbitrary inputs in state vector and density matrix simulations has been handled in https://github.com/TeamGraphix/graphix/pull/135 .

mgarnier59:
Edit: in the current version, Pauli preprocessing can only handle stabiliser states.

PR #347, opened 2025-10-06
Enable type-checking for `visualization.py`

thierry-martinez:
This commit adds type annotations to `visualization.py`, along with its corresponding tests, enabling full type-checking with mypy while preserving existing functionality.

**Related issue:**
This PR continues the work started in #302, #308, and #312.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - #354, #357: `RZZ` gates are now supported by the transpiler.
 
+- #347: The `visualization.py` module has been partially refactored
+  and is now well-typed. Pauli nodes are now systematically detected.
+
 ### Changed
 
 - #337: Dropped dependence on `sympy` and `galois`.
@@ -50,6 +53,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - #220, #332: `Pattern.get_measurements_commands` is renamed into
   `Pattern.extract_measurement_commands`.
+
+- #347: There is no longer `save` flag in `Pattern.draw_graph`: the
+  figure is saved to a file as soon as `filename` is not `None` (i.e.,
+  contains a `Path`). In this case, no window is opened to show the
+  plot interactively (i.e., the plot is shown interactively in a
+  window only if `filename` is `None`).
 
 ## [0.3.2] - 2025-08-12
 


graphix/pattern.py
emlynsg, line 1: Eventually we should update the docstrings for consistency too.

graphix/visualization.py
emlynsg, line 1: - Do we want to change all the figsize variables to _Point = tuple[float, float] type?
- Line 768 do we want to give a more descriptive ValueError?
- Additional bug noticed in GraphVisualizer __init__ . If no measurement planes input, generated measurement planes include measurements on output nodes. I will make an issue for this.

thierry-martinez: I think that even if `figsize` could formally be of type `_Point`, it makes sense to distinguish them since a size is semantically different than a point. The two other issues should have been fixed by the refactoring. Thanks!

thierry-martinez:
> @thierry-martinez @matulni this may have been dealt with already (identified as past fixed bug), or a new issue. If the latter it would make sense as a separate issue and PR.

This change of behavior was introduced by #337, indeed: the definition of the gflow was strengthened to match the definition from the paper, where input nodes cannot be corrected by a gflow. This strengthened definition seems to be incompatible with Pauli presimulation preserving the gflow, leading to examples where a gflow was detected before #337 but is no longer detected. However, Pauli presimulation preserving the gflow is a desirable and expected property, so we should open a separate issue about this.

Anyway, I think this problem is outside the scope of this PR.

In my last commits:
- I changed the behavior of `draw` functions when a `filename` is given so that they do not open the drawing in a window, since simply saving the drawing to a file seems more useful.
- I use `pytest-mpl` to compare the output against a reference file to catch future regressions in plotting more easily.

matulni:
> This change of behavior was introduced by #337, indeed: the definition of the gflow was strengthened to match the definition from the paper, where input nodes cannot be corrected by a gflow. This strengthened definition seems to be incompatible with Pauli presimulation preserving the gflow, leading to examples where a gflow was detected before #337 but is no longer detected. However, Pauli presimulation preserving the gflow is a desirable and expected property, so we should open a separate issue about this.
> 
> Anyway, I think this problem is outside the scope of this PR.

I fully agree. For completeness, I link a related issue: https://github.com/TeamGraphix/graphix/issues/168

graphix/visualization.py
```diff
@@ -42,12 +55,12 @@ class GraphVisualizer:
...
-        v_in: list[int],
-        v_out: list[int],
-        meas_plane: dict[int, Plane] | None = None,
-        meas_angles: dict[int, ExpressionOrFloat] | None = None,
-        local_clifford: dict[int, Clifford] | None = None,
+        g: nx.Graph[int],
+        v_in: Collection[int],
+        v_out: Collection[int],
+        meas_plane: Mapping[int, Plane] | None = None,
+        meas_angles: Mapping[int, ExpressionOrFloat] | None = None,
```
matulni, line 62: Would it make sense to pass an `OpenGraph` as argument instead (or even, adding an `OpenGraph` attribute to the `GraphVisualizer` class ?
(We can leave this as is, and modify it when interfacing the visualization module with the flow/opengraph refactoring)

graphix/visualization.py
```diff
@@ -437,334 +331,134 @@ def visualize_w_gflow(
...
+
         plt.figure(figsize=figsize)
 
         for edge in edge_path:
             if len(edge_path[edge]) == 2:
                 nx.draw_networkx_edges(self.graph, pos, edgelist=[edge], style="dashed", alpha=0.7)
             else:
-                t = np.linspace(0, 1, 100)
-                curve = self._bezier_curve(edge_path[edge], t)
+                curve = self._bezier_curve_linspace(edge_path[edge])
```
matulni: Maybe slightly more idiomatic:
```python
for edge, path in edge_path.items():
...
```
And replace `edge_path[edge]` by `path`

graphix/visualization.py
```diff
@@ -437,334 +331,134 @@ def visualize_w_gflow(
...
-            )  # Add layer label at bottom
+        if l_k is not None:
+            # Draw the vertical lines to separate different layers
+            for layer in range(min(l_k.values()), max(l_k.values())):
+                plt.axvline(
+                    x=(layer + 0.5) * node_distance[0], color="gray", linestyle="--", alpha=0.5
+                )  # Draw line between layers
+            for layer in range(min(l_k.values()), max(l_k.values()) + 1):
+                plt.text(
+                    layer * node_distance[0], y_min - 0.5, f"l: {max(l_k.values()) - layer}", ha="center", va="top"
```
matulni: In matplotlib's default font, `l` looks like `I`. I propose to capitalize `L` for clarity.
```suggestion
                    layer * node_distance[0], y_min - 0.5, f"L: {max(l_k.values()) - layer}", ha="center", va="top"
```

graphix/visualization.py
```diff
@@ -437,334 +331,134 @@ def visualize_w_gflow(
...
-            Filename of the saved plot.
-        """
-        pos = self.get_pos_wo_structure()
-        pos = {k: (v[0] * node_distance[0], v[1] * node_distance[1]) for k, v in pos.items()}  # Scale the layout
-
-        if figsize is None:
-            figsize = self.get_figsize(None, pos, node_distance=node_distance)
-        plt.figure(figsize=figsize)
+        if corrections is not None:
+            plt.legend(loc="upper right", fontsize=10)
```
matulni: Can we merge this block with the `if corrections is not None` block in L407 above ?

graphix/visualization.py
```diff
@@ -84,11 +97,10 @@ def visualize(
         show_local_clifford: bool = False,
         show_measurement_planes: bool = False,
         show_loop: bool = True,
-        node_distance: tuple[int, int] = (1, 1),
+        node_distance: tuple[float, float] = (1, 1),
```
matulni, line 100: I notice that the parameters
```python
show_pauli_measurement,
show_local_clifford,
show_measurement_planes,
show_loop,
node_distance,
figsize,
filename
```
are passed around by various methods. Could we make the code more organised by considering them as attributes of `GraphVisualizer` (with the current default values) ?

graphix/visualization.py
```diff
@@ -388,31 +276,31 @@ def __draw_nodes_role(self, pos: dict[int, tuple[float, float]], show_pauli_meas
                 inner_color = "lightgray"
             elif (
                 show_pauli_measurement
-                and isinstance(self.meas_angles, SupportsFloat)
-                and (
-                    2 * self.meas_angles[node] == int(2 * self.meas_angles[node])
-                )  # measurement angle is integer or half-integer
+                and self.meas_angles is not None
+                and PauliMeasurement.try_from(Plane.XY, self.meas_angles[node]) is not None
```
matulni, line 305: Why `Plane.XY` instead of `self.meas_plane[node]`?

graphix/visualization.py
```diff
@@ -892,38 +553,43 @@ def _point_from_node(pos, dist, angle):
...
+        bezier_path = list(bezier_path)
+        max_iter = 5
+        iteration = 0
+        nodes = set(self.graph.nodes())
+        while True:
+            iteration += 1
+            intersect = False
+            if iteration > max_iter:
+                break
+            ctrl_points = []
```
matulni: Annotate for better readability ?
```suggestion
            ctrl_points: list[tuple[int, _Point]] = []
```

graphix/visualization.py
```diff
@@ -892,38 +553,43 @@ def _point_from_node(pos, dist, angle):
...
+            for i in range(len(bezier_path) - 1):
+                start = bezier_path[i]
+                end = bezier_path[i + 1]
+                for node in set(nodes):
+                    if node != arrow[0] and node != arrow[1] and self._edge_intersects_node(start, end, pos[node]):
+                        intersect = True
+                        ctrl_points.append(
+                            (
+                                i,
+                                self._control_point(start, end, pos[node], distance=0.6 / iteration),
```
matulni: I think that before the refactor it was `self._control_point(bezier_path[0], bezier_path[-1], pos[node], distance=0.6 / iteration)`.
I'd say they are not equivalent, is this change intentional (part of a bugfix) ?

graphix/visualization.py
thierry-martinez: Fixed in f3cd9f0.

graphix/visualization.py
thierry-martinez: Fixed in f3cd9f0.

graphix/visualization.py
thierry-martinez: I added the following comment:
```
                # Pauli nodes are checked with Plane.XY by default,
                # because the actual plane does not change whether the
                # node is Pauli or node, and the current API allows
                # self.meas_plane to be None while self.meas_angles is
                # defined.
```

graphix/visualization.py
thierry-martinez: Nice catch: indeed, the change wasn't intentional and I believe the original code makes more sense.

graphix/visualization.py
thierry-martinez: I would prefer to keep this kind of refactoring for a later PR.

graphix/visualization.py
thierry-martinez: Likewise, I would prefer to keep this kind of refactoring for a later PR.

graphix/visualization.py
matulni: Sounds good

graphix/visualization.py
matulni: I see, thanks!

thierry-martinez:
Nice catch, @emlynsg ! I added the `--mpl` option, which allowed me to catch and fix another bug with scaling: it appears to be essential to defer the edge computation after the scaling of paths to properly avoid overlapping nodes. These tests also reveal there has been a change between Python 3.9 and 3.10 where the Y-axis in matplotlib was reversed. I disabled `--mpl` option in Python 3.9 to circumvent the problem.

PR #357, opened 2025-11-03
Fix #354: Add RZZ gate support to the transpiler

### LINKED ISSUE #354: [Bug]: ```Circuit.transpile``` doesn't support $\text{RZZ}$ gates while the rest of ```transpiler.py``` does.
**Describe the bug**
```transpiler.py``` currently supports the common gate operations in Graphix, including the $\text{RZZ}$ gate. However, the method ```Circuit.transpile``` does not and raises an error.

**To Reproduce**
```
import numpy as np
from graphix import Circuit
circuit = Circuit(2, instr=[instruction.RZZ(0, 1, np.pi / 4)])

**Expected behavior**
The gate should be managed in the transpiler in an equivalent way to other gates.

**Environment (please complete the following information):**

- OS: Ubuntu
- Python version: 3.13.7

**Additional context**
Problem to be avoided completely by upcoming transpiler reworking.

thierry-martinez:
This commit adds RZZ gate support to the transpiler that was missing (#354).

The commit defines the internal `_remove_rzz` function, which is a transpiler for removing the RZZ gate from the gate set. It can be part of a broader circuit-to-circuit transpiler in the future.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,7 +2,7 @@
 mypy==1.18.2
 pre-commit # for language-agnostic hooks
 pyright
-ruff==0.14.3
+ruff==0.14.5
 
 # Stubs
 types-networkx==3.5.0.20250712
@@ -11,7 +11,7 @@ types-setuptools
 
 # Tests
 # Keep in sync with CI
-nox==2025.10.16
+nox==2025.11.12
 psutil
 pytest
 pytest-benchmark


matulni (approved):
Looks good, thanks!

graphix/transpiler.py
```diff
@@ -1013,3 +1013,13 @@ def _extend_domain(measure: M, domain: set[int]) -> None:
         measure.s_domain ^= domain
     else:
         measure.t_domain ^= domain
+
+
+def _remove_rzz(instructions: Iterable[Instruction]) -> Iterator[Instruction]:
```
shinich1: we're not removing `rzz`. So perhaps `transpile_rzz`?

thierry-martinez: Fixed in ec639bf. Thanks!

Issue #306, opened 2025-06-23
Add a function that checks if a pattern implements a given flow


thierry-martinez:
We are missing a function that checks whether a given pattern conforms to a given flow.

This function would be complementary to the open issue #120, which proposes extending `generate_from_pattern` to enable it to generate a pattern from a given open graph and a given flow.

EarlMilktea:
@thierry-martinez 

I'm working on this in the new PR (takeover of #280) and now encountering some problems.
Please let me know your comments on them.

- On `.get_layers`

Is `.get_layers` guaranteed to give the minimum number of distinct layers?
If so, is it consistent with maximally-delayed flow/gflow (can we always find a corresponding correction set)?

- How to fix layers

Is the algorithm `.get_layers` the only way to find layers?

thierry-martinez:
> I'm working on this in the new PR (takeover of [#280](https://github.com/TeamGraphix/graphix/pull/280)) and now encountering some problems.

I’m happy that you’re working on this new issue, but it does not necessarily have to be solved in the same PR as the integration with fastflow. From my perspective, #280 was quite close to being merged: my only remarks were mainly to revert the introduction of `gflow_shim` and the change in the signature of the `find_flow` functions. It would be better, in my opinion, to avoid postponing the merge and to keep the PR manageable. However, before the merge, it would be good to have benchmarks that show the performance gain compared to the previous `find_flow` implementations.
     
> Is `.get_layers` guaranteed to give the minimum number of distinct layers?

A priori, yes: taking for each layer the nodes that are minimal once we remove the nodes placed in previous layers is optimal.

> If so, is it consistent with maximally-delayed flow/gflow (can we always find a corresponding correction set)?

No, we have neither uniqueness of maximally-delayed flow nor uniqueness of layering for a given flow.

I think the right approach to check whether a pattern implements a given flow is:
- To check that the pattern is runnable (that is to say, that N command prepare fresh nodes, E and M commands are applied on prepared but not yet measured nodes, and that the domains of M, X, and Z only consist only on measured nodes).
- To check that the corrections match those implied by the flow function.

Issue #196, opened 2024-08-05
Improve testing of Open graph to pattern conversions


wlcsm:
Currently we only test converting Open graphs to (`OpenGraph.to_pattern()`) and from (`OpenGraph.from_pattern()`) patterns with a couple of handcrafted examples in `tests/test_opengraph.py`.

It would instead be more robust to test these methods on many randomly generated graphs and patterns (by modifying `random_objects.py` with random open graph generation), and check the results against a simulator.

pafloxy:
How are you planning to generate the open-graphs ?

wlcsm:
Great question. I do not know. Open graphs need four items, a graph, measurements for non-output nodes, inputs and outputs. All of these could be randomised however it is very likely the resulting open graph would not have flow.

We could simply repeat the procedure enough times, checking each time whether the open graph has flow and then stop when we find one. It is not very efficient but it is simple.

Another way would be to randomly generate circuits in PyZX, then convert those to a graph using PyZX's in-built `to_graph` method and then convert these to open graphs via the `OpenGraph.from_pyzx_graph` method. This has the advantage of always generating an OpenGraph with flow, but the problem is that we create a reliance on the PyZX module which is supposed to be optional

thierry-martinez:
I think that #358 at least partially solved this issue by adding new graphs in the test suite for `OpenGraph`, and the test `test_from_to_pattern`, which tests the round-trip from random circuits.

These tests could be completed with tests on random graphs once we have a random generator for `OpenGraph` (#377).

PR #358, opened 2025-11-03
Refactor of flow tools - Part I

matulni:
## Summary

This is the first of a series of PRs refactoring the flow tools in the library.

In the current implementation, the transformation "Open graph" $\rightarrow$ "Flow" $\rightarrow$ "XZ-corrections" $\rightarrow$ "Pattern" occurs under the hood in the `:func: OpenGraph.to_pattern()` method. This refactor aims at exposing the intermediate objects to offer users more control on pattern extraction and ease the analysis of flow objects.

Related issues: #120, #306, #196, #276, #181  

## File structure

- `graphix.opengraph.py` introduces the new `OpenGraph` objects. The associated file `tests.test_opengraph.py` contains the bulk of the unit tests verifying the correctness of the flow-finding algorithms (these tests have been refactor from `tests.test_gflow.py` and `tests.test_find_pflow`).

- `graphix.flow.core.py` contains all the flow-related functionalities, except the flow-finding algorithms which are placed in `graphix.flow._find_cflow.py` and `graphix.flow._find_gpflow.py`

- The existing files `graphix.gflow.py`, `graphix.find_pflow.py` will be eliminated once the refactor is completed. Their modification consists of just minor changes to comply with the new interface of `OpenGraph`. The module `graphix.pattern` still depends on the old flow implementation.

### Public API

```mermaid
flowchart LR
n0(**OpenGraph**)
subgraph s1[" "]
	n10(**PauliFlow**)
	n11(**GFlow**)
	n12(**CausalFlow**)
end
n2(**XZCorrections**)
n3(**Pattern**)

n0 --> s1
s1 .-> n0
n2 -- "+ total order" --> n3
n3 --> n2
n2 .-> n0
n3 .-> n0
s1 --> n2

n10 --> n11
n11 --> n12


linkStyle 7 stroke: #276cf5ff
linkStyle 8 stroke: #276cf5ff
linkStyle 2 stroke: #F54927
linkStyle 3 stroke: #F54927
linkStyle 5 stroke: #F54927
```
- Blue arrows indicate inheritance (from parent to child).
- Orange arrows indicate that the conversion contains information on the measurement angles.
- Dashed arrows indicate "containment", e.g. `XZCorrections` contains an `OpenGraph`.

### Examples

**Define an open graph without measurement angles and extract a gflow**

```python
import networkx as nx
from graphix.opengraph_ import OpenGraph

og = OpenGraph(
        graph=nx.Graph([(1, 3), (2, 4), (3, 4), (3, 5), (4, 6)]),
        input_nodes=[1, 2],
        output_nodes=[5, 6],
        measurements=dict.fromkeys(range(1, 5), Plane.XY),
    )
gflow = og.find_gflow()
print(gflow.correction_function) # {1: {3, 6}, 2: {4, 5}, 3: {5}, 4: {6}}
print(gflow.partial_order_layers) # [{5, 6}, {3, 4}, {1, 2}]
```

**Define an open graph with measurement angles and extract flow, corrections and pattern**
```python
import networkx as nx
from graphix.measurements import Measurement
from graphix.opengraph_ import OpenGraph

og = OpenGraph(
        graph=nx.Graph([(0, 1), (1, 2), (2, 3)]),
        input_nodes=[0],
        output_nodes=[3],
        measurements={
            0: Measurement(0, Plane.XY), # `Plane.XY` if causal or gflow, `Axis.X` if Pauli flow
            1: Measurement(0, Plane.XY), # `Plane.XY` if causal or gflow, `Axis.X` if Pauli flow
            2: Measurement(0.2, Plane.XY),
        },
    )

# Causal flow
cflow = og.find_causal_flow()

print(cflow.correction_function) # {2: {3}, 1: {2}, 0: {1}}
print(cflow.partial_order_layers) # [{3}, {2}, {1}, {0}]

corr = cflow.to_corrections() # {domain: nodes}
print(corr.x_corrections) # {2: {3}, 1: {2}, 0: {1}}
print(corr.z_corrections) # {1: {3}, 0: {2}}
print(corr.to_pattern().to_unicode()) # X₃² M₂(π/5) X₂¹ Z₃¹ M₁ X₁⁰ Z₂⁰ M₀ E₂₋₃ E₁₋₂ E₀₋₁ N₃ N₂ N₁


# Pauli flow
pflow = og.find_pauli_flow()

print(pflow.correction_function) # {0: {1, 3}, 1: {2}, 2: {3}}
print(pflow.partial_order_layers) # [{3}, {2}, {0, 1}]

corr = pflow.to_corrections() # {domain: nodes}
print(corr.x_corrections) # {2: {3}, 0: {3}, 1: {2}}
print(corr.z_corrections) # {1: {3}}
print(corr.to_pattern().to_unicode()) # X₃² M₂(π/5) X₂¹ Z₃¹ M₁ X₃⁰ M₀ E₂₋₃ E₁₋₂ E₀₋₁ N₃ N₂ N₁
```

### Attributes of the objects in the public API

- **OpenGraph** (frozen dataclass)
	- `graph: nx.Graph[int]`
	- `input_nodes: Sequence[int]`
	- `output_nodes: Sequence[int]`
	- `measurements: Mapping[int, _M_co]`

- **PauliFlow** (frozen dataclass)
	- `og: OpenGraph[_M_co]`
	- `correction_function: Mapping[int, AbstractSet[int]]`
	- `partial_order_layers: Sequence[AbstractSet[int]]`

- **GFlow, CausalFlow**

	Same as **PauliFlow** but with `og: OpenGraph[_PM_co]`

- **XZCorrections** (frozen dataclass)
	- `og: OpenGraph[_M_co]`
	- `x_corrections: Mapping[int, AbstractSet[int]]`
	- `z_corrections: Mapping[int, AbstractSet[int]]`
    - `partial_order_layers: Sequence[AcstractSet[int]]`

All the new classes inherit from `Generic[_M_co]`. We introduce the following parametric types:
- `_M_co = TypeVar("_M_co", bound=AbstractMeasurement, covariant=True)`
- `_PM_co = TypeVar("_PM_co", bound=AbstractPlanarMeasurement, covariant=True)`

##### Why do `PauliFlow`, its children and `XZCorrections` have a  `partial_order_layers` attribute ?

In Ref. [1], it's noted that the correction function and the partial order of a flow are not independent, but rather, that the former determines the latter:

"In the problem of finding Pauli flow, instead of looking for correction function and partial order, it now suffices to look only for a (focused) extensive correction function. [...] Given a correction function $c$ we can define a minimal relation induced by the Pauli flow definition, and consider whether this relation extends to a partial order."

This observation suggests that flow objects should not have a "partial order" attribute, but instead, that it should be computed from the correction function when needed, e.g., when computing the $X,Z$-corrections for a Pauli flow. However, this operation is costly and, crucially, the partial order is a byproduct of the flow-finding algorithms. Therefore, it is reasonable to store it as an attribute at initialization time instead.

We store in the form of layers owing to the following reasons:

- It is the most convenient form to compute the $X,Z$-corrections for a Pauli flow.
- The recursive $O(N^2)$-algorithm (for causal flow) _simultaneously_ computes the partial order in the form of layers and the correction function.
- The algebraic $O(N^3)$-algorithm (for gflow and Pauli flow) computes the adjacency matrix of a directed graph. In order to verify that such directed graph is acyclical (and therefore, that it indeed encodes a partial order), we need to perform a topological sort, which amounts to computing the partial order layers.
- The partial order of flows is often expressed in terms of an ordered sequence of sets in the literature. 

Nevertheless, we endow the flow objects with class methods allowing to initialize them directly from a correction matrix or a correction function.

Similarly, the mappings `XZCorrections.x_corrections` and `XZCorrections.z_corrections` define a partial order which is necessary to extract a pattern (specifically, the input "total order" has to be compatible with the intrinsic partial order of the XZ-corrections). Computing this partial order is expensive (it amounts to extracting a DAG and doing a topological sort as well), therefore we also store it as an attribute. Further, when the `XZCorrections` instance is derived from a flow object, the flow's partial order is a valid partial order for the XZ-corrections, so it is useful to pass it as a parameter to the dataclass constructor.

We also introduce an static method allowing to initialize `XZCorrections` instances directly from the `x_corrections` and the `z_corrections` mappings.

##### Why do we treat open graphs as parametric objects ?
- Trying to find a gflow (or a causal flow) only makes sense on an open graph whose measurement labels are planes (not Pauli axes). We want this check to be performed by the static type checker.
- Extracting a pattern from an open graph (or an `XZCorrections` object) requires information on the measurement angles. Treating the angles list as an object independent of the open graph demands a rather cumbersome bookkeeping (e.g., angles must be defined w.r.t. measured nodes, for Pauli axes the notion of angle is replaced by a `Sign`, etc.). At the same time, certain operations on an open graph (namely, trying to find flows), do not need information on the angles. Therefore, it should be possible to have "plane/axis-open graphs" and "measurement-open graphs", respectively without and with information about the angles. The static type-checker should only allow "to_pattern" conversions on open graphs with angle information.

We propose to achieve this by implementing a "type superstructure" on top of the existing classes `Axis`, `Plane` and `Measurement` to avoid modifying the existing functionalities unrelated to flows. We detail this next.

### Measurement types

```mermaid
flowchart TD
	n(**ABC**)

	n0("**AbstractMeasurement**")

	n1("**AbstractPlanarMeasurement**")

	n2("**Enum**")
	
	n3("**Plane**")
	n4("**Axis**")
	n5("**Measurement** (dataclass)")

n .-> n0
n0 .-> n1
n0 .-> n4
n1 .-> n5
n2 .-> n3
n2 .-> n4
n1 .-> n3
```

This type structure should allow the following:

- On `OpenGraph[Axis|Plane]` we can look for Pauli flow only (enforced statically).
- On `OpenGraph[Plane]` we can look for causal, g- and Pauli flow.
- On `OpenGraph[Measurement]` we can look for causal, g- and Pauli flow.
	- If we look for causal or g-flow, measurements with a Pauli angle should be intepreted as **planes**.
	- If we look for a Pauli flow, measurements with a Pauli angle should be intepreted as **axes**.

To that purpose, we introduce the following abstract class methods:

- `:func: AbstractMeasurement.to_plane_or_axis`
- `:func: AbstractPlanarMeasurement.to_plane`

Further, a conversion to `Pattern` is only possible from `OpenGraph[Measurement]` and `XZCorrections[Measurement]` objects (which have information on the measurement angles).

### Methods implemented in this PR
(See docstrings for further details)

#### `OpenGraph`
- `from_pattern` (static method)
- `to_pattern`
- `neighbors`
- `odd_neighbors`
- `find_causal_flow`
- `find_gflow`
- `find_pauli_flow`

#### `PauliFlow` (and children)

- `from_correction_matrix` (class method)

- `to_corrections`

#### `XZCorrections`

- `from_measured_nodes_mapping` (static method)

- `to_pattern`

- `generate_total_measurement_order`
- `extract_dag`
- `is_compatible`

#### Comments

- The old function `graphix.generator._pflow2pattern`,  `graphix.generator._gflow2pattern` and `graphix.generator._flow2pattern` are respectively implemented in `PauliFlow/GFlow/CausalFlow.to_corrections` and `XZCorrections.to_pattern`.

- The method `OpenGraph.to_pattern` provides an immediate way to extract a deterministic pattern on a given resource open graph state in similar way to the old API: it attempts to compute a flow (first causal, then Pauli) and then uses the recipe in Ref. [3] to write the pattern.

    However, the new API also allows the following:
    - Extract a pattern from the open graphs's Pauli flow even if it also has a causal flow.
    - Extract a pattern from a custom flow.
    - Extract a pattern from a custom set of XZ-corrections (runnable, but not necessarily deterministic).
	- Construct a pattern from a set of XZ-corrections with a custom total order.

### Finding flows of an open graph

We provide here additional details on the interface of the `OpenGraph` object with the flow-finding algorithms.

#### Causal flow

```mermaid
flowchart TD

	n0["OpenGraph.find_causal_flow()"]

	n1["find_cflow(self)"]

	n2("None")

	n3("CausalFlow")

n0 --> n1
n1 --> n2
n1 --> n3

style n2 stroke:#A6280A, stroke-width:4px
style n3 stroke:#0AA643, stroke-width:4px
```

The function `graphix.flow._find_cflow.find_cflow` implements the layer-by-layer algoritm with $O(N^2)$ complexity in Ref. [2]

#### Gflow and Pauli flow

```mermaid
---
config:
layout: elk
---
flowchart TD

	n0["OpenGraph.find_gflow()"]

	n00["PlanarAlgebraicOpenGraph(self)"]

	n1["compute_correction_matrix(aog)"]

	n3["GFlow.from_correction_matrix(correction_matrix)"]

	n9("GFlow")
	
	n2("None")

	nn0["OpenGraph.find_pauli_flow()"]

	nn00["AlgebraicOpenGraph(self)"]

	nn1["compute_correction_matrix(aog)"]

	nn3["PauliFlow.from_correction_matrix(correction_matrix)"]

	nn9("PauliFlow")
	
	nn2("None")

n0 --> n00
n00 --"aog"--> n1
n1 --> n2
n1 -- "correction_matrix"--> n3
n3 --> n9
n3 --> n2

nn0 --> nn00
nn00 --"aog"--> nn1
nn1 --> nn2
nn1 -- "correction_matrix"--> nn3
nn3 --> nn9
nn3 --> nn2

style n2 stroke:#A6280A, stroke-width:4px
style n9 stroke:#0AA643, stroke-width:4px

style nn2 stroke:#A6280A, stroke-width:4px
style nn9 stroke:#0AA643, stroke-width:4px
```

---

The function `graphix.flow._find_gpflow.compute_correction_matrix` performs the first part of the algebraic flow-finding algorithm in Ref. [1]. The second part (i.e., verifying that a partial order compatible with the correction matrix exists) is done by the class method `.from_correction_matrix`:

```mermaid
---
config:
layout: elk
---
flowchart TD

	n3[".from_correction_matrix(correction_matrix)"]

	n7["compute_partial_order_layers(correction_matrix)"]

	n5["correction_matrix.to_correction_function()"]
	n2("None")

	n9("cls(
	aog.og, correction_function, partial_order_layers)")
	


n3 --> n5
n3 --> n7
n7 -- "partial_order_layers" --> n9
n5 -- "correction_function" --> n9
n7 --> n2


style n2 stroke:#A6280A, stroke-width:4px
style n9 stroke:#0AA643, stroke-width:4px
```

##### Details on the interface with the algebraic flow-finding algorithm

We introduce two parametric dataclasses:

```mermaid
flowchart TD
	n0("**AlgebraicOpenGraph[_M_co]**")
	n1("**PlanarAlgebraicOpenGraph[_PM_co]**")
n0 --> n1

linkStyle 0 stroke: #276cf5ff
```
where the parametric type variable `_M_co` and `_PM_co` where defined [above](#attributes-of-the-objects-in-the-public-api).

`AlgebraicOpenGraph` is a dataclass to manage the mapping between the open graph nodes and the row and column indices of the matrices involved in the algebraic flow-finding algorithm. This class replaces the class `OpenGraphIndex` introduced in [PR 337](https://github.com/TeamGraphix/graphix/pull/337).

The class `PlanarAlgebraicOpenGraph` only rewrites the method `_compute_og_matrices` which calculates the flow-demand $M$ and the order-demand $N$ matrices by reading out the measurement planes (or axes) of the open graph. In particular,

- `AlgebraicOpenGraph._compute_og_matrices` calls `self.og.measurements[node].to_plane_or_axis` which will intepret measurements with a Pauli angle as axes.

	This is the adequate behavior when we seek to compute a Pauli flow.

- `PlanarAlgebraicOpenGraph._compute_og_matrices` calls `self.og.measurements[node].to_plane` which will intepret measurements with a Pauli angle as planes.

	This is the adequate behavior when we seek to compute a gflow or a causal flow.

## Feautures to be included in the next PR of this series

- Methods to verify the correctness of flow and XZ-corrections objects:

    - `PauliFlow.is_well_formed`
	- `GFlow.is_well_formed`
	- `CausalFlow.is_well_formed`
	- `XZCorrections.is_well_formed`

- Adapt exisiting methods in the class `OpenGraph`:
	- Adapt `OpenGraph.compose`
	- Adapt `OpenGraph.is_close`

- Methods to replace parametric angles with float values in `PauliFlow`, its children and  `XZCorrections`.

- Method to extract `XZCorrections` from a `Pattern` instance.

- Class method `.from_correction_function` for `PauliFlow` and its children.

- Method to a extract a `GFlow` from a `Pattern` instance. This requires analysis of the existing algorithms in the codebase.  

- Re-introduce benchmark tests on flow algorithms (formerly in `tests.test_find_pflow.py`. 

- Visualization features (optional)
	- Implement pretty print methods for flow and XZ-corrections objects.
	- Introduce a tool to visualize flow objects (similar to the existing tool for pattern visualization).

## References
[1] Mitosek and Backens, 2024 (arXiv:2410.23439).

[2] Mhalla and Perdrix, (2008), Finding Optimal Flows Efficiently, doi.org/10.1007/978-3-540-70575-8_70

[3] Browne et al., 2007 New J. Phys. 9 250 (arXiv:quant-ph/0702212).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-
+- #358: Refactor of flow tools - Part I
+  - New module `graphix.flow.core` which introduces classes `PauliFlow`, `GFlow`, `CausalFlow` and `XZCorrections` allowing a finer analysis of MBQC flows. This module subsumes `graphix.generator` which has been removed and part of `graphix.gflow` which will be removed in the future. 
+  - New module `graphix.flow._find_cflow` with the existing causal-flow finding algorithm.
+  - New module `graphix.flow._find_gpflow` with the existing g- and Pauli-flow finding algorithm introduced in #337.
+  - New abstract types `graphix.fundamentals.AbstractMeasurement` and `graphix.fundamentals.AbstractPlanarMeasurement` which serve as an umbrella of the existing types `graphix.measurements.Measurement`, `graphix.fundamentals.Plane` and `graphix.fundamentals.Axis`.
+  - New method `graphix.pattern.Pattern.extract_opengraph` which subsumes the static method `graphix.opengraph.OpenGraph.from_pattern`.
+  - New methods of `graphix.opengraph.OpenGraph` which allow to extract a causal, g- or Pauli flow.
 ### Fixed
 
 ### Changed
+ - #358: Refactor of flow tools - Part I
+  - API for the `graphix.opengraph.OpenGraph` class:
+    - `OpenGraphs` are parametrically typed so that they can be defined on planes and axes mappings in addition to measurements mappings.
+    - Attribute names are now `graph`, `input_nodes`, `output_nodes` and `measurements`.
+
 
 ## [0.3.3] - 2025-10-23
 


graphix/gflow.py
```diff
@@ -86,12 +86,12 @@ def find_gflow(
...
-        inside=graph,
-        inputs=list(iset),
-        outputs=list(oset),
+        graph=graph,
+        input_nodes=list(iset),
+        output_nodes=list(oset),
         measurements=meas,
     )
-    gf = _find_pflow(og)
+    gf = graphix.find_pflow.find_pflow(og)
```
emlynsg, line 94: I might need this explained, but unclear why we are calling the pflow function for gflow. If this is correct, I think we should reconsider the naming convention.

graphix/opengraph.py
```diff
@@ -1,68 +1,90 @@
...
+from graphix.fundamentals import AbstractMeasurement, AbstractPlanarMeasurement
 from graphix.measurements import Measurement
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable, Mapping
+    from collections.abc import Collection, Iterable, Mapping, Sequence
 
     from graphix.pattern import Pattern
 
+# TODO: Maybe move these definitions to graphix.fundamentals and graphix.measurements ? Now they are redefined in graphix.flow._find_gpflow, not very elegant.
```
emlynsg, line 22: Agree these definitions would make more sense in those files rather than in `opengraph.py`

graphix/gflow.py
matulni: Hi @emlynsg, this file is the old API of the flow utilities of the library. The reason why we call `find_pflow` for `gflow` is that they are found with same algorithm (introduced in #337 ).
Note that this file will be removed in the second part of the PR when all the different parts of the library use the flow functionalities in the new API.
The modifications in `graphix.gflow.py` are just a temporary measure to avoid circular imports and pass the unit tests.

emlynsg: Got it. The new naming conventions are much clearer!

graphix/flow/_find_cflow.py
```diff
@@ -0,0 +1,120 @@
...
+    """Return the causal flow of the input open graph if it exists.
+
+    Parameters
+    ----------
+    og : OpenGraph[_PM_co]
+        Open graph whose causal flow is calculated.
+
+    Returns
+    -------
+    CausalFlow | None
```
thierry-martinez:
```suggestion
    CausalFlow[_PM_co] | None
```

graphix/flow/_find_cflow.py
```diff
@@ -0,0 +1,120 @@
...
+        Nodes which could correct a node at the time of calling the function. This set can never contain input nodes, uncorrected nodes or nodes which already correct another node.
+    cf : dict[int, frozenset[int]]
+        Causal flow correction function. `cf[i]` is the one-qubit set correcting the measurement of qubit `i`.
+    layers : list[frozenset[int]]
+        Partial order between corrected qubits in a layer form. The set `layers[i]` comprises the nodes in layer `i`. Nodes in layer `i` are "larger" in the partial order than nodes in layer `i+1`.
+
+
+    Returns
+    -------
+    CausalFlow | None
```
thierry-martinez:
```suggestion
    CausalFlow[_PM_co] | None
```

graphix/flow/_find_cflow.py
```diff
@@ -0,0 +1,120 @@
...
+    corrected_nodes: AbstractSet[int],
+    corrector_candidates: AbstractSet[int],
+    cf: dict[int, frozenset[int]],
+    layers: list[frozenset[int]],
+) -> CausalFlow[_PM_co] | None:
+    """Find one layer of the causal flow.
+
+    Parameters
+    ----------
+    og : OpenGraph[Plane]
```
thierry-martinez:
```suggestion
    og : OpenGraph[_PM_co]
```

graphix/flow/_find_gpflow.py
```diff
@@ -0,0 +1,668 @@
...
+        for n1, n2 in graph.edges:
+            for u, v in ((n1, n2), (n2, n1)):
+                if u in row_tags and v in col_tags:
+                    i, j = row_tags.index(u), col_tags.index(v)
+                    adj_red[i, j] = 1
+
+        return adj_red
+
+    @cached_property
+    def _compute_og_matrices(self) -> tuple[MatGF2, MatGF2]:
```
thierry-martinez: I would remove the verb for naming a property.

```suggestion
    def _og_matrices(self) -> tuple[MatGF2, MatGF2]:
```

graphix/flow/_find_gpflow.py
```diff
@@ -0,0 +1,668 @@
...
+        order_demand_matrix = flow_demand_matrix.copy()
+
+        inputs_set = set(self.og.input_nodes)
+
+        row_tags = self.non_outputs
+        col_tags = self.non_inputs
+
+        for v in row_tags:  # v is a node tag
+            i = row_tags.index(v)
+            plane_v = self.og.measurements[v].to_plane()
```
thierry-martinez: I think we could reduce the code duplication by adding a method `_get_plane_or_axis(v)` in `AlgebraicOpenGraph` and redefining only this method in `PlanarAlgebraicOpenGraph` to make it call `to_plane`.

graphix/pyzx.py
```diff
@@ -31,7 +31,8 @@ def _fraction_of_angle(angle: ExpressionOrFloat) -> Fraction:
     return Fraction(angle)
 
 
-def to_pyzx_graph(og: OpenGraph) -> BaseGraph[int, tuple[int, int]]:
+# TODO: Adapt to new OpenGrpah APi
```
thierry-martinez:
```suggestion
# TODO: Adapt to new OpenGraph API
```

graphix/opengraph.py
```diff
@@ -1,68 +1,90 @@
...
-        if not all(node in self.inside.nodes for node in self.outputs):
+        if not outputs.issubset(all_nodes):
             raise ValueError("All output nodes must be part of the graph's nodes.")
-        if any(node in self.outputs for node in self.measurements):
-            raise ValueError("Output node cannot be measured.")
-        if len(set(self.inputs)) != len(self.inputs):
+        if outputs & self.measurements.keys():
+            raise ValueError("Output nodes cannot be measured.")
+        if all_nodes - outputs != self.measurements.keys():
+            raise ValueError("All non-ouptut nodes must be measured.")
```
```suggestion
            raise ValueError("All non-output nodes must be measured.")
```

graphix/opengraph.py
```diff
@@ -86,37 +108,169 @@ def isclose(self, other: OpenGraph, rel_tol: float = 1e-09, abs_tol: float = 0.0
...
         meas_angles = pattern.get_angles()
-        meas = {node: Measurement(meas_angles[node], meas_planes[node]) for node in meas_angles}
+        measurements: Mapping[int, Measurement] = {
+            node: Measurement(meas_angles[node], meas_planes[node]) for node in meas_angles
+        }
+
+        return OpenGraph(graph, input_nodes, output_nodes, measurements)
 
-        return OpenGraph(graph, meas, inputs, outputs)
+    def to_pattern(self: OpenGraph[Measurement]) -> Pattern | None:
```
thierry-martinez: I would expect the return type of `to_pattern` to be `Pattern`, and to raise an exception if there is no deterministic pattern instead of returning `None`. If we want a version that doesn’t raise, perhaps we could call it `try_to_pattern`.

graphix/flow/core.py
```diff
@@ -0,0 +1,477 @@
...
+        This method computes the partial order induced by the XZ-corrections.
+        """
+        x_corrections = x_corrections or {}
+        z_corrections = z_corrections or {}
+
+        nodes_set = set(og.graph.nodes)
+        outputs_set = frozenset(og.output_nodes)
+        non_outputs_set = nodes_set - outputs_set
+
+        if not set(x_corrections).issubset(non_outputs_set):
```
thierry-martinez: Nitpicking: `issuperset` allows you to achieve the same result without instantiating a new set.

```suggestion
        if not non_outputs_set.issuperset(x_corrections):
```

graphix/flow/core.py
```diff
@@ -0,0 +1,477 @@
...
+        partial_order_layers = _dag_to_partial_order_layers(dag)
+
+        if partial_order_layers is None:
+            raise ValueError(
+                "Input XZ-corrections are not runnable since the induced directed graph contains closed loops."
+            )
+
+        # If the open graph has outputs, the first element in the output of `_dag_to_partial_order_layers(dag)` may or may not contain a subset of the output nodes.
+        if outputs_set:
+            shift = 1 if partial_order_layers[0].issubset(outputs_set) else 0
```
thierry-martinez: We can have `partial_order_layers == []` while `output_sets == True`: it suffices that there are no corrections, e.g., with a graph state. In this case, `partial_order_layers[0]` raises an `IndexError`.

```python
    def test_output_nodes_without_inner_nodes(self) -> None:
        og = OpenGraph(
            graph=nx.Graph([(0, 1)]),
            input_nodes=[],
            output_nodes=[0, 1],
            measurements={},
        )

        XZCorrections.from_measured_nodes_mapping(og=og)
```

graphix/flow/core.py
```diff
@@ -0,0 +1,477 @@
...
+            partial_order_layers = [outputs_set, *partial_order_layers[shift:]]
```
thierry-martinez: We can have `shift == 0` (i.e., `not partial_order_layers[0].issubset(outputs_set)`) without having `partial_order_layers[0].isdisjoint(outputs_set)`. Consequently, `[outputs_set, *partial_order_layers[shift:]]` is not necessarily a partition (because `outputs_set` and `partial_order_layers[0]` may have elements in common).

```python
    def test_partier_order_layers_partition(self) -> None:
        og = OpenGraph(
            graph=nx.Graph([(0, 1), (2, 3)]),
            input_nodes=[],
            output_nodes=[3],
            measurements=dict.fromkeys([0, 1, 2], Measurement(angle=0, plane=Plane.XY)),
        )
        x_corrections = {0: {1}, 2: {3}}

        corrections = XZCorrections.from_measured_nodes_mapping(og=og, x_corrections=x_corrections)

        assert all(sum(1 for layer in corrections.partial_order_layers if node in layer) == 1 for node in og.graph.nodes)
```

graphix/flow/core.py
```diff
@@ -0,0 +1,477 @@
...
+
+        ordered_nodes = {node for layer in partial_order_layers for node in layer}
```
thierry-martinez:
```suggestion
        ordered_nodes = set.union(*partial_order_layers)
```

graphix/flow/core.py
```diff
@@ -0,0 +1,477 @@
...
+        # We append to the last layer (first measured nodes) all the non-output nodes not involved in the corrections.
+        if unordered_nodes := frozenset(nodes_set - ordered_nodes):
+            partial_order_layers.append(unordered_nodes)
+
+        return XZCorrections(og, x_corrections, z_corrections, tuple(partial_order_layers))
+
+    def to_pattern(
+        self: XZCorrections[Measurement],
+        total_measurement_order: TotalOrder | None = None,
+    ) -> Pattern:
```
thierry-martinez, line 139: This function could return `StandardizedPattern` (provided `StandardizedPattern` becomes a mere dataclass, and its current constructor becomes a static method `from_pattern`).

Another strategy could be to return a space-optimal pattern for the given measurement order (using the algorithm implemented in `Pattern._reorder_pattern`). This could be done via `StandardizedPattern` and #360.

graphix/flow/core.py
```diff
@@ -0,0 +1,477 @@
...
+    -----
+    See :func:`XZCorrections.extract_dag`.
+    """
+    relations: set[tuple[int, int]] = set()
+
+    for measured_node, corrected_nodes in x_corrections.items():
+        relations.update(product([measured_node], corrected_nodes))
+
+    for measured_node, corrected_nodes in z_corrections.items():
+        relations.update(product([measured_node], corrected_nodes))
```
thierry-martinez:
```suggestion
    relations = (
        (measured_node, corrected_node)
        for corrections in (x_corrections, z_corrections)
        for measured_node, corrected_nodes in corrections.items()
        for corrected_node in corrected_nodes
    )
```

graphix/flow/core.py
matulni: Nice, I didn't know that idiom, thanks!

graphix/flow/core.py
matulni: Very nice!

graphix/opengraph.py
matulni: Sounds fine, I don't have a strong preference for any of the options. Which one do prefer?

graphix/flow/core.py
matulni: This seems to be problematic because elements of `partial_order_layers` are of type `frozenset`. The motivation was to enforce inadvertent modifications, but maybe it's not a good idea as it appears to hinder usability. I revert your suggestion so that tests don't fail, but we can discuss it.

graphix/flow/_find_gpflow.py
matulni: Done in [6e7d36b](https://github.com/TeamGraphix/graphix/pull/358/commits/6e7d36bbc946eb0def5d40e10bacf7a1b6e3fd03)

graphix/flow/_find_gpflow.py
matulni: Done in [6e7d36b](https://github.com/TeamGraphix/graphix/pull/358/commits/6e7d36bbc946eb0def5d40e10bacf7a1b6e3fd03)

graphix/flow/core.py
thierry-martinez: Sorry for the wrong suggestion; the first element of `partial_order_layers` is a `frozenset`, so we should use `frozenset.union`.

graphix/flow/core.py
matulni: Fixed in [3f394a5](https://github.com/TeamGraphix/graphix/pull/358/commits/3f394a5287e9608e7a32c344315cceda86e28aaf)

graphix/flow/core.py
matulni: Fixed in [3f394a5](https://github.com/TeamGraphix/graphix/pull/358/commits/3f394a5287e9608e7a32c344315cceda86e28aaf)

graphix/flow/core.py
matulni: Thanks, done in [b3a82c2](https://github.com/TeamGraphix/graphix/pull/358/commits/b3a82c2a044f0ac21efdb10766032fa8271f5497)

graphix/opengraph.py
thierry-martinez: I think `to_pattern` is more useful and idiomatic. We could add `try_to_pattern` afterwards if we notice it will be useful.

graphix/flow/core.py
thierry-martinez: See #361.

graphix/flow/core.py
```diff
@@ -0,0 +1,488 @@
...
+            partial_order_layers = [outputs_set] if outputs_set else []
+            if non_outputs_set:
+                partial_order_layers.append(frozenset(non_outputs_set))
+            return XZCorrections(og, x_corrections, z_corrections, tuple(partial_order_layers))
+
+        # If the open graph has outputs, the first element in the output of `_dag_to_partial_order_layers(dag)` may or may not contain output nodes.
+        if outputs_set:
+            partial_order_layers = [
+                outputs_set,
+                frozenset(partial_order_layers[0] - outputs_set),
```
thierry-martinez: This set can be empty:
```python
    def test_partial_order_layers_partition2(self) -> None:
        og = OpenGraph(
            graph=nx.Graph([(0, 1), (2, 3)]),
            input_nodes=[],
            output_nodes=[1, 3],
            measurements=dict.fromkeys([0, 2], Measurement(angle=0, plane=Plane.XY)),
        )
        x_corrections = {0: {1}, 2: {3}}

        corrections = XZCorrections.from_measured_nodes_mapping(og=og, x_corrections=x_corrections)

        assert all(corrections.partial_order_layers)
        assert all(
            sum(1 for layer in corrections.partial_order_layers if node in layer) == 1 for node in og.graph.nodes
        )
```

graphix/flow/core.py
```diff
@@ -0,0 +1,488 @@
...
+            ]
+
+        ordered_nodes = frozenset.union(*partial_order_layers)
+
+        if not ordered_nodes.issubset(nodes_set):
+            raise ValueError("Values of input mapping contain labels which are not nodes of the input open graph.")
+
+        # We append to the last layer (first measured nodes) all the non-output nodes not involved in the corrections.
+        if unordered_nodes := frozenset(nodes_set - ordered_nodes):
+            partial_order_layers.append(unordered_nodes)
```
thierry-martinez: Actually, `unordered_nodes` can be added to any layer, they do not need to be put in their own layer. I would say that if `partial_order_layers != []` then `partial_order_layers[-1] = partial_order_layers[-1] | unordered_nodes`.

graphix/flow/core.py
matulni: I see `StandardizedPattern` as an internal representation rather than the final output of a calculation. I think it makes more sense to return a space-optimal pattern for the given measurement order. I'm happy to modify it accordingly once #361 is merged (which looks good to go for me).

graphix/flow/core.py
matulni: Fixed (hopefully) in [0fe3113](https://github.com/TeamGraphix/graphix/pull/358/commits/0fe31139d83e10e2cf6fc9738ddffacb6b0d3791)

graphix/flow/core.py
matulni: See [0fe3113](https://github.com/TeamGraphix/graphix/pull/358/commits/0fe31139d83e10e2cf6fc9738ddffacb6b0d3791)

matulni:
Commit [a7c6778](https://github.com/TeamGraphix/graphix/pull/358/commits/a7c677867ed82288022405d82cdd264d10e67c4e) removes the static method `OpenGraph.from_pattern` and introduces `Pattern.extract_opengraph` which has the same logic. Methods `Pattern.extract_graph`, `Pattern.extract_nodes`, etc. call `Pattern.extract_opengraph` reducing code repetition.

graphix/opengraph.py
matulni: Done in [3b4a2b8](https://github.com/TeamGraphix/graphix/pull/358/commits/3b4a2b8e4687f9b8100d1ab38890533a4f30c2b2)

graphix/pattern.py
```diff
@@ -1113,36 +1135,25 @@ def extract_graph(self) -> nx.Graph[int]:
...
-        return nodes
+        """Return the nodes of the pattern.
+
+        Returns
+        -------
+        set[int]
+        """
+        return set(self.extract_graph().nodes)
 
     def extract_isolated_nodes(self) -> set[int]:
```
thierry-martinez: This change has a non-negligible impact on performances.

In the following benchmark, I compare the former implementation of `extract_nodes` and `extract_graph` with the new `extract_opengraph`, and I propose two other implementations `extract_graph_simple` (which does not build a networkx graph) and `extract_graph_measurements` (which collect measurements in addition to the graph) for comparison. I am afraid that this is a situation where some code duplication is inevitable due to bad Python performance with high-level data structures. I also tested replacing `tuple[ExpressionOrFloat, Plane]` with `Measurement`, and the performance loss is around 13%.

The performance table:
```python
timeit(pattern.extract_opengraph, number=10000)=4.035559833049774
timeit(pattern.extract_graph, number=10000)=3.1880758749321103
timeit(pattern.extract_graph_simple, number=10000)=1.162953041959554
timeit(pattern.extract_graph_measurements, number=10000)=1.5968126250663772
timeit(pattern.extract_nodes, number=10000)=0.5324231670238078
```

The code for `extract_graph_simple` and `extract_graph_measurements`:
```python
    def extract_graph_simple(self) -> tuple[list[int], set[tuple[int, int]]]:
        """Return the graph state from the command sequence, extracted from 'N' and 'E' commands.

        Returns
        -------
        graph_state: nx.Graph[int]
        """
        nodes = []
        edges = set()
        for cmd in self.__seq:
            if cmd.kind == CommandKind.N:
                nodes.append(cmd.node)
            elif cmd.kind == CommandKind.E:
                u, v = cmd.nodes
                if u > v:
                    u, v = v, u
                edges.symmetric_difference_update({(u, v)})
        return nodes, edges

    def extract_graph_measurements(self) -> tuple[list[int], set[tuple[int, int]], dict[int, tuple[ExpressionOrFloat, Plane]]]:
        """Return the graph state from the command sequence, extracted from 'N' and 'E' commands, together with the measurement datas.

        Returns
        -------
        graph_state: nx.Graph[int]
        """
        nodes = []
        edges = set()
        measurements: dict[int, cmd.M] = {}
        for cmd in self.__seq:
            if cmd.kind == CommandKind.N:
                nodes.append(cmd.node)
            elif cmd.kind == CommandKind.E:
                u, v = cmd.nodes
                if u > v:
                    u, v = v, u
                edges.symmetric_difference_update({(u, v)})
                measurements[cmd.node] = (cmd.angle, cmd.plane)
        return nodes, edges, measurements
```

The code for benchmarking:
```python
from timeit import timeit
from graphix.random_objects import rand_circuit
import numpy as np

rng = np.random.default_rng(42)
circuit = rand_circuit(10, 10, rng)

print(f"{timeit(pattern.extract_opengraph, number=10000)=}")
print(f"{timeit(pattern.extract_graph, number=10000)=}")
print(f"{timeit(pattern.extract_graph_simple, number=10000)=}")
print(f"{timeit(pattern.extract_graph_measurements, number=10000)=}")
print(f"{timeit(pattern.extract_nodes, number=10000)=}")
```

matulni: Hi @thierry-martinez, thanks for taking the time to do the benchmark. In view of your analysis, I propose:
- `extract_graph` -> returns the graph as  a tuple of sets (nodes and edges)
- `extract_opengraph` -> returns the open graph object (instead of the  constructor `OpenGraph.from_pattern`)
- `extract_measurements` -> returns a dictionary of measurements only (iterating over the pattern, without calling `extract_opengraph`
-  `extract_planes` and `extract_angles` -> call `self.extract_measurements` and iterate over the result. This incurs in an unnecessary iteration over the measurements but makes the code more succinct

This way we have a lightweight representation of the graph resource state and also access to the nx.Graph object via `OpenGraph`.

If you agree, I'll implement this in a new commit.

thierry-martinez: I agree for `extract_graph` and `extract_opengraph`. In addition, I would keep `extract_nodes` as is. For `extract_planes` and `extract_angles`, I would say that either we remove these methods (as users can make the projection themselves if needed), or we provide dedicated implementations that do not call `extract_measurements`.

matulni: Reverted in [235ab71](https://github.com/TeamGraphix/graphix/pull/358/commits/235ab71ce17ed9171d9b8ee278717816def6d099)

graphix/opengraph.py
matulni: Hi @thierry-martinez , following this logic, should we replace `find_causal_flow` by `extract_causal_flow`, etc. and raise an exception instead of returning `None` ?

I think this is better design because otherwise statements like `og.find_gflow().to_corrections().to_pattern()` yield a typing error by the static analyzer (because `None` does not have `.to_corrections` method).

thierry-martinez: Indeed, I think that raising an exception is more idiomatic and useful than returning `None` in particular for the kinds of dot chains you are mentioning.

matulni: Done in [0614633](https://github.com/TeamGraphix/graphix/pull/358/commits/06146333075568cba90ddf3fcc1d2b653f4ffc70)

graphix/fundamentals.py
```diff
@@ -317,3 +368,22 @@ def from_axes(a: Axis, b: Axis) -> Plane:
...
+    def to_plane_or_axis(self) -> Plane:
+        """Return the plane.
+
+        Returns
+        -------
+        Plane
+        """
+        return self
+
+    def to_plane(self) -> Plane:
```
thierry-martinez, line 383: Add `@override`.

graphix/opengraph.py
```diff
@@ -85,38 +108,171 @@ def isclose(self, other: OpenGraph, rel_tol: float = 1e-09, abs_tol: float = 0.0
...
+
+        References
+        ----------
+        [1] Browne et al., NJP 9, 250 (2007)
+        """
+        for extractor in (self.extract_causal_flow, self.extract_pauli_flow):
+            try:
+                return extractor().to_corrections().to_pattern()
+            except OpenGraphError:  # noqa: PERF203
+                continue
```
thierry-martinez: Sorry, I was not expecting to go that far about preferring raising an exception instead of returning `None`:
- `extract_causal_flow` and `extract_pauli_flow` now need to raise an `OpenGraphError` from the `None` value returned by `find_` functions, for the exception to be immediately caught here. We have the feeling we would be better off testing for `None` here and not having to coin an error message in the `extract_` functions;
- the scope of the `try`-`except` is a bit too large: we only catch exceptions in `extractor`; `to_corrections().to_pattern()` should go in the `else:` clause of the `try`, but a simple `if` test for `None` is more readable;
- I think that the efficiency cost that `PERF203` pretends to avoid is negligible here, but if we can avoid ignoring this rule for free, it is even better!

tests/test_opengraph.py
```diff
@@ -1,50 +1,621 @@
...
+        results = []
+
+        for plane in {Plane.XY, Plane.XZ, Plane.YZ}:
+            alpha = 2 * np.pi * fx_rng.random()
+            state_ref = pattern_ref.simulate_pattern(input_state=PlanarState(plane, alpha))
+            state = pattern.simulate_pattern(input_state=PlanarState(plane, alpha))
+            results.append(np.abs(np.dot(state.flatten().conjugate(), state_ref.flatten())))
+
+        avg = sum(results) / 3
+        assert avg == pytest.approx(1)
```
thierry-martinez: Why is the mean relevant here ? Couldn't we just do the following?



```suggestion
            assert np.abs(np.dot(state.flatten().conjugate(), state_ref.flatten())) == pytest.approx(1)
```

tests/test_opengraph.py
```diff
@@ -1,50 +1,621 @@
...
+        },
+    )
+    return OpenGraphFlowTestCase(og, has_cflow=False, has_gflow=False, has_pflow=True)
+
+
+def prepare_test_og_flow() -> list[OpenGraphFlowTestCase]:
+    n_og_samples = 20
+    test_cases: list[OpenGraphFlowTestCase] = [globals()[f"_og_{i}"]() for i in range(n_og_samples)]
+
+    return test_cases
```
thierry-martinez: I would avoid playing with dynamic lookup of functions by names. I think this might be a better way of doing this: https://github.com/thierry-martinez/graphix/commit/555600e736c3f008a34101e465889e139a1e954e

tests/test_opengraph.py
matulni: Indeed it's not, I copied and (only partially) modified from the determinism check function where various simulations are combined into a single bool (where is not necessary either).

graphix/fundamentals.py
matulni: Fixed in [c4111fb](https://github.com/TeamGraphix/graphix/pull/358/commits/c4111fb7b35e5742be71525986d150fa477d1a66)

tests/test_opengraph.py
matulni: Fixed in [c4111fb](https://github.com/TeamGraphix/graphix/pull/358/commits/c4111fb7b35e5742be71525986d150fa477d1a66)

graphix/opengraph.py
matulni: Fixed in [0c21256](https://github.com/TeamGraphix/graphix/pull/358/commits/0c2125652c2a78edca1afda16aaba032ffa634b0)

tests/test_opengraph.py
matulni: Thanks! Indeed, much better, this way we can use these tests in other modules. Merged it in [339ba1c](https://github.com/TeamGraphix/graphix/pull/358/commits/339ba1c0888d5864cd4eb3f1498f8ce72eda9435)

matulni:
Hi @thierry-martinez, I've being thinking and I'm not fully convinced by [0c21256](https://github.com/TeamGraphix/graphix/pull/358/commits/0c2125652c2a78edca1afda16aaba032ffa634b0)

By introducing the new functions `find_gflow`, `find_pflow` to generate flows without catching exceptions, we break a bit the object-oriented approach. The user may encounter this need too (e.g., generating random open graphs which may or may not have flow inside a loop), and recurring to `find_gflow`, `find_pflow` hidden in another module feels a bit hacky. 

In addition, these new functions lead to a circular import structure (which, if not a red boundary on its own, can be symptomatic of bad design).

At the same time, as discussed above, having functions which return one type only is very handy for chaining method calls without getting static-typing errors.

Should we add two methods to `OpenGraph` (one with two return types and one that raises exceptions) ? I'm also not a big fan of this idea because it seems overly complex, but maybe it's justified ?

(Sorry to bring this up after your pass!)

thierry-martinez:
I think that having both `find_` and `extract_` methods in `OpenGraph` is more symmetrical indeed.

PR #361, opened 2025-11-12
Fix #360: `StandardizedPattern` as dataclass, `to_space_optimal_pattern`

### LINKED ISSUE #360: Expose `Pattern._reorder_pattern` to obtain a space-optimal from a `StandardizedPattern`
The private function `Pattern._reorder_pattern` returns a space-optimal pattern given a measurement order and a pattern. This algorithm is useful enough to deserve to be public. It could be exposed as `StandardizedPattern.to_space_optimal_pattern`.

thierry-martinez:
This commit makes `StandardizedPattern` a frozen dataclass with immutable data structures (tuples and `MappingProxyType`).

`StandardizedPattern` no longer requires construction from an initial pattern: new fields `input_nodes`, `output_nodes` and `results` embed the values that were used from the original pattern.

The new method `to_space_optimal_pattern` returns a pattern that is space-optimal for the measurement order fixed in the dataclass.  This new method subsumes the implementation that was performed in `Pattern._reorder_pattern`: the latter now uses
`to_space_optimal_pattern` and replaces the sequence of commands in-place. The new implementation preserves the angles of the `N` commands if any, while they were discarded by the previous implementation.


graphix/optimization.py
```diff
@@ -161,17 +207,53 @@ def extract_graph(self) -> nx.Graph[int]:
...
+    def to_space_optimal_pattern(self) -> Pattern:
+        """Return a pattern that is space-optimal for the given measurement order."""
+        pattern = graphix.pattern.Pattern(input_nodes=self.input_nodes)
+        pattern.results = dict(self.results)
+        active = set(self.input_nodes)
+        done: set[Node] = set()
+        n_dict = {n.node: n for n in self.n_list}
+        graph = self.extract_graph()
+
+        def ensure_active(node: Node) -> None:
```
matulni, line 248:
```suggestion
        def ensure_active(node: Node) -> None:
            """Initialize node in pattern if it has not been initialized before."""
```

graphix/optimization.py
```diff
@@ -161,17 +207,53 @@ def extract_graph(self) -> nx.Graph[int]:
...
+            if node not in active:
+                pattern.add(n_dict[node])
+                active.add(node)
+
+        def ensure_neighborhood(node: Node) -> None:
```
matulni, line 254:
```suggestion
        def ensure_neighborhood(node: Node) -> None:
            """Initialize and entangle the inactive nodes in the neighbourhood of ``node``."""
```

graphix/pattern.py
matulni, line 1279: Some comment (maybe for a future PR/the flow refactor 2):
If all these operations require the pattern to be in standard form, wouldn't it make sense that they are methods of the `StandardizedPattern` class?

graphix/optimization.py
```diff
@@ -45,86 +48,126 @@ def standardize(pattern: Pattern) -> Pattern:
...
+    results: MappingProxyType[Node, Outcome]
+    n_list: tuple[command.N, ...]
+    e_set: frozenset[frozenset[Node]]
+    m_list: tuple[command.M, ...]
+    c_dict: MappingProxyType[Node, Clifford]
+    z_dict: MappingProxyType[Node, frozenset[Node]]
+    x_dict: MappingProxyType[Node, frozenset[Node]]
+
+    @classmethod
+    def new(
```
matulni: If in the end we opt for this strategy to combine frozen data classes, immutable attributes and a general constructor, I'd suggest using `from_data` instead of `new` to avoid confusion with the `__new__` method, and establish some parallelism with `from_pattern`, etc. We can adopt this convention in `OpenGraph`, flows, etc.

graphix/pattern.py
thierry-martinez: I think these functions can indeed be refactored in a later PR.

graphix/optimization.py
thierry-martinez: I proposed another approach in 5bba746: instead of relying on a specific class method, I introduced the frozen dataclass into a dedicated superclass, and I used the `__init__` method to initialize the fields while calling `super().__init__()`.

thierry-martinez:
@matulni:

1. Unfortunately, in the absence of `frozendict` (PEP 416), there is no type annotation that can guarantee that a `Mapping` is immutable (even if a value is of type `MappingProxyType`, the underlying dictionary can have its contents change over time). Therefore, if we had a no-copy initialization, we would be able to build an instance whose `Mapping` fields could change over time. Note that for `tuple` and `frozenset`, there is already no copy made if the provided values are already of the correct type (`tuple(x)` and `frozenset(x)` are no-ops if `x` is already a tuple or a frozen set, respectively).
2. Yes, we can add new fields to `StandardizedPattern` instances, but the guarantee that the fields in the class interface do not change still holds. We could also prevent the insertion of new fields and be slightly more efficient by using `slots=True` in the dataclass definition, but this requires Python 3.10 (PR #371).

thierry-martinez:
We can use `slots=True` (ce86f1c), now that #371 is merged.

PR #362, opened 2025-11-13
Fix #349: ensure flow for patterns transpiled from circuit

### LINKED ISSUE #349: [Bug]: `pattern.draw` fails to find flow or flow
**Bug description**
The method `Pattern.draw_graph` fails to find a flow on a pattern transpiled from circuit if the parameter `flow_from_pattern` is set to `True` (default behaviour).

If `flow_from_pattern == False`, the flow search is done on the underlying open graph and the function correctly finds a flow.

**To Reproduce**
```python
circuit = Circuit(2)
circuit.cnot(0, 1)
pattern.standardize()
print(pattern) # X(3,{2}) Z(0,{1}) Z(3,{1}) M(2) M(1) E(1,2) E(2,3) E(0,2) N(3) N(2)
pattern.draw_graph() # The pattern is not consistent with flow or gflow structure.
pattern.draw_graph(flow_from_pattern=False) # Flow detected in the graph.
```

**Additional comments**

This issue goes back at least to 73bac1ff (unable to find a particular commit which might have broken this functionality).

thierry-martinez:
This commit fixes domains in the transpiler so that every pattern transpiled from a circuit has a flow.  Previously, some domains were incompatible with any flow, despite the patterns being deterministic, because some measurement angles were zero, causing the measurements to ignore certain signals.

This commit also adds `optimization.remove_useless_domains` to remove the domains ignored by measurements with angle zero, to recover the same "optimized" patterns as before when needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- #347:
+- #374:
   - Introduced new method `graphix.opengraph.OpenGraph.is_equal_structurally` which compares the underlying structure of two open graphs.
   - Added new method `isclose` to `graphix.fundamentals.AbstractMeasurement` which defaults to `==` comparison.
 
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - #349, #362: Patterns transpiled from circuits always have causal flow.
 
 ### Changed
+
+- #374: Adapted existing method `graphix.opengraph.OpenGraph.isclose` to the new API introduced in #358.
+
+- #375: Adapted existing method `graphix.opengraph.OpenGraph.compose` to the new API introduced in #358.
 
 ## [0.3.3] - 2025-10-23
 
--- a/graphix/opengraph.py
+++ b/graphix/opengraph.py
@@ -360,25 +360,22 @@ class OpenGraph(Generic[_M_co]):
             correction_matrix
         )  # The constructor returns `None` if the correction matrix is not compatible with any partial order on the open graph.
 
-    # TODO: Generalise `compose` to any type of OpenGraph
-    def compose(
-        self: OpenGraph[Measurement], other: OpenGraph[Measurement], mapping: Mapping[int, int]
-    ) -> tuple[OpenGraph[Measurement], dict[int, int]]:
-        r"""Compose two open graphs by merging subsets of nodes from `self` and `other`, and relabeling the nodes of `other` that were not merged.
+    def compose(self, other: OpenGraph[_M_co], mapping: Mapping[int, int]) -> tuple[OpenGraph[_M_co], dict[int, int]]:
+        r"""Compose two open graphs by merging subsets of nodes from ``self`` and ``other``, and relabeling the nodes of ``other`` that were not merged.
 
         Parameters
         ----------
-        other : OpenGraph
-            Open graph to be composed with `self`.
+        other : OpenGraph[_M_co]
+            Open graph to be composed with ``self``.
         mapping: dict[int, int]
-            Partial relabelling of the nodes in `other`, with `keys` and `values` denoting the old and new node labels, respectively.
+            Partial relabelling of the nodes in ``other``, with ``keys`` and ``values`` denoting the old and new node labels, respectively.
 
         Returns
         -------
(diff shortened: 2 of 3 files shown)


graphix/random_objects.py
```diff
@@ -426,3 +426,23 @@ def rand_circuit(
...
+        The power of 2 for the vector size
+
+    Returns
+    -------
+    numpy.ndarray
+        Normalized complex vector of size 2^nqubits
+    """
+    rng = ensure_rng(rng)
+    dim = 1 << nqubits  # 2**nqubits is typed Any
+    vec = (rng.random(dim) + 1j * rng.random(dim)).astype(np.complex128)
```
matulni: This only generates vectors with a phase $\theta \in [0, \pi/2]$. Could it be an issue ?
I think at least we should indicate in the docstring.

tests/test_transpiler.py
```diff
@@ -156,27 +178,21 @@ def test_add_extend(self) -> None:
...
+        assert f is not None
+
+    @pytest.mark.parametrize("jumps", range(1, 11))
+    @pytest.mark.parametrize("instruction", INSTRUCTION_TEST_CASES)
+    def test_instructions(self, fx_bg: PCG64, jumps: int, instruction: InstructionTestCase) -> None:
+        rng = Generator(fx_bg.jumped(jumps))
+        circuit = Circuit(3, instr=[instruction(rng)])
+        pattern = circuit.transpile().pattern
+        input_state = rand_state_vector(3, rng=rng)
+        state = circuit.simulate_statevector(input_state=input_state).statevec
```
matulni, line 196:
```suggestion
        state = circuit.simulate_statevector(input_state=input_state, rng=rng).statevec
```

tests/test_optimization.py
```diff
@@ -86,6 +86,22 @@ def test_flow_after_pauli_preprocessing(fx_bg: PCG64, jumps: int) -> None:
     assert f is not None
 
 
+@pytest.mark.parametrize("jumps", range(1, 11))
+def test_remove_useless_domains(fx_bg: PCG64, jumps: int) -> None:
+    rng = Generator(fx_bg.jumped(jumps))
+    nqubits = 3
+    depth = 3
+    circuit = rand_circuit(nqubits, depth, rng)
+    pattern = circuit.transpile().pattern
```
matulni, line 95: Maybe nitpicking, but I think that the transpiler only outputs patterns with measurements on the XY plane, so I think this test does not verify the suppression of `t_domain`.

graphix/optimization.py
```diff
@@ -460,3 +461,23 @@ def incorporate_pauli_results(pattern: Pattern) -> Pattern:
             result.add(cmd)
     result.reorder_output_nodes(pattern.output_nodes)
     return result
+
+
+def remove_useless_domains(pattern: Pattern) -> Pattern:
+    """Return an equivalent pattern where results from Pauli presimulation are integrated in corrections."""
```
matulni: I'm a bit confused by the docstring. This function does not look into `Pattern.results` so why does it say "results from Pauli presimulation" ? Wouldn't it be:

`"""Return an equivalent pattern where redundant domains of zero-angle measurements are removed."""`

Additionally, would there be some interest in having this as a method of `Pattern` ?

graphix/random_objects.py
thierry-martinez: I fixed the implementation `rand_state_vector` and added some tests for it in 2effb1d.

tests/test_transpiler.py
thierry-martinez: I think `simulate_statevector` only uses the random number generator for measurement instructions, and we don't test such instructions here.

tests/test_optimization.py
thierry-martinez: Non-trivial `t_domain` can appear after Pauli presimulation.

graphix/optimization.py
thierry-martinez: Oops, I just copy-pasted it from `incorporate_pauli_results` and forgot to change the docstring. I fixed the docstring in a57148e.

We could add a method to `Pattern` to call this, but I tend to prefer putting data-structure-related code in `pattern.py` and algorithmic code in `optimization.py`.

tests/test_optimization.py
matulni: oh, yes you are right!

matulni (approved):
Thanks for the edits, LGTM!

PR #364, opened 2025-11-13
Fix #193: add `Pattern.check_runnability`

### LINKED ISSUE #193: Secure pattern generation
A clear and concise description of what the problem is.

Graphix does not guarantee the runnability and deterministic properties of a pattern when it is passed to a simulator or QPU. It would be beneficial if these fundamanetal properties are ensured during the pattern generation process.

One possible way to do this is to generate a pattern from the underlying graph and flowlike objects(including causal flow, gflow, and pauli flow). Flows can be understood as a feedforward protocol that ensures deterministic property of MBQC.

1. Generalize the graph-based pattern generator(`generator.generate_from_graph`) with an arbitrary flowlike object(related to #120 )
2. Generate every pattern through the above generator
3. Implement flags in the pattern class such as `pattern.is_deterministic` and `pattern.is_runnable`.

I plan to refactor some methods in the pattern class to clarify which operations preserve the properties. For example, `standardize` and `shift_signals` preserve the properties, but `add` does not. I will implement a property-preserving pattern concatenation method for those who want to directly edit the pattern. 

**Additional context**

thierry-martinez:
This commit adds the method `Pattern.check_runnability` that ensures a pattern is runnable. This covers #193, given that determinism is already checkable with Pauli flow finding, and pattern concatenation is already implemented by `Pattern.compose`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,9 +24,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- #364: `Pattern.simulate_pattern`, `Pattern.standardize`,
-  `Pattern.perform_pauli_measurements`, `Pattern.minimize_space`,
-  `Pattern.get_layers` check that the pattern is runnable beforehand.
+- #364: `Pattern.simulate_pattern`, `Pattern.shift_signals`,
+  `Pattern.standardize`, `Pattern.perform_pauli_measurements`,
+  `Pattern.minimize_space`, `Pattern.get_layers` check that the
+  pattern is runnable beforehand.
+
+- #364: `Pattern.compute_max_degree` and `Pattern.draw_graph` no longer
+  fail on empty patterns.
+
 
 ### Changed
 
@@ -36,6 +41,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - Attribute names are now `graph`, `input_nodes`, `output_nodes` and `measurements`.
 
 - #361: `StandardizedPattern` is now an immutable dataclass. The class method `StandardizedPattern.from_pattern` instantiates a `StandardizedPattern` from `Pattern`.
+
+- #364: `StandardizedPattern.perform_pauli_pushing` replaces `Pattern.move_pauli_measurements_to_the_front`.
 
 - #371: Drop support for Python 3.9
 


graphix/pattern.py
```diff
@@ -135,10 +125,12 @@ def add(self, cmd: Command) -> None:
...
             if cmd.node in self.__output_nodes:
-                raise NodeAlreadyPreparedError(cmd.node)
+                raise RunnabilityError(cmd, cmd.node, RunnabilityErrorReason.AlreadyActive)
             self.__n_node += 1
             self.__output_nodes.append(cmd.node)
         elif cmd.kind == CommandKind.M:
+            if cmd.node not in self.__output_nodes:
+                raise RunnabilityError(cmd, cmd.node, RunnabilityErrorReason.AlreadyMeasured)
             self.__output_nodes.remove(cmd.node)
         self.__seq.append(cmd)
```
matulni, line 132: I would add a comment emphasising that this method dos not ensure a secure pattern generation (which has to be verified with `:func:Pattern.check_runnability`.

Some minor comment: if we add a measurement command on a node `i` which is not in the pattern, we get an error message `M(i): node i is already measured.` which is a bit misleading. One way to fix it could be to store the existing nodes in the pattern on the fly (which would also allow to ensure that we can only add `E` and `C` commands on existing nodes).

graphix/pattern.py
matulni, line 1: One general comment: should `Pattern.check_runnability` raise an exception on empty patterns? (`p = Pattern()`). Probably not because  `p.simulate_pattern()` returns 1 (which seems a reasonable convention), so empty patterns are runnable, but I mention this because other methods such as `p.draw_graph` and `p.compute_max_degree` raise cryptic `ValueErrors`.

(For context, I encountered this question of how to deal with empty patterns when writing the methods to extract flows from patterns).

emlynsg: I agree that a pattern is runnable when empty, but maybe we should display a warning in this case.

graphix/pattern.py
emlynsg: Checking N and M here means that we don't run into problems with `extract_graph`, `space_list` and similar methods, but I do think that there may be some niche use for allowing unrunnable patterns (not sure what though).
If we leave this as is, I would suggest adding `check_runnability` to methods associated with using the pattern for computation (or at the very least a flag `_is_runnable` that is checked when using these methods that may display a warning). Specifically `simulate_pattern` and `perform_pauli_measurements` but may be worthwhile (I'm not confident there won't be errors, so I think it's worth checking) in `minimize_space`, `parallelize_pattern` and `shift_signals`.

thierry-martinez: In 49616d8, I removed all the partial runnability checks from `Pattern.add`: I think we would have the choice either to perform a full runnability check in `add` to guarantee that `Pattern` objects are always runnable or, as it is done for now, to make no such guarantee and to provide a runnability checker instead, but having only a partial test was awkward. Moreover, having `add` implementing a part of the runnability check makes it difficult to write tests for `check_runnability` that were already handled by `add`, so it is more convenient to be able to let non-runnable patterns being built.

Since standardization and signal shifting could have turned non-runnable patterns into runnable ones (and hence could hide some code-logic errors), I added a call to `check_runnability` at the beginning of standardization and signal shifting, so in particular all methods that require the pattern to be standardized first check for runnability (`minimize_space`, `parallelize_pattern`, `perform_pauli_measurements`), and I added a check in `simulate_pattern` too, so that error messages are clearer and raised before starting the simulation.

graphix/pattern.py
thierry-martinez: `draw_graph` and `compute_max_degree` now work with empty patterns.

PR #369, opened 2025-11-17
Enable type-checking for `random_objects.py`

thierry-martinez:
This commit adds type annotations to `random_objects.py`, along with its corresponding tests, enabling full type-checking with mypy while preserving existing functionality.

The exclusion of `device_interface.py` is removed from `pyproject.toml` since this module was removed in #261.

`scipy-stubs` is added to `requirements-dev.txt` since `scipy` is used in `random_objects.py`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,42 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- #358: Refactor of flow tools - Part I
-  - New module `graphix.flow.core` which introduces classes `PauliFlow`, `GFlow`, `CausalFlow` and `XZCorrections` allowing a finer analysis of MBQC flows. This module subsumes `graphix.generator` which has been removed and part of `graphix.gflow` which will be removed in the future. 
-  - New module `graphix.flow._find_cflow` with the existing causal-flow finding algorithm.
-  - New module `graphix.flow._find_gpflow` with the existing g- and Pauli-flow finding algorithm introduced in #337.
-  - New abstract types `graphix.fundamentals.AbstractMeasurement` and `graphix.fundamentals.AbstractPlanarMeasurement` which serve as an umbrella of the existing types `graphix.measurements.Measurement`, `graphix.fundamentals.Plane` and `graphix.fundamentals.Axis`.
-  - New method `graphix.pattern.Pattern.extract_opengraph` which subsumes the static method `graphix.opengraph.OpenGraph.from_pattern`.
-  - New methods of `graphix.opengraph.OpenGraph` which allow to extract a causal, g- or Pauli flow.
-
-- #360, #361: `StandardizedPattern.to_space_optimal_pattern` generates
-  a pattern that is space-optimal for a given measurement order.
-
-- #193, #364: `Pattern.check_runnability` ensures that a pattern is runnable.
-
 ### Fixed
 
-- #364: `Pattern.simulate_pattern`, `Pattern.shift_signals`,
-  `Pattern.standardize`, `Pattern.perform_pauli_measurements`,
-  `Pattern.minimize_space`, `Pattern.get_layers` check that the
-  pattern is runnable beforehand.
-
-- #364: `Pattern.compute_max_degree` and `Pattern.draw_graph` no longer
-  fail on empty patterns.
-
-
 ### Changed
-
-- #358: Refactor of flow tools - Part I
-  - API for the `graphix.opengraph.OpenGraph` class:
-    - `OpenGraphs` are parametrically typed so that they can be defined on planes and axes mappings in addition to measurements mappings.
-    - Attribute names are now `graph`, `input_nodes`, `output_nodes` and `measurements`.
-
-- #361: `StandardizedPattern` is now an immutable dataclass. The class method `StandardizedPattern.from_pattern` instantiates a `StandardizedPattern` from `Pattern`.
-
-- #364: `StandardizedPattern.perform_pauli_pushing` replaces `Pattern.move_pauli_measurements_to_the_front`.
-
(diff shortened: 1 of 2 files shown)


graphix/random_objects.py
```diff
@@ -16,21 +17,24 @@
...
+
+def rand_herm(sz: IntLike, rng: Generator | None = None) -> npt.NDArray[np.complex128]:
     """Generate random hermitian matrix of size sz*sz."""
     rng = ensure_rng(rng)
     tmp = rng.random(size=(sz, sz)) + 1j * rng.random(size=(sz, sz))
     return tmp + tmp.conj().T
 
 
-def rand_unit(sz: int, rng: Generator | None = None) -> npt.NDArray[np.complex128]:
+def rand_unit(sz: IntLike, rng: Generator | None = None) -> npt.NDArray[np.float64]:
```
matulni: Why `np.float64` instead of `np.complex128` ?

graphix/random_objects.py
matulni:
```suggestion
   sig : float | Literal["ginibre"] | None
       Standard deviation of random variates.
       ``sig = "ginibre"`` draws from the Ginibre ensemble, i.e.,  ``sig = 1 / sqrt(2 * dim)``.
       ``None`` draws from ``sig = 1 / sqrt(2)``.
```

graphix/random_objects.py
```diff
@@ -122,16 +137,13 @@ def rand_channel_kraus(
     if rank is None:
         rank = dim**2
 
-    if sig == "ginibre":
-        sig = 1.0 / np.sqrt(2 * dim)
-
     if not isinstance(rank, int):
         raise TypeError("The rank of a Kraus expansion must be an integer.")
```
matulni: Isn't this ensured by the static type checker? I think it's safe to remove it, since `range(float)` will raise an exception either way.

Same comments in `rand_pauli_channel_kraus` below.

graphix/random_objects.py
```diff
@@ -41,7 +45,7 @@ def rand_unit(sz: int, rng: Generator | None = None) -> npt.NDArray[np.complex12
 UNITS = np.array([1, 1j])
 
 
-def rand_dm(dim: int, rng: Generator | None = None, rank: int | None = None) -> npt.NDArray[np.complex128]:
+def rand_dm(dim: IntLike, rng: Generator | None = None, rank: IntLike | None = None) -> npt.NDArray[np.float64]:
```
matulni: Here too, why not `np.complex128` ?

graphix/random_objects.py
thierry-martinez: I added this comment in ffe24c2:
```python
    # unitary_group.rvs returns onp.Array3D[np.float64]
    # https://github.com/scipy/scipy-stubs/blob/3b629159e8da5cc3aa82b871135489d6d2fd5f8e/scipy-stubs/stats/_multivariate.pyi#L370
```

graphix/random_objects.py
thierry-martinez: Fixed in 74ca821.

graphix/random_objects.py
thierry-martinez: Indeed. Fixed in 274977d.

graphix/random_objects.py
thierry-martinez: The return type is inherited from `rand_unit`.

graphix/random_objects.py
matulni: I think this is a mistake in the stubs:
```python
import numpy as np
from graphix.random_objects import rand_unit

assert rand_unit(3).dtype == np.complex128
```

Also, `rand_unit(sz=1)` clearly returns a complex number.

thierry-martinez: Oops, yes, you're right of course! I reported the issue upstream (https://github.com/scipy/scipy-stubs/issues/987) and I temporarily fixed the annotation in 202788b by coercing the result with `.astype(np.complex128, copy=False)` (which should be a no-op, since the array at runtime is already `np.complex128`).

graphix/random_objects.py
```diff
@@ -34,20 +34,20 @@ def rand_herm(sz: IntLike, rng: Generator | None = None) -> npt.NDArray[np.compl
...
     """Generate haar random unitary matrix of size sz*sz."""
     rng = ensure_rng(rng)
     if sz == 1:
         return np.array([np.exp(1j * rng.random(size=1) * 2 * np.pi)])
-    # unitary_group.rvs returns onp.Array3D[np.float64]
-    # https://github.com/scipy/scipy-stubs/blob/3b629159e8da5cc3aa82b871135489d6d2fd5f8e/scipy-stubs/stats/_multivariate.pyi#L370
-    return unitary_group.rvs(sz, random_state=rng)
+    # unitary_group.rvs is currently annotated onp.Array3D[np.float64] in scipy-stubs
+    # See https://github.com/scipy/scipy-stubs/issues/987
+    return unitary_group.rvs(sz, random_state=rng).astype(np.complex128, copy=False)
```
matulni: Your bug report appears to be [fixed and released](https://github.com/scipy/scipy-stubs/releases/tag/v1.16.3.1) so we can remove that. Thanks!

```suggestion
    return unitary_group.rvs(sz, random_state=rng)
```

graphix/random_objects.py
```diff
@@ -84,21 +99,21 @@ def rand_gauss_cpx_mat(
     ----------
     dim : int
         Linear dimension of the (square) matrix
-    sig : float
+    sig : float | Literal["ginibre"] | None
```
emlynsg, line 102: Do you want to use `_SIG` type here too?

thierry-martinez: Since `_SIG` is private, I think it is preferable to avoid making it appear in the documentation.

PR #371, opened 2025-11-20
Drop support for Python 3.9

thierry-martinez:
This commit drops support for Python 3.9, the [end of life](https://devguide.python.org/versions/) of which was on 2025-10-31. This will enable the use of Python 3.10 features in the codebase: pattern matching, dataclass slots, etc.


matulni (approved):
Thanks !

shinich1 (approved):

PR #374, opened 2025-11-25
Refactor of flow tools - `OpenGraph.isclose`

matulni:
This commit adapts the existing method `:func: OpenGraph.isclose` to the new API introduced in #358.
Additionally, it introduces a new method `:func: OpenGraph.__eq__` relevant for open graphs of parametric type `graphix.fundamentals.Plane` or `graphix.fundamentals.Axis`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- #347:
+  - Introduced new method `graphix.opengraph.OpenGraph.is_equal_structurally` which compares the underlying structure of two open graphs.
+  - Added new method `isclose` to `graphix.fundamentals.AbstractMeasurement` which defaults to `==` comparison.
 
 ### Fixed
 
@@ -15,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.3.3] - 2025-10-23
 
+- #347: Adapted existing method `graphix.opengraph.OpenGraph.isclose` to the new API introduced in #358.
 ### Added
 
 - #343: Circuit exporter to OpenQASM3:
--- a/graphix/measurements.py
+++ b/graphix/measurements.py
@@ -90,13 +90,15 @@ class Measurement(AbstractPlanarMeasurement):
         >>> Measurement(0.1, Plane.XY).isclose(Measurement(0.0, Plane.XY))
         False
         """
-        if isinstance(other, Measurement):
-            return (
+        return (
+            isinstance(other, Measurement)
+            and (
                 math.isclose(self.angle, other.angle, rel_tol=rel_tol, abs_tol=abs_tol)
                 if isinstance(self.angle, float) and isinstance(other.angle, float)
                 else self.angle == other.angle
-            ) and self.plane == other.plane
-        return False
+            )
+            and self.plane == other.plane
+        )
 
     def to_plane_or_axis(self) -> Plane | Axis:
         """Return the measurements's plane or axis.


graphix/opengraph.py
```diff
@@ -405,5 +436,27 @@ def merge_ports(p1: Iterable[int], p2: Iterable[int]) -> list[int]:
         return OpenGraph(g, inputs, outputs, measurements), mapping_complete
 
 
+def _compare_opengraph_structure(og_1: OpenGraph[_M_co], og_2: OpenGraph[_M_co]) -> bool:
```
thierry-martinez: I would expose it as a method `OpenGraph.is_equal_structurally`.

graphix/opengraph.py
```diff
@@ -405,5 +436,27 @@ def merge_ports(p1: Iterable[int], p2: Iterable[int]) -> list[int]:
...
+    bool
+        ``True`` if both open graphs have the same underlying structure.
+    """
+    if not nx.utils.graphs_equal(og_1.graph, og_2.graph):
+        return False
+
+    if og_1.input_nodes != og_2.input_nodes or og_1.output_nodes != og_2.output_nodes:
+        return False
+
+    return set(og_1.measurements.keys()) == set(og_2.measurements.keys())
```
thierry-martinez: Isn't it redundant? If both `og_1` and `og_2` are well-formed, then these keys are their non-output nodes, and we already know that the two graphs have the same nodes and the same output nodes.

graphix/opengraph.py
```diff
@@ -405,5 +436,27 @@ def merge_ports(p1: Iterable[int], p2: Iterable[int]) -> list[int]:
...
+    Returns
+    -------
+    bool
+        ``True`` if both open graphs have the same underlying structure.
+    """
+    if not nx.utils.graphs_equal(og_1.graph, og_2.graph):
+        return False
+
+    if og_1.input_nodes != og_2.input_nodes or og_1.output_nodes != og_2.output_nodes:
+        return False
```
```suggestion
    return (
        nx.utils.graphs_equal(og_1.graph, og_2.graph)
        and og_1.input_nodes == og_2.input_nodes
        and og_1.output_nodes == og_2.output_nodes
    )
```

graphix/opengraph.py
```diff
@@ -140,6 +114,63 @@ def to_pattern(self: OpenGraph[Measurement]) -> Pattern:
 
         raise OpenGraphError("The open graph does not have flow. It does not support a deterministic pattern.")
 
+    def __eq__(self, other: object) -> bool:
```
thierry-martinez: I would avoid defining `__eq__`, since there are several possible equivalence relations. In this case, I would prefer `is_equal_up_to_angles` (acting on both `OpenGraph`s).

tests/test_opengraph.py
```diff
@@ -633,8 +633,76 @@ def test_from_to_pattern(self, fx_rng: Generator) -> None:
...
+                    measurements={0: Measurement(0.6, Plane.XY)},
+                ),
+            ),
+        ],
+    )
+    def test_eq(self, test_case: Sequence[tuple[OpenGraph[_M_co], OpenGraph[_M_co]]]) -> None:
+        og_1, og_2 = test_case
+        assert og_1 == og_1  # noqa: PLR0124
+        assert og_1 != og_2
+        assert og_2 == og_2  # noqa: PLR0124
```
thierry-martinez: There is an interesting case missing when two graphs have different angles but are equal up to angles.

graphix/opengraph.py
matulni: Yes, in a context where we don't allow incorrect open graphs, this is redundant. However, I thought that we were moving away from this model, i.e., substituting the `post_init` by a `check_well_formed` method. Anticipating this design, I thought it might be interesting to ensure that two open graphs are equal, even if they are not correct (for debugging purposes, mainly). If you think the sacrifice in readability and efficiency is not worth, I'm ok with suppressing the check on the measured nodes.

graphix/opengraph.py
matulni: I agree that `__eq__` is not a good idea because there're several possible equivalence relations. However, I'm not very comfortable with a method named `is_equal_up_to_angles` with the current implementation because then
```python
og1 = OpenGraph(
                    graph=nx.Graph([(0, 1)]),
                    input_nodes=[0],
                    output_nodes=[1],
                    measurements={0: Measurement(0.6, Plane.XY)},
                )
og2 = OpenGraph(
                    graph=nx.Graph([(0, 1)]),
                    input_nodes=[0],
                    output_nodes=[1],
                    measurements={0: Measurement(0, Plane.XY)},
                )
og3 = OpenGraph(
                    graph=nx.Graph([(0, 1)]),
                    input_nodes=[0],
                    output_nodes=[1],
                    measurements={0: Measurement(0.1, Plane.XY)},
                )

og1.is_equal_up_to_angles(og2) # -> False
og1.is_equal_up_to_angles(og3) # -> True
```

(This issue is already present in the current implementation)

Maybe we can just keep `isclose` and introduce a measurement comparer as in https://github.com/TeamGraphix/graphix/pull/375#discussion_r2560348372 ?

thierry-martinez: Indeed, if there is an `AbstractMeasurement.isclose`, using it to define `OpenGraph.isclose` seems to be the proper way to go.

matulni: Fixed in [ecbbacb](https://github.com/TeamGraphix/graphix/pull/374/commits/ecbbacbfa746929bd855e0f52ffb6aead52972a3)

graphix/opengraph.py
matulni: Done in [ecbbacb](https://github.com/TeamGraphix/graphix/pull/374/commits/ecbbacbfa746929bd855e0f52ffb6aead52972a3)

graphix/opengraph.py
thierry-martinez: Okay, if the plan is to move towards allowing incorrect open graphs to be constructed and possibly checked afterwards, it makes sense to have a comparison function that even works on incorrect graphs.

matulni: Hi again @thierry-martinez. When writing #381 and #378 I realised that it's quite practical to assume that open graphs are well formed. I propose that we keep (at least for now) the `_post_init_` design pattern for open graphs. 

I think that for open graphs this is justified because open graphs are more often input directly by the user (contrary to flow objects, which result from built-in generation routines).

In [5cd51d6](https://github.com/TeamGraphix/graphix/pull/374/commits/5cd51d61e955e21c5e0a1f7f5315b53cbff35e9b), I removed the redundancy in the structural comparison of (well-formed) open graphs as you suggested.

PR #375, opened 2025-11-25
Refactor of flow tools - `OpenGraph.compose`

matulni:
This commit adapts the existing method `:func: OpenGraph.compose` to the new API introduced in https://github.com/TeamGraphix/graphix/pull/358.

Note that tests use the comparison utilities introduced in #374. To ease the review, this is a pull request on the [TeamGraphix:rf_og_compare branch](https://github.com/TeamGraphix/graphix/tree/rf_og_compare).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,17 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-- #347:
+- #374:
   - Introduced new method `graphix.opengraph.OpenGraph.is_equal_structurally` which compares the underlying structure of two open graphs.
   - Added new method `isclose` to `graphix.fundamentals.AbstractMeasurement` which defaults to `==` comparison.
 
 ### Fixed
 
 ### Changed
+- #374: Adapted existing method `graphix.opengraph.OpenGraph.isclose` to the new API introduced in #358.
+- #375: Adapted existing method `graphix.opengraph.OpenGraph.compose` to the new API introduced in #358.
 
 ## [0.3.3] - 2025-10-23
 
-- #347: Adapted existing method `graphix.opengraph.OpenGraph.isclose` to the new API introduced in #358.
 ### Added
 
 - #343: Circuit exporter to OpenQASM3:


graphix/opengraph.py
```diff
@@ -399,13 +396,19 @@ def compose(
             raise ValueError("Keys of mapping must be correspond to nodes of other.")
         if len(mapping) != len(set(mapping.values())):
             raise ValueError("Values in mapping contain duplicates.")
+
+        def equal_measurements(vm: AbstractMeasurement, um: AbstractMeasurement) -> bool:
+            return vm.isclose(um) if isinstance(vm, Measurement) and isinstance(um, Measurement) else vm == um
```
thierry-martinez: Couldn't we have `isclose` defined in `AbstractMeasurement`, implemented by default as `==`, and overridden in `Measurement`?

matulni: I agree, it feels natural, but I couldn't figure out how to type the abstract method.

If you do something  like

```python
class AbstractMeasurement(ABC):
    @abstractmethod
    def isclose(self, other: AbstractMeasurement, rel_tol: float = 1e-09, abs_tol: float = 0.0) -> bool:
        return self == other
```

and you keep the implementation of `Measurement.isclose` as is, you have a weird situation where 
`Measurement(0.1, Plane.XY).isclose(Plane.XY)` is forbidden by the type-checker (and rightly so, because `Plane.XY` does not have an `angle` attribute), but `Plane.XY.isclose(Measurement(0.1, Plane.XY))` is legal and returns `False`.

thierry-martinez: The problem is that the method `isclose` is shadowed by its re-implementation in `Measurement`. If you define `isclose` as you propose in `AbstractMeasurement` and add the `@override` annotation on its implementation in `Measurement`, you will get the following type error with `@override`:
```python
error: Argument 1 of "isclose" is incompatible with supertype "graphix.fundamentals.AbstractMeasurement"; supertype defines the argument type as "AbstractMeasurement"  [override]
```
Indeed, the type for the parameter `other: AbstractMeasurement` cannot be narrowed by overriding (the type of parameters can only be widened by overriding). I think that the proper fix is to take `other: AbstractMeasurement` in both cases, and use `isinstance` in `Measurement` to perform the check on angle values.

With this change, `Measurement(0.1, Plane.XY).isclose(Plane.XY)` will type-check and return `False.`

matulni: Thanks for the explanation !
Implemented in [ecbbacb](https://github.com/TeamGraphix/graphix/pull/374/commits/ecbbacbfa746929bd855e0f52ffb6aead52972a3)

Commit [4e4260e](https://github.com/TeamGraphix/graphix/pull/375/commits/4e4260e24bf2714745bfcfc81fecf6ea6fc929d7)
updates `OpenGraph.compose` with the new comparison method.

tests/test_opengraph.py
```diff
@@ -544,6 +555,241 @@ def _og_19() -> OpenGraphFlowTestCase:
...
+    [0] -- 17 -- (23)
+           |
+    [3] -- 4  -- (13)
+
+    Graph 2
+    [6] -- 17 -- (1)
+     |     |
+    [7] -- 4  -- (2)
+
+    Mapping: 6 -> 23, 7 -> 13, 1 -> 100, 2 -> 200, 17 -> 90
```
emlynsg, line 615: Do you want `4 -> 201` in the description?

matulni: No, because `4 -> 201` is not specified in the input mapping to `OpenGraph.compose`, but done automatically under the hood. See L634

PR #378, opened 2025-11-27
Refactor of flow tools - Verification of flow objects

matulni:
This PR introduces the methods `:func: PauliFlow.check_well_formed`, `:func: GFlow.check_well_formed` and `:func: CausalFlow.check_well_formed` which verify the correctness of flow objects and raise exceptions when the flow is incorrect.

Exception classes are grouped in the new module `graphix.flow.exceptions`.

The error messages explain which proposition in the flow is violated. Flow-finding algorithms always (in principle) output well-formed flows, but it is possible to instantiate flow objects by passing arbitrary parameters to the constructors. In such cases, there is not any guarantee that the flow objects are well formed. The methods introduced here can be useful for debugging or researching. The exception handling adapts the pattern introduced in #364.

The new methods subsume `:func: gflow.verify_flow`, `:func: gflow.verify_gflow`, `:func: gflow.verify_pauli_flow`.

Additionally, this PR introduces the methods `:func: PauliFlow.get_measurement_label`, `:func: GFlow.get_measurement_label` which return the measurement label of a given node. They follow same criteria employed in the flow-finding algorithms, that is, querying this method on a node with a measurement `Measurement(0, Plane.XY)` will return `Plane.XY` in a `GFlow` or `CausalFlow` and `Axis.X` in a `PauliFlow`.

**Additional information on the exception management**

```mermaid
---
config:
  layout: elk
  elk:
    mergeEdges: false
    nodePlacementStrategy: LINEAR_SEGMENTS
---
flowchart TD

a0(**Exception**)

n0(**FlowError**)

n1(FlowGenericError)
n2(FlowPropositionError)
n3(FlowPropositionOrderError)
n4(PartialOrderError)
n5(PartialOrderLayerError)

a0 --> n0

n0 --FGEReason--> n1
n0 --FPEReason--> n2
n0 --FPOEReason--> n3
n0 --POEReason--> n4
n0 --POLEReason--> n5
```
- Arrows indicate inheritance (from parent to child).

- "Reasons" (`FlowPropositionErrorReason`, `PartialOrderLayerErrorReason`, etc.) are `Enum` classes.

- Error subclasses:
    - `FlowPropositionError`
        - Violations of the flow-definition propositions which concern the correction function only (C0, C1, G1, G3, G4, G5, P4, P5, P6, P7, P8, P9).
        - Additional parameters:
            - `node`
            - `correction_set`
    - `FlowPropositionOrderError`
        - Violations of the flow-definition propositions which concern the correction function and the partial order (C2, C3, G1, G2, P1, P2, P3).
        - Additional parameters:
            - `node`
            - `correction_set`
            - `past_and_present_nodes`
    - `FlowGenericError`
        - General errors in the flow correction function, XY planes in causal flow.
        - Does not require additional parameters.
    - `PartialOrderError`
        - General flow and XZ-corrections errors in the partial order.
        - Does not require additional parameters.
    - `PartialOrderLayerError`
        - Flow and XZ-corrections errors concerning a specific layer of the partial order.
        - Additional parameters:
            - `layer_index`
            - `layer`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- #378:
+  - Introduced new method `graphix.flow.core.PauliFlow.check_well_formed`, `graphix.flow.core.GFlow.check_well_formed` and `graphix.flow.core.CausalFlow.check_well_formed` which verify the correctness of flow objects and raise exceptions when the flow is incorrect.
+  - Introduced new method `graphix.flow.core.PauliFlow.is_well_formed` which verify the correctness of flow objects and returns a boolean when the flow is incorrect.
+  - Introduced new module `graphix.flow.exceptions` grouping flow exceptions.
+  - Introduced new methods `graphix.flow.core.PauliFlow.get_measurement_label` and `graphix.flow.core.GFlow.get_measurement_label` which return the measurement label of a given node following same criteria employed in the flow-finding algorithms.
 
 - #374:
   - Introduced new method `graphix.opengraph.OpenGraph.is_equal_structurally` which compares the underlying structure of two open graphs.


matulni:
Note: the functionality introduced here assumes that the open graph is well formed (currently ensured at instantiation). This should be updated when/if we move away from this design. See https://github.com/TeamGraphix/graphix/pull/374#discussion_r2565403949

PR #379, opened 2025-11-28
Addition of `CZ` instruction and related components, and removal of `meas_index` from `R` instructions

emlynsg:
**Context:**
Up to now, Graphix has not explicitly included a `CZ` instruction. This may be something users pass as a gate in a circuit. It is also a useful addition for using in alternative transpilers including JCZ.

A `CZ` gate has been added to the set of instructions in `instruction.py`.
Downstream changes have been made including:
- Added to `InstructionKind` enumeration
- Added to multiple points in `transpiler.py` for circuit transpilation.

This PR also updates other instructions. It removes an unused variable `meas_index` in the `RZZ`, `RZ`, `RY`, `RX` instructions. Other minor typing fixes have also been made to some test suites where identified during testing.

**Related issue:**
See parallel PR in [graphix-qasm-parser](https://github.com/TeamGraphix/graphix-qasm-parser/pull/5).

Given parallel PR was required, some workarounds have been included to prevent dependency problems. These can be reverted when CI is shown to be passing in both repos.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+
 - #385
   - Introduced `graphix.flow.core.XZCorrections.check_well_formed` which verifies the correctness of an XZ-corrections instance and raises an exception if incorrect.
   - Added XZ-correction exceptions to module `graphix.flow.core.exceptions`.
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = { file = ["requirements.txt"] }
 dev = { file = ["requirements-dev.txt"] }
 extra = { file = ["requirements-extra.txt"] }
 doc = { file = ["requirements-doc.txt"] }
+no-deps = { file = ["requirements-no-deps.txt"] }
```
thierry-martinez: Why do we need to split between `dev` and `no-deps`? I think, we can pin `graphix-qasm-parser` in `requirements-dev.txt` while we need to install a specific branch.

tests/test_transpiler.py
```diff
@@ -16,6 +16,14 @@
 
 
 class TestTranspilerUnitGates:
+    def test_cz(self, fx_rng: Generator) -> None:
+        circuit = Circuit(2)
+        circuit.cz(0, 1)
+        pattern = circuit.transpile().pattern
+        state = circuit.simulate_statevector().statevec
```
matulni:
```suggestion
        state = circuit.simulate_statevector(rng=fx_rng).statevec
```

tests/test_tnsim.py
```diff
@@ -288,6 +288,18 @@ def test_i(self, fx_rng: Generator) -> None:
         value2 = tn_mbqc.expectation_value(random_op1, [0])
         assert value1 == pytest.approx(value2)
 
+    def test_cz(self, fx_rng: Generator) -> None:
+        circuit = Circuit(2)
+        circuit.cz(0, 1)
+        pattern = circuit.transpile().pattern
+        pattern.standardize()
+        state = circuit.simulate_statevector().statevec
```
matulni, line 296:
```suggestion
        state = circuit.simulate_statevector(rng=fx_rng).statevec
```

thierry-martinez: The random number generator is only useful when there are measurements in the circuit.

emlynsg:
Agree @thierry-martinez I think this is ready to go now.

PR #383, opened 2025-12-01
Introduce `BaseN` and `PrepareMethod`

thierry-martinez:
This commit introduces the classes `BaseN` and `PrepareMethod`, which allow the user to customize how `N` commands are handled. For instance, in the VBQC context of Veriphix, `N` commands do not necessarily prepare the node in the |+> state: for each prepared qubit, the preparation should be performed by the client, so that the server does not even see the state.


graphix/pattern.py
```diff
@@ -1544,11 +1544,18 @@ def check_measured(cmd: Command, node: int) -> None:
                 check_active(cmd, n1)
             elif cmd.kind == CommandKind.M:
                 check_active(cmd, cmd.node)
-                for domain in cmd.s_domain, cmd.t_domain:
-                    if cmd.node in domain:
-                        raise RunnabilityError(cmd, cmd.node, RunnabilityErrorReason.DomainSelfLoop)
-                    for node in domain:
-                        check_measured(cmd, node)
+                if isinstance(cmd, command.M):
```
emlynsg, line 1556: This leaves out part of the runnability checking. Do we think it's worth adding a warning here? Or just add a clarifying line in the docstring that makes it clear?

thierry-martinez: I added some notes in the docstring in e707caf.

graphix/command.py
```diff
@@ -57,7 +81,6 @@ class N(_KindChecker, DataclassReprMixin):
         Initial state, defaults to :class:`~graphix.states.BasicStates.PLUS`.
     """
 
-    node: Node
     state: State = dataclasses.field(default_factory=lambda: BasicStates.PLUS)
     kind: ClassVar[Literal[CommandKind.N]] = dataclasses.field(default=CommandKind.N, init=False)
```
matulni, line 85: Genuine question: why do we redefine this attribute if it already exists in the parent dataclass `BaseN`?

thierry-martinez: Because the parent class `_KindChecker` makes the class declaration fail with the error `TypeError: kind must be annotated` if the `kind` field is missing.

PR #385, opened 2025-12-04
Refactor of flow tools - Verification of `XZCorrections` objects

matulni:
This PR adds a new method `:func: XZCorrections.check_well_formed` to verify the correctness of XZ-corrections objects.

A correct `XZCorrections` instance verifies the following properties:
    - Keys of the correction dictionaries are measured nodes, i.e., a subset of $O^c$.
    - Corrections respect the partial order.
    - The first layer of the partial order contains all the output nodes if there are any.
    - The partial order contains all the nodes (without duplicates) and it does not have empty layers.

This PR builds on the exception classes introduced in #378. We detail next the full exception landscape.

```mermaid
---
config:
  layout: elk
  elk:
    mergeEdges: false
    nodePlacementStrategy: LINEAR_SEGMENTS
---
flowchart TD

a0(**Exception**)

m0(**XZCorrectionsError**)

n0(**FlowError**)

n1(FlowGenericError)
n2(FlowPropositionError)
n3(FlowPropositionOrderError)
n4(PartialOrderError)
n5(PartialOrderLayerError)

m1(XZCorrectionsOrderError)
m2(XZCorrectionsGenericError)

a0 --> m0
a0 --> n0

m0 --POEReason--> n4
m0 --POLEReason--> n5
m0 --XZCOEReason--> m1
m0 --XZCGEReason--> m2

n0 --FGEReason--> n1
n0 --FPEReason--> n2
n0 --FPOEReason--> n3
n0 --POEReason--> n4
n0 --POLEReason--> n5
```
- Arrows indicate inheritance (from parent to child).

- "Reasons" (`XZCorrectionsOrderErrorReason`, etc.) are `Enum` classes.

- New error subclasses:
    - `XZCorrectionsOrderError`
        - Errors in XZ-corrections objects concerning the correction dictionaries and the partial order.
        - Additional parameters:
            - `node`
            - `correction_set`
            - `past_and_present_nodes`
    - `XZCorrectionsGenericError`
        - Generic errors.
        - Does not require additional parameters.

### DIFF
--- a/graphix/graphsim.py
+++ b/graphix/graphsim.py
@@ -462,7 +462,7 @@ class GraphState(Graph):
         self.remove_node(node)
         return result
 
-    def draw(self, fill_color: str = "C0", **kwargs: dict[str, Any]) -> None:
+    def draw(self, fill_color: str = "C0") -> None:
         """Draw decorated graph state.
 
         Negative nodes are indicated by negative sign of node labels.
@@ -471,8 +471,6 @@ class GraphState(Graph):
         ----------
         fill_color : str
             optional, fill color of nodes
-        kwargs :
-            optional, additional arguments to supply networkx.draw().
         """
         nqubit = len(self.nodes)
         nodes = list(self.nodes)
@@ -489,7 +487,7 @@ class GraphState(Graph):
         g: nx.Graph[int] = nx.Graph()
         g.add_nodes_from(nodes)
         g.add_edges_from(edges)
-        nx.draw(g, labels=labels, node_color=colors, edgecolors="k", **kwargs)
+        nx.draw(g, labels=labels, node_color=colors, edgecolors="k")
 
     def to_statevector(self) -> Statevec:
         """Convert the graph state into a state vector."""
--- a/graphix/visualization.py
+++ b/graphix/visualization.py
@@ -699,15 +699,10 @@ class GraphVisualizer:
 
         pos = nx.multipartite_layout(g_prime)
 
-        for node, layer in l_k.items():
-            pos[node][0] = l_max - layer
-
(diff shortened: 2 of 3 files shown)


MERGE_MSG
emlynsg, line 1: Can this be removed?

graphix/flow/core.py
```diff
@@ -273,6 +284,73 @@ def is_compatible(self, total_measurement_order: TotalOrder) -> bool:
...
+        # Unlike for flows, XZCorrections can be well defined on open graphs without outputs
+        if o_set:
+            if first_layer != o_set:
+                raise PartialOrderLayerError(PartialOrderLayerErrorReason.FirstLayer, layer_index=0, layer=first_layer)
+            shift = 1
+        else:
+            shift = 0
+
+        measured_layers = reversed(self.partial_order_layers[shift:])
+        layer_idx = len(self.partial_order_layers) - 1
```
emlynsg: Should this be
```python
layer_idx = len(self.partial_order_layers) - shift
``` 
?

MERGE_MSG
matulni: Yes, thanks! Done in c13057d5d57ed34298e8f7c08327aeba6b763f55

graphix/flow/core.py
matulni: mmm, I don't think so... `shift` removes the first element of `self.partial_order_layers` if it contains output nodes. We initialize `layer_idx` to the last layer index (that is, the length of the list minus 1, whether we removed the first element or not) to keep track of the layer index when iterating in reverse order. I added a comment in 095a2ded6aeffec623eb085f71a1eef097d2666c.

Issue #363, opened 2025-11-13
Strengthened definition of gflow means Pauli presimulation does not preserve gflow.


emlynsg:
With reference to PR #337 and PR #347, and issue #168.

**Describe the bug**
Before and after PR #337, [example two in the docs](https://graphix.readthedocs.io/en/latest/gallery/visualization.html) has changed.

<img width="320" height="300" alt="Image" src="https://github.com/user-attachments/assets/e23d5fbf-a895-43e2-8f97-17dd64888136" />

`"No flow or gflow detected in the graph."`

**Additional context**
As explained by Thierry and Mateo, the definition of gflow was strengthened in PR #337 to match the definition from the paper (input nodes cannot be corrected by a gflow). This strengthened definition seems to be incompatible with Pauli presimulation preserving the gflow, leading to examples where a gflow was detected before.

However, Pauli presimulation preserving the gflow is a desirable and expected property, so we should consider how to deal with this change.

Need to change Pauli presimulation to force plus states.

mgarnier59:
Not sure I understand what you mean. 

> input nodes cannot be corrected by a gflow

Do you mean that input nodes cannot can never be part of any correction set and thus have to be measured in the XY plane?

> However, Pauli presimulation preserving the gflow is a desirable and expected property

Is it expected? I think it should preserve Pauli flow (cf [https://arxiv.org/abs/2207.09368](https://arxiv.org/abs/2207.09368)) but I'm not sure about gflow

emlynsg:
Main phase resolved by PR #392 and remaining phase 2 to be considered in future.

PR #392, opened 2025-12-09
Removed inputs from Pauli presimulation

emlynsg:
PR correcting issues identified in discussion of issue #363 . Opening PR with a couple of remaining todos to allow for discussion about the merits of removing the `keep_inputs` field across the Graphix codebase.

This PR removes the `keep_inputs` field and related code from Pauli presimulation components of patterns and tests.
This also results in the removal of any inputs passed through the Pauli presimulator, and new inputs being generated for output patterns.

*Todo:*
- Discuss change to FFT test - number of nodes has changed compared to previous code, but the rest of the tests still pass. Some artefact of the change from input to created nodes.

### DIFF
--- a/tests/test_pattern.py
+++ b/tests/test_pattern.py
@@ -247,8 +247,8 @@ class TestPattern:
         pattern.remove_input_nodes()
         pattern.perform_pauli_measurements()
         isolated_nodes = pattern.extract_isolated_nodes()
-        # 30-node is the isolated and output node.
-        isolated_nodes_ref = {30}
+        # 42-node is the isolated and output node.
+        isolated_nodes_ref = {42}
         assert isolated_nodes == isolated_nodes_ref
 
     def test_pauli_measurement_error(self, fx_rng: Generator) -> None:
@@ -813,6 +813,7 @@ class TestPattern:
         c = Circuit(1)
         c.rz(0, 0.2)
         p = c.transpile().pattern
+        p.remove_input_nodes()
         p.perform_pauli_measurements()
         assert p.extract_partial_order_layers() == (frozenset({2}), frozenset({0}))
 
@@ -938,6 +939,8 @@ class TestPattern:
         p_ref = circuit_1.transpile().pattern
         p_test = p_ref.extract_causal_flow().to_corrections().to_pattern()
 
+        p_ref.remove_input_nodes()
+        p_test.remove_input_nodes()
         p_ref.perform_pauli_measurements()
         p_test.perform_pauli_measurements()
 
@@ -954,7 +957,8 @@ class TestPattern:
         circuit_1 = rand_circuit(nqubits, depth, rng, use_ccx=False)
         p_ref = circuit_1.transpile().pattern
         p_test = p_ref.extract_gflow().to_corrections().to_pattern()
-
+        p_ref.remove_input_nodes()
+        p_test.remove_input_nodes()
         p_ref.perform_pauli_measurements()
         p_test.perform_pauli_measurements()
 


graphix/pattern.py
```diff
@@ -1659,8 +1651,8 @@ def measure_pauli(
     graph = standardized_pattern.extract_graph()
     graph_state = GraphState(nodes=graph.nodes, edges=graph.edges, vops=standardized_pattern.c_dict)
     results: dict[int, Outcome] = {}
-    to_measure, non_pauli_meas = pauli_nodes(standardized_pattern, leave_input)
-    if not leave_input and len(list(set(pattern.input_nodes) & {i[0].node for i in to_measure})) > 0:
+    to_measure, non_pauli_meas = pauli_nodes(standardized_pattern)
+    if len(list(set(pattern.input_nodes) & {i[0].node for i in to_measure})) > 0:
```
```suggestion
    if len(set(pattern.input_nodes) & {i[0].node for i in to_measure}) > 0:
```

tests/test_pattern.py
```diff
@@ -216,21 +216,6 @@ def test_pauli_measurement_single(self, plane: Plane, angle: float) -> None:
...
-        circuit = rand_circuit(nqubits, depth, rng)
-        pattern = circuit.transpile().pattern
-        pattern.standardize()
-        pattern.shift_signals(method="mc")
-        pattern.perform_pauli_measurements(leave_input=True)
-        pattern.minimize_space()
-        state = circuit.simulate_statevector().statevec
-        state_mbqc = pattern.simulate_pattern(rng=rng)
-        assert np.abs(np.dot(state_mbqc.flatten().conjugate(), state.flatten())) == pytest.approx(1)
-
```
matulni, line 238: Is there a reason to remove this test? In principle, it should pass if we remove `leave_input` from L228, right?
Or does it become a redundant test then?

tests/test_pattern.py
```diff
@@ -259,34 +244,6 @@ def test_pauli_measurement(self) -> None:
...
-        pattern.standardize()
-        pattern.shift_signals(method="mc")
-        pattern.perform_pauli_measurements(leave_input=True)
-
-        isolated_nodes = pattern.extract_isolated_nodes()
-        # There is no isolated node.
-        isolated_nodes_ref: set[int] = set()
-
-        assert isolated_nodes == isolated_nodes_ref
-
```
matulni, line 294: I have no idea about what to expect for the output of this function, so I'm ok with removing the test. I believe that functions `cp` and `swap` can be safely removed from the test module as well.

matulni (commented):
Thanks @emlynsg for this PR, it addresses a crucial issue!

I made some small code remarks and some comment about the API below. We can discuss it.

We can think of a strongly deterministic pattern $P$ with $m$ inputs and $n$ ouputs as unitary mapping between two Hilbert spaces with dimensions $m$ and $n$,

$$P: \mathcal{H}^{\otimes m} \rightarrow \mathcal{H}^{\otimes n},$$

with $m \geq n$.

With the current API, `Pattern.perform_pauli_measurements` can be understood as a transformation $T$, such that $T[P] = P'$, where 

$$P': \widetilde{0} \rightarrow \mathcal{H}^{\otimes n},$$

where $\widetilde{0}$ denotes a zero-dimensional Hilbert space.

This implies that $P$ and $P'$ represent the same unitary if and only if $m=0$ in the first place.

I understand that this is a limitation of the Pauli presimulator, and it's a longterm goal to improve it, but my concern is that this dimensionality reduction occurs inadvertently. In particular, the example in https://github.com/TeamGraphix/graphix/issues/168#issuecomment-3442161093 still fails. At the same time, I think it's natural to _expect_ that the Pauli presimulation preserves the unitary the pattern represents.


I propose two alternative solutions:

- Defensive approach: `Pattern.perform_pauli_measurements` raises an exception if called on a pattern with a finite number of inputs.

- Flexible approach: `Pattern.perform_pauli_measurements` raises an warning when the number of inputs is reduced.

In either case, this behavior should be clearly explained in the docstring.

tests/test_pattern.py
emlynsg: Yes duplicate test at that point.

tests/test_pattern.py
emlynsg: Yes agree. I'll check the functions.

graphix/pattern.py
thierry-martinez: `len` is useless here: the truth value of a set is `True` iff it is not empty. But I don't understand why there is a test here: `new_inputs` should now always be `[]`, shouldn't it?

emlynsg: Actually yes I think you're right Thierry. I'll have another quick look and push an update.

shinich1:
could you also update the tutorial before merging? thanks.

graphix/pattern.py
```diff
@@ -1377,14 +1382,18 @@ def simulate_pattern(
         sim.run(input_state, rng=rng)
         return sim.backend.state
 
-    def perform_pauli_measurements(self, leave_input: bool = False, ignore_pauli_with_deps: bool = False) -> None:
+    def remove_input_nodes(self) -> None:
+        """Remove the input nodes from the pattern."""
+        for node in reversed(self.input_nodes):
+            self.__seq.insert(0, command.N(node=node))
```
thierry-martinez: Nitpicking: every time you insert an element in front of a list, all the former elements are shifted to the right, so the loop has a quadratic time complexity. We can use slice assignment to insert several elements at once:

```suggestion
        self.__seq[0:0] = [command.N(node=node) for node in self.input_nodes]
```

graphix/pattern.py
```diff
@@ -1661,18 +1665,19 @@ def measure_pauli(
 
     .. seealso:: :class:`graphix.graphsim.GraphState`
     """
+    pat = Pattern() if copy else pattern
+    if pat._pauli_preprocessed is True:
+        return pat
```
thierry-martinez: This test is fragile:
- If `copy` is `True`, `pat._pauli_preprocessed` will always be `False`, even if `pattern` has been Pauli-processed.
- Even if `copy` is `False`, if we Pauli-preprocess a pattern and then add an element to it, `_pauli_preprocessed` will be reset, but `results` still contains measurements that will be discarded.

I think properly handling the former `results` contents is the way to go.

thierry-martinez (approved):
LGTM!

Issue #367, opened 2025-11-17
[Bug]: `Pattern.get_layers` assumes that inputs cannot be outputs


matulni:
`Pattern.get_layers` assumes that inputs cannot be outputs.

This results in inconsistent behaviour, e.g.,

```python
from graphix.pattern import Pattern

p1 = Pattern(input_nodes=[0, 1], cmds=[E((0, 1))])
print(p1) # E(0,1)

meas_order_1 = p1._measurement_order_depth()
print(meas_order_1) # [0, 1] -> Incorrect, [0, 1] are not measured.

p2 = Pattern(input_nodes=[], cmds=[N(0), N(1), E((0, 1))])
print(p2) # E(0,1) N(1) N(0)

meas_order_2 = p2._measurement_order_depth()
print(meas_order_2) # [] -> Correct

assert set(meas_order_1) == set(meas_order_2) # Error
```

To be fixed in Flow refactor part 2

matulni:
Fixed in #407: Method `graphix.Pattern.get_layers` is subsumed by `graphix.Pattern.extract_partial_order_layers`.

PR #393, opened 2025-12-09
Refactor of flow tools - Flow from pattern

matulni:
This PR introduces some tooling to extract a causal flow and a generalised flow from a pattern without running the extraction algorithms on the underlying open graph.

It directly addresses #306 and additionally solves #367.

### Description of changes

- We introduce the following methods:

    - `graphix.optimization.StandardizedPattern.extract_partial_order_layers`
        - Constructs a partial order layering from the dependency domains of $M$, $X$ and $Z$ commands. The layering follows the convention of the ``partial_order_layers`` attribute of `:class: PauliFlow` and `:class: XZCorrections` objects :
            - Nodes in the same layer can be measured simultaneously.
            - Nodes in layer `i` must be measured after nodes in layer `i + 1`.
            - All output nodes (if any) are in the first layer.
            - There cannot be any empty layers.

    - `graphix.optimization.StandardizedPattern.extract_causal_flow`
        - Attempts to extract a causal flow from a standardized pattern. The algorithm goes as follows:
            - The correction function is constructed by reading out $X$ and $M$ commands. In particular, a command $X_j^i$ implies $c(i) = j$ in the correction function. Analogously, a command $M_j^{s=i}$ also contributes $c(i) = j$ to the correction function.
            - A partial order of the pattern is extracted with `graphix.optimization.StandardizedPattern.extract_partial_order_layers`.
            - A `CausalFlow` is constructed combining these you objects.
            - The wellformedness of the `CausalFlow` object is checked.

    - `graphix.optimization.StandardizedPattern.extract_gflow`
        - Attempts to extract a gflow from a standardized pattern. The algorithm is similar to that of the causal flow, but in addition we allow  $c(i) = i$ if the node $i$ is measured in planes $XZ$ or $YZ$.

    - `graphix.optimization.StandardizedPattern.extract_opengraph`
       - Extracts an open graph from a standardized pattern. 

    - `graphix.pattern.Pattern.extract_partial_order_layers`, `graphix.pattern.Pattern.extract_causal_flow` and `graphix.pattern.Pattern.extract_gflow`
        - Wrappers which standardize the pattern and call the corresponding extraction routine.

- We introduce the a new module `graphix.flow._partial_order.py` with the function `:func: compute_topological_generations`.

    - This function was used in the algebraic Pauli-flow finding routine and is now used to extract the partial order from the pattern.

    - In the future, we should drop the dependency of `graphix.flow.core` on `nx.DiGraph` objects and replace `:func: graphix.flow.core._dag_to_partial_order_layers` with this function.

- Dependencies of the module `graphix.visualization.py` on `graphix.gflow.py` are removed where possible, but the new flow API is not fully exploited. This is out of the scope of this PR.

### DIFF
--- a/graphix/optimization.py
+++ b/graphix/optimization.py
@@ -522,7 +522,7 @@ class StandardizedPattern(_StandardizedPattern):
 
         Notes
         -----
-        The notes provided in :meth:`extract_causal_flow` apply here as well.
+        The notes provided in :func:`self.extract_causal_flow` apply here as well.
         """
         correction_function: dict[int, set[int]] = {}
         pre_measured_nodes = self.results.keys()  # Not included in the flow.
--- a/graphix/pattern.py
+++ b/graphix/pattern.py
@@ -956,8 +956,7 @@ class Pattern:
 
         Notes
         -----
-        - See :func:`optimization.StandardizedPattern.extract_gflow` for additional information on why it is required to standardized the pattern to extract a gflow.
-        - Applying the chain ``Pattern.extract_gflow().to_corrections().to_pattern()`` to a strongly deterministic pattern returns a new pattern implementing the same unitary transformation. This equivalence holds as long as the original pattern contains no Clifford commands, since those are discarded during open-graph extraction.
+        The notes provided in :func:`self.extract_causal_flow` apply here as well.
         """
         return optimization.StandardizedPattern.from_pattern(self).extract_gflow()
 


graphix/optimization.py
```diff
@@ -362,6 +371,186 @@ def ensure_neighborhood(node: Node) -> None:
...
+        -----
+        The returned object follows the same conventions as the ``partial_order_layers`` attribute of :class:`PauliFlow` and :class:`XZCorrections` objects:
+            - Nodes in the same layer can be measured simultaneously.
+            - Nodes in layer ``i`` must be measured after nodes in layer ``i + 1``.
+            - All output nodes (if any) are in the first layer.
+            - There cannot be any empty layers.
+        """
+        oset = frozenset(self.output_nodes)  # First layer by convention.
+        pre_measured_nodes = set(self.results.keys())  # Not included in the partial order layers.
+        excluded_nodes = oset | pre_measured_nodes
```
thierry-martinez: `dict_keys` is a subclass of `AbstractSet` and can be used directly as the right-hand side of `frozenset` `__or__` operator.

```suggestion
        pre_measured_nodes = self.results.keys()  # Not included in the partial order layers.
        excluded_nodes = oset | pre_measured_nodes
```

graphix/optimization.py
```diff
@@ -362,6 +371,186 @@ def ensure_neighborhood(node: Node) -> None:
...
+        zero_indegree = set(self.input_nodes) - excluded_nodes
+        dag: dict[int, set[int]] = {
+            node: set() for node in zero_indegree
+        }  # `i: {j}` represents `i -> j` which means that node `i` must be measured before node `j`.
+        indegree_map: dict[int, int] = {}
+
+        for n in self.n_list:
+            if n.node not in oset:  # Pre-measured nodes only appear in domains.
+                dag[n.node] = set()
+                zero_indegree.add(n.node)
```
thierry-martinez: The initialization of `N` nodes is the same as for input nodes and this can be reflected more explicitly.

```suggestion
        zero_indegree = set(self.input_nodes).union(n.node for n in self.n_list) - excluded_nodes
        dag: dict[int, set[int]] = {
            node: set() for node in zero_indegree
        }  # `i: {j}` represents `i -> j` which means that node `i` must be measured before node `j`.
        indegree_map: dict[int, int] = {}
```

graphix/optimization.py
```diff
@@ -362,6 +371,186 @@ def ensure_neighborhood(node: Node) -> None:
...
+
+        Notes
+        -----
+        This method makes use of :func:`StandardizedPattern.extract_partial_order_layers` which computes the pattern's direct acyclical graph (DAG) induced by the corrections and returns a particular layer stratification (obtained by doing a topological sort on the DAG). Further, it constructs the pattern's induced correction function from :math:`M` and :math:`X` commands.
+        In general, there may exist various layerings which represent the corrections of the pattern. To ensure that a given layering is compatible with the pattern's induced correction function, the partial order must be extracted from a standardized pattern. Commutation of entanglement commands with X and Z corrections in the standardization procedure may generate new corrections, which guarantees that all the topological information of the underlying graph is encoded in the extracted partial order.
+        """
+        correction_function: dict[int, set[int]] = defaultdict(set)
+        pre_measured_nodes = set(self.results.keys())  # Not included in the flow.
```
thierry-martinez:
```suggestion
        pre_measured_nodes = self.results.keys()  # Not included in the flow.
```

graphix/optimization.py
```diff
@@ -362,6 +371,186 @@ def ensure_neighborhood(node: Node) -> None:
...
+            the well-formedness conditions required for a valid gflow.
+        ValueError
+            If `N` commands in the pattern do not represent a |+⟩ state or if the pattern corrections form closed loops.
+
+        Notes
+        -----
+        This method makes use of :func:`StandardizedPattern.extract_partial_order_layers` which computes the pattern's direct acyclical graph (DAG) induced by the corrections and returns a particular layer stratification (obtained by doing a topological sort on the DAG). Further, it constructs the pattern's induced correction function from :math:`M` and :math:`X` commands.
+        In general, there may exist various layerings which represent the corrections of the pattern. To ensure that a given layering is compatible with the pattern's induced correction function, the partial order must be extracted from a standardized pattern. Commutation of entanglement commands with X and Z corrections in the standardization procedure may generate new corrections, which guarantees that all the topological information of the underlying graph is encoded in the extracted partial order.
```
thierry-martinez: I would refer to the notes of `extract_causal_flow` instead of duplicating them.


```suggestion
        The notes provided in :meth:`extract_causal_flow` apply here as well.
```

graphix/optimization.py
```diff
@@ -362,6 +371,186 @@ def ensure_neighborhood(node: Node) -> None:
...
+        """
+        correction_function: dict[int, set[int]] = defaultdict(set)
+        pre_measured_nodes = set(self.results.keys())  # Not included in the flow.
```
thierry-martinez:
```suggestion
        pre_measured_nodes = self.results.keys()  # Not included in the flow.
```

graphix/optimization.py
```diff
@@ -418,6 +607,36 @@ def _incorporate_pauli_results_in_domain(
...
+        The updated correction dictionary with `node` added to the correction
+        sets of all nodes in `domain`.
+
+    Notes
+    -----
+    This function is used to extract the correction function from :math:`X`, :math:`Z` and :math:`M` commands when constructing a flow.
+    """
+    for measured_node in domain:
+        correction[measured_node].add(node)
+    return correction
```
thierry-martinez: I would recommend returning `None` to make explicit that calls to `_update_corrections` operate by side effects.

graphix/optimization.py
```diff
@@ -362,6 +371,186 @@ def ensure_neighborhood(node: Node) -> None:
...
+        for node, domain in self.x_dict.items():
+            correction_function = _update_corrections(node, domain - pre_measured_nodes, correction_function)
+
+        og = (
+            self.extract_opengraph()
+        )  # Raises a `ValueError` if `N` commands in the pattern do not represent a |+⟩ state.
+        partial_order_layers = (
+            self.extract_partial_order_layers()
+        )  # Raises a `ValueError` if the pattern corrections form closed loops.
+        gf = GFlow(og, dict(correction_function), partial_order_layers)
```
thierry-martinez: `correction_function` is already a `dict`.
```suggestion
        gf = GFlow(og, correction_function, partial_order_layers)
```

graphix/optimization.py
```diff
@@ -362,6 +371,186 @@ def ensure_neighborhood(node: Node) -> None:
...
+        ValueError
+            If `N` commands in the pattern do not represent a |+⟩ state or if the pattern corrections form closed loops.
+
+        Notes
+        -----
+        This method makes use of :func:`StandardizedPattern.extract_partial_order_layers` which computes the pattern's direct acyclical graph (DAG) induced by the corrections and returns a particular layer stratification (obtained by doing a topological sort on the DAG). Further, it constructs the pattern's induced correction function from :math:`M` and :math:`X` commands.
+        In general, there may exist various layerings which represent the corrections of the pattern. To ensure that a given layering is compatible with the pattern's induced correction function, the partial order must be extracted from a standardized pattern. Commutation of entanglement commands with X and Z corrections in the standardization procedure may generate new corrections, which guarantees that all the topological information of the underlying graph is encoded in the extracted partial order.
+        """
+        correction_function: dict[int, set[int]] = defaultdict(set)
```
thierry-martinez: I would use a simple dictionary here:
- The typing does not reflect the fact that `_update_corrections` relies on a `defaultdict`.
- I suppose this `defaultdict` motivates casting to `dict` when constructing the `GFlow`. I think it is simpler to have a mere `dict` in the first place.

```suggestion
        correction_function: dict[int, set[int]] = {}
```

graphix/optimization.py
```diff
@@ -418,6 +607,36 @@ def _incorporate_pauli_results_in_domain(
...
+    dict[Node, set[Node]]
+        The updated correction dictionary with `node` added to the correction
+        sets of all nodes in `domain`.
+
+    Notes
+    -----
+    This function is used to extract the correction function from :math:`X`, :math:`Z` and :math:`M` commands when constructing a flow.
+    """
+    for measured_node in domain:
+        correction[measured_node].add(node)
```
thierry-martinez: If `correction` is just a `dict`:

```suggestion
        correction.setdefault(measured_node, set()).add(node)
```

graphix/optimization.py
```diff
@@ -362,6 +371,186 @@ def ensure_neighborhood(node: Node) -> None:
...
+        This method makes use of :func:`StandardizedPattern.extract_partial_order_layers` which computes the pattern's direct acyclical graph (DAG) induced by the corrections and returns a particular layer stratification (obtained by doing a topological sort on the DAG). Further, it constructs the pattern's induced correction function from :math:`M` and :math:`X` commands.
+        In general, there may exist various layerings which represent the corrections of the pattern. To ensure that a given layering is compatible with the pattern's induced correction function, the partial order must be extracted from a standardized pattern. Commutation of entanglement commands with X and Z corrections in the standardization procedure may generate new corrections, which guarantees that all the topological information of the underlying graph is encoded in the extracted partial order.
+        """
+        correction_function: dict[int, set[int]] = defaultdict(set)
+        pre_measured_nodes = set(self.results.keys())  # Not included in the flow.
+
+        for m in self.m_list:
+            if m.plane in {Plane.XZ, Plane.YZ}:
+                correction_function[m.node].add(m.node)
```
thierry-martinez: If `correction_function` is a mere `dict`:
```suggestion
                correction_function.setdefault(m.node, set()).add(m.node)
```

graphix/pattern.py
```diff
@@ -885,6 +886,85 @@ def update_dependency(measured: AbstractSet[int], dependency: dict[int, set[int]
...
+        RunnabilityError
+            If the pattern is not runnable.
+
+        Notes
+        -----
+        The returned object follows the same conventions as the ``partial_order_layers`` attribute of :class:`PauliFlow` and :class:`XZCorrections` objects:
+            - Nodes in the same layer can be measured simultaneously.
+            - Nodes in layer ``i`` must be measured after nodes in layer ``i + 1``.
+            - All output nodes (if any) are in the first layer.
+            - There cannot be any empty layers.
```
thierry-martinez:
```suggestion
        - This function wraps :func:`optimization.StandardizedPattern.extract_partial_order_layers`, and the returned object is described in the notes of this method.
```

graphix/pattern.py
```diff
@@ -1287,7 +1316,8 @@ def minimize_space(self) -> None:
             self.standardize()
         meas_order = None
         if not self._pauli_preprocessed:
-            meas_order = self.get_measurement_order_from_flow()
+            cf = self.extract_opengraph().find_causal_flow()
```
thierry-martinez, line 1314: Note that this change preserves bug #157, but I think fixing it would be outside the scope of this PR. Ideally, we would like `cf = self.try_extract_causal_flow()`, but this may cause some regressions: `minimize_space` will be correct in the sense that it will preserve runnability, but in some examples the minimized pattern may use more space than the current heuristics.

graphix/optimization.py
matulni: Done in [975f7aa](https://github.com/TeamGraphix/graphix/pull/393/commits/975f7aa3429340fa0ac28264aaa837de02867eea)

graphix/pattern.py
matulni: Agree, thanks for the remark.

thierry-martinez (approved):
LGTM!

mgarnier59 (approved):
Looks good, nothing to add. Thanks!

Issue #233, opened 2024-11-06
Ensuring consistent angle convention (radians/pi or radians)


EarlMilktea:
When we refer to angle (such as `.angle` in `M`), should we use radians or radians / $\pi$ ?

shinich1:
radians / pi perhaps?

EarlMilktea:
Personally I prefer raw radians.

@shinich1 

Are there any specific reasons to prefer $/\pi$ ?

PR #394, opened 2025-12-09
Fix #352: Limit circuit measurements to Pauli axes

### LINKED ISSUE #352: Circuit measurements should be limited to axes X, Y, and Z
In the discussion for #343, in particular in https://github.com/TeamGraphix/graphix/pull/343#discussion_r2438749887, we noticed that OpenQASM only support Z measurements.

In general, for circuits, we could limit measurements on X, Y, and Z, and give a way to reformulate them as Z measurements only.

thierry-martinez:
This commit implements the suggestion in #352 to restrict circuit measurements to X, Y, and Z axes. It also adds a method `Circuit.transpile_measurements_to_z_axis`, which rewrites a circuit to use only Z-axis measurements, enabling export to OpenQASM.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - #383: Simulators are now parameterized by `PrepareMethod` (which defaults to `DefaultPrepareMethod`) to customize how `N` commands are handled, and the class `BaseN` can be used as a base class for custom preparation commands.
 
+- #394: The method `Circuit.transpile_measurements_to_z_axis` returns an equivalent circuit where all measurements are on the Z axis. This can be used to prepare a circuit for export to OpenQASM with `circuit_to_qasm3`.
+
 ### Fixed
 
 - #379: Removed unnecessary `meas_index` from API for rotation instructions `RZ`, `RY` and `RX`.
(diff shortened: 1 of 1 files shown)


graphix/transpiler.py
```diff
@@ -1011,3 +1035,11 @@ def _transpile_rzz(instructions: Iterable[Instruction]) -> Iterator[Instruction]
             yield instruction.CNOT(control=instr.control, target=instr.target)
         else:
             yield instr
+
+
+def _measurement_of_axis(axis: Axis) -> Measurement:
+    if axis == Axis.X:
+        return Measurement(plane=Plane.XY, angle=0)
+    if axis == Axis.Y:
+        return Measurement(plane=Plane.YZ, angle=pi / 2)
```
matulni: Even if this function is "private", I think it's better to adhere to Graphix convention (docstring of `Measurement`), and pass the parameter `angle` in units of $\pi$ (between 0 and 2).  Would you agree ?

```suggestion
        return Measurement(plane=Plane.YZ, angle=0.5)
```

graphix/transpiler.py
```diff
@@ -486,24 +490,23 @@ def _cnot_command(
...
+            measurement basis
 
         Returns
         -------
         commands : list
             list of MBQC commands
         """
-        return [M(node=input_node, plane=plane, angle=angle)]
+        measurement = _measurement_of_axis(axis)
+        return [M(node=input_node, plane=measurement.plane, angle=measurement.angle / pi)]
```
matulni: See comment above.

```suggestion
        return [M(node=input_node, plane=measurement.plane, angle=measurement.angle)]
```

graphix/transpiler.py
matulni: (I think this is important since we use` _measurement_of_axis` not only to transpile `M` commands, but we also pass its output to the simulator...

thierry-martinez: I fixed some of the conventions in 3f4752c and inserted comments where conversions are needed. Conventions should be unified more generally in a later PR (issue #233).

thierry-martinez:
I fixed the angle conventions in 3f4752c and inlined the `perform_measure` function into `Backend.measure` since it was not used elsewhere.

matulni (approved):
Thanks Thierry, I didn't expect the radians/pi unit issue to extend over so many places...
LGTM!

PR #396, opened 2025-12-10
Remove `BackendState` base class for simulator backend states.

emlynsg:
This PR removes the `BackendState` parent class from the `sim/base_backend` module and downstream classes/methods. The intent of this PR is to generalise the simulator requirements, to allow for future additions as plugins or in the Graphix repository itself.

### Summary of changes
- Generalised methods with `Backend` or `BackendState` outputs have now been split into overloaded functions, with concrete return objects `Statevec | DensityMatrix | MBQCTensorNet` or backend equivalents. This is because we can no longer return `BackendState` (removed) or `Backend[_StateT_co]` (for some typing scope reason I don't quite understand).
- Related to this, the `initialise_backend` function has been extracted from the `PatternSimulator` class for readability.

### For discussion
Does this approach meet our requirements? It will force further Graphix PRs or additional external functions if we want to make use of other backends as plugins.

### DIFF
--- a/tests/test_branch_selector.py
+++ b/tests/test_branch_selector.py
@@ -20,7 +20,7 @@ if TYPE_CHECKING:
     from numpy.random import Generator
 
     from graphix.measurements import Outcome
-    from graphix.sim import _BackendLiteral
+    from graphix.simulator import _BackendLiteral
 
 NB_ROUNDS = 100
 


graphix/sim/base_backend.py
```diff
@@ -528,7 +505,37 @@ def _op_mat_from_result(
...
+
+    def f_expectation0() -> float:
+        exp_val = state.expectation_single(get_op_mat0(), qubit_loc)
+        assert math.isclose(exp_val.imag, 0, abs_tol=1e-10)
+        return exp_val.real
+
+    result = branch_selector.measure(qubit_node, f_expectation0, rng)
+    op_mat = _op_mat_from_result(vec, 1, symbolic=symbolic) if result else get_op_mat0()
+    state.evolve_single(op_mat, qubit_loc)
+    return result
```
thierry-martinez: This function is no longer used.

tests/test_pattern.py
```diff
@@ -31,11 +30,10 @@
 if TYPE_CHECKING:
     from collections.abc import Sequence
 
-    from graphix.fundamentals import Angle
-    from graphix.sim.base_backend import BackendState
+    from graphix.sim import _BackendLiteral
 
 
-def compare_backend_result_with_statevec(backend_state: BackendState, statevec: Statevec) -> float:
+def compare_backend_result_with_statevec(backend_state: object, statevec: Statevec) -> float:
```
mgarnier59: Why is the typing `object` here (and below)?

thierry-martinez: There is no longer a base class for backend states; therefore in general, we can only say that a backend state is an `object`. However, here, the function only supports `Statevec` and `DensityMatrix`, which can be reflected by typing `backend_state : Statevec | DensityMatrix`, provided that the type for `backend` in tests that call this function is `Literal["statevector", "densitymatrix"]`.

tests/test_pattern.py
```diff
@@ -119,14 +117,12 @@ def test_minimize_space_with_gflow(self, fx_bg: PCG64, jumps: int) -> None:
...
-    def test_empty_output_nodes(
-        self, backend_type: typing.Literal["statevector", "densitymatrix", "tensornetwork"]
-    ) -> None:
+    def test_empty_output_nodes(self, backend_type: _BackendLiteral) -> None:
         pattern = Pattern(input_nodes=[0])
         pattern.add(M(node=0, angle=0.5))
 
         def simulate_and_measure() -> int:
-            sim = PatternSimulator(pattern, backend_type)
+            sim: PatternSimulator[object] = PatternSimulator(pattern, backend_type)
```
thierry-martinez: In the same vein as Maxime's comment above, I just realized we can be more precise here: `sim: PatternSimulator[Statevec | DensityMatrix | MBQCTensorNet]`.

tests/test_pattern.py
emlynsg: Thanks both, good catch. I think these were changes I made before our last discussion Thierry, and I missed fixing them up in the tests.

thierry-martinez (approved):
LGTM! Thanks.

mgarnier59 (approved):
No additional comments, thanks! Just need to resolve conflicts and merge.

PR #402, opened 2025-12-22
Support Python 3.14

thierry-martinez:
- Bump the upper Python version bound to <3.15.
- Remove the optional qiskit-aer dependency; it’s currently unused and fails to compile (at least with recent Clang on macOS).
- Constrain NumPy to <2.4 because Numba does not yet support NumPy 2.4: https://github.com/numba/numba/issues/10263

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 matplotlib
 networkx
-numpy>=2,<2.4
+numpy>=2
 opt_einsum
 quimb
 scipy


requirements-dev.txt
```diff
@@ -22,7 +22,7 @@ pytest-mpl
 
 # Optional dependencies
 qiskit>=1.0
-qiskit-aer
+# qiskit-aer
```
matulni: Why commenting instead of removing ?

thierry-martinez: Line removed in 96202e9.

PR #404, opened 2025-12-24
Fix #177: Correct pattern export to OpenQASM 3

### LINKED ISSUE #177: QASM3 export only supports XY plane measurements
**Describe the bug**
only XY plane is supported, and `VOP` is ignored. 
https://github.com/TeamGraphix/graphix/blob/cef611d3b9555867aac8dfbbbdb21b484de01a45/graphix/pattern.py#L2127-L2143

NB: `vop` property of measurement commands will be removed and integrated into meas. plane and change of angles.

thierry-martinez:
This commit introduces `pattern_to_qasm3`, improving the export of patterns to OpenQASM 3:

- Planes YZ and XZ are now handled correctly (#177).

- For plane XY, the angle is no longer ignored when the s-domain is empty.

- Support for |0⟩ and |+⟩ initial states.

- Arithmetic expressions unsupported by `qiskit-qasm3-import` are replaced with a simpler encoding, and the `single_qubit_domains` pass ensures compatibility with Qiskit.

- `test_qasm3_exporter_to_qiskit.py` verifies that the Graphix pattern simulator and Qiskit AER simulator produce equivalent statevectors for exported circuits.


tests/test_qasm3_exporter_to_qiskit.py
```diff
@@ -0,0 +1,128 @@
...
+    check_qasm3(Pattern(input_nodes=[0, 1], cmds=[N(2), E((1, 2))]))
+
+
+@pytest.mark.parametrize("clifford", Clifford)
+@pytest.mark.parametrize("state", [BasicStates.ZERO, BasicStates.PLUS])
+def test_to_qasm3_clifford(clifford: Clifford, state: State) -> None:
+    check_qasm3(Pattern(cmds=[N(0, state), C(0, clifford)]))
+
+
+@pytest.mark.parametrize("state", [BasicStates.ZERO, BasicStates.PLUS])
```
emlynsg, line 100: Add unsupported input state test?

graphix/pattern.py
```diff
@@ -1461,27 +1463,21 @@ def draw_graph(
...
+        See :func:`graphix.qasm3_exporter.pattern_to_qasm3`.
+
         Parameters
         ----------
         filename : Path | str
-            file name to export to. example: "filename.qasm"
+            File name to export to. Example: ``"filename.qasm"``.
+
+        input_state : dict[int, State] | State, default BasicStates.PLUS
+            The initial state for each input node. Only ``|0⟩`` or ``|+⟩`` states are supported.
```
emlynsg, line 1514: May be helpful to catch unsupported input states here with a raise?

graphix/qasm3_exporter.py
```diff
@@ -121,3 +124,139 @@ def instruction_to_qasm3(instruction: Instruction) -> str:
...
+    if cmd.kind == CommandKind.N:
+        yield f"qubit q{cmd.node};\n"
+        yield from state_to_qasm3_lines(cmd.node, cmd.state)
+
+    elif cmd.kind == CommandKind.E:
+        n0, n1 = cmd.nodes
+        yield f"cz q{n0}, q{n1};\n"
+
+    elif cmd.kind == CommandKind.M:
+        yield from domain_to_qasm3_lines(cmd.s_domain, f"x q{cmd.node}")
```
emlynsg, line 199: Add a test for measurements with s/t domains?

tests/test_qasm3_exporter_to_qiskit.py
thierry-martinez: This is now fixed.

graphix/pattern.py
thierry-martinez: I noted the limitation in the docstring for convenience, but I prefer not to duplicate the code and let `pattern_to_qasm3_lines` raise the exception, if appropriate.

graphix/qasm3_exporter.py
thierry-martinez: I separated the test suite into two: `test_qasm3_exporter.py` and `test_qasm3_exporter_to_qiskit.py`, and only the second one requires Qiskit. That let the tests in `test_qasm3_exporter.py` skip the normalization of measurements and exercise that code path.

graphix/qasm3_exporter.py
```diff
@@ -118,3 +121,140 @@ def instruction_to_qasm3(instruction: Instruction) -> str:
...
+    """Export pattern to line-by-line OpenQASM 3.0 representation.
+
+    See :func:`pattern_to_qasm3`.
+    """
+    yield "// generated by graphix\n"
+    yield "OPENQASM 3;\n"
+    yield 'include "stdgates.inc";\n'
+    yield "\n"
+    for node in pattern.input_nodes:
+        yield f"// prepare input qubit {node} in |+⟩\n"
```
matulni: Isn't this misleading, since input qubits may be prepared in |+> or |0> ?
I think it should be moved to the branch `elif state == BasicStates.PLUS:` in `:func: state_to_qasm3_lines`, or alternatively
```python
f"// prepare input qubit {node} in |+⟩ or |0⟩\n"
```

graphix/qasm3_exporter.py
```diff
@@ -118,3 +121,140 @@ def instruction_to_qasm3(instruction: Instruction) -> str:
...
+
+    Yields
+    ------
+    string
+        translated controlled command in OpenQASM 3.0 language
+    """
+    condition = " ^ ".join(f"c{node}" for node in domain)
+    if not condition:
+        return
+    yield f"if ({condition}) {{\n"
```
matulni, line 265: Is it possible that there's one `{` too many?

graphix/qasm3_exporter.py
```diff
@@ -118,3 +121,140 @@ def instruction_to_qasm3(instruction: Instruction) -> str:
...
+        n0, n1 = cmd.nodes
+        yield f"cz q{n0}, q{n1};\n"
+
+    elif cmd.kind == CommandKind.M:
+        yield from domain_to_qasm3_lines(cmd.s_domain, f"x q{cmd.node}")
+        yield from domain_to_qasm3_lines(cmd.t_domain, f"z q{cmd.node}")
+        if cmd.plane == Plane.XY:
+            yield f"h q{cmd.node};\n"
+        if cmd.angle != 0:
+            rad_angle = angle_to_rad(cmd.angle)
```
matulni: Shouldn't we use `angle_to_qasm3` ? (This would prevent errors if the `cmd.angle` is parametric).

graphix/qasm3_exporter.py
```diff
@@ -118,3 +121,140 @@ def instruction_to_qasm3(instruction: Instruction) -> str:
...
+            if cmd.plane == Plane.XY:
+                yield f"rx({-rad_angle}) q{cmd.node};\n"
+            elif cmd.plane == Plane.XZ:
+                yield f"ry({-rad_angle}) q{cmd.node};\n"
+            else:
```
matulni, line 213: To improve readability, I would suggest putting `elif cmd.plane == Plane.YZ` followed by `typing_extensions.assert_never` (or `match`. Actually I'm not sure if there is an advantage on using `match` over chained `if-else`, or if it's just a matter of taste. Would we need `typing_extensions.assert_never` if we were using match ?)

graphix/qasm3_exporter.py
```diff
@@ -118,3 +121,140 @@ def instruction_to_qasm3(instruction: Instruction) -> str:
...
+    """
+    return "".join(pattern_to_qasm3_lines(pattern, input_state=input_state))
+
+
+def pattern_to_qasm3_lines(pattern: Pattern, input_state: dict[int, State] | State = BasicStates.PLUS) -> Iterator[str]:
+    """Export pattern to line-by-line OpenQASM 3.0 representation.
+
+    See :func:`pattern_to_qasm3`.
+    """
+    yield "// generated by graphix\n"
```
matulni: Could it be useful to add graphix version here ? Not sure if we expose it at the moment, if not, I wouldn't bother adding it to this PR.

graphix/qasm3_exporter.py
thierry-martinez: Nice catch! This is now fixed.

graphix/qasm3_exporter.py
thierry-martinez: In f-strings, `{` should be doubled to appear in the final string. https://docs.python.org/3/reference/lexical_analysis.html#f-strings:

> Any doubled curly braces ({{ or }}) outside replacement fields are replaced with the corresponding single curly brace:
> ```
> print(f'{{...}}')
> {...}
> ```

graphix/qasm3_exporter.py
thierry-martinez: Yes, you are right, `angle_to_qasm3` is better here. This is fixed.

graphix/qasm3_exporter.py
thierry-martinez: Fixed. There is no default exhaustiveness check with `match` in Python, any more than with chained `if/elif`: we need to add a default clause, `case e: assert_never(e)`, to ensure exhaustiveness.

graphix/qasm3_exporter.py
thierry-martinez: Good idea. Fixed.

graphix/qasm3_exporter.py
matulni: You are right, thanks!

PR #405, opened 2025-12-24
Fix #231: define `IXYZ = Literal[I] | Axis`

### LINKED ISSUE #231: `Axis` should be a subtype of `IXYZ`
Currently, `Axis` and `IXYZ` are defined as two distinct `Enum` types, leading to a preference for `IXYZ` in contexts where `Axis` would be more appropriate but would require an undesirable cast. See [this discussion](https://github.com/TeamGraphix/graphix/pull/228/files#r1825893149) for instance.

Ideally, we should redefine `IXYZ` as `Literal[I] | Axis`.

thierry-martinez:
This commit replaces the `IXYZ` enum class with a type alias defined as the union `Literal[I] | Axis`, where `I` is defined in the singleton enum `SingletonI`. This change makes `Alias` a subtype of `IXYZ`, removes cumbersome conversions between `IXYZ` and `Axis`, and eliminates some code duplication.

### DIFF
--- a/graphix/_db.py
+++ b/graphix/_db.py
@@ -7,7 +7,7 @@ from typing import NamedTuple
 import numpy as np
 
 from graphix import utils
-from graphix.fundamentals import IXYZ, Sign
+from graphix.fundamentals import Axis, Sign
 from graphix.ops import Ops
 
 # 24 unique 1-qubit Clifford gates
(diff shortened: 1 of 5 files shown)


matulni (approved):
Looks good, thank you!

PR #406, opened 2025-12-24
Fix #253: add classes `BaseCommand` and `BaseInstruction`

### LINKED ISSUE #253: Refacto `command.py` to have a proper `class Command`
Right now, there is a `class _KindChecker` that ensures the existence of a `kind` attribute for the commands. This class is not suited to implement class methods for the commands.

 It may be better to have a proper `class Command` that could inherit from `_KindChecker` in order to keep ensuring the `kind` attribute are setted. This class will be useful to put methods related to converting commands into latex, ascii or unicode string representations.

**Additional context**
This issue is related to [PR#245](https://github.com/TeamGraphix/graphix/pull/245#discussion_r1963320062) and follows the discussion on turning the functions for drawing commands into class methods.

thierry-martinez:
This commit introduces the classes `BaseCommand` and `BaseInstruction`, and removes the unused instructions `_XC` and `_ZC`.  These classes can be used for type checking. They contrast with the type aliases `Command` and `Instruction`:

- `BaseCommand` and `BaseInstruction` are open types and the user can define new subclasses. For instance, Veriphix introduces blind preparation commands and measurements.

- `Command` and `Instruction` are closed types that contain exactly the commands of the measurement calculus (`N | M | E | C | X | Z | S | T`) and the instructions supported by the transpiler.


graphix/instruction.py
matulni: If I understood correctly, we can remove `_XC` and `_ZC` from here as well, right ?
```suggestion
```
Note: `InstructionKind._XC` and `InstructionKind._ZC` appear in `graphix.qasm3_exporter` and `graphix.transpiler`.

mgarnier59:
Thanks! And good idea to generalize to the circuit instructions too. This might come in handy in veriphix.
LGTM.

graphix/instruction.py
thierry-martinez: Fixed in ebd178d.

matulni (approved):
Thanks, LGTM!

PR #410, opened 2026-01-06
Replacement of parametric angles in `OpenGraph`, `PauliFlow` and `XZCorrections`

matulni:
This commit adds the necessary methods to replace parametric angles by numerical values in `OpenGraph`, `PauliFlow` and `XZCorrections` instances.

### DIFF
--- a/tests/test_flow_core.py
+++ b/tests/test_flow_core.py
@@ -35,6 +35,7 @@ from graphix.flow.exceptions import (
 from graphix.fundamentals import ANGLE_PI, AbstractMeasurement, AbstractPlanarMeasurement, Axis, Plane
 from graphix.measurements import Measurement
 from graphix.opengraph import OpenGraph
+from graphix.parameter import Placeholder
 from graphix.pattern import Pattern
 from graphix.states import PlanarState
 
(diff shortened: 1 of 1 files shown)


graphix/flow/core.py
```diff
@@ -603,6 +653,52 @@ def get_measurement_label(self, node: int) -> Plane | Axis:
         """
         return self.og.measurements[node].to_plane_or_axis()
 
+    def subs(  # noqa: PYI019 Annotating with `Self` is not possible since `self` must be of parametric type `Measurement`.
```
thierry-martinez, line 656: Why is `self: PauliFlow[Measurement]` not enough? I think both `PauliFlow` and `Measurement` could be considered final classes.

matulni: I had an issue with the return type: if it is `PauliFlow[Measurement]` the following code is problematic:

```python
alpha = Placeholder("alpha")
value = 0.3
og = OpenGraph(graph=nx.Graph([(0, 1)]), input_nodes=[0], output_nodes=[], measurements={0: Measurement(alpha, Plane.XY), 1: Measurement(0.3, Plane.YZ)})
gflow = og.extract_gflow()

gflow_new: GFlow[Measurement] = gflow.subs(alpha, value) # mypy complains about the annotation.
```

thierry-martinez: Oh, indeed! Sorry, I forgot that `PauliFlow` was not final.

thierry-martinez: Oh, indeed! Sorry, I forgot that `PauliFlow` was not final.

matulni:
> Minor suggestion to add basic tests for xreplace and subs in `PauliFlow` and `XZCorrections` (not just directly with `OpenGraph`). Not a dealbreaker so approving anyway.

Thanks for the review, I implemented the tests in [c4e882a](https://github.com/TeamGraphix/graphix/pull/410/commits/c4e882aa41f68788e6d17f70c66ede4182ce5c9d)

PR #412, opened 2026-01-12
Pretty-print methods for `PauliFlow` and `XZCorrections` objects

matulni:
This commit implements pretty-print methods (`to_ascii`, `to_latex` and `to_unicode`) for `PauliFlow` and `XZCorrections` objects, whereby the correction function(s) and partial order layers are formatted in a string. It also implements the `__str__` method as a call to `self.to_ascii`.

Further, we fix a minor unreported bug where the method `PauliFlow.to_corrections` was not casting the `z_corrections` to `frozenset` and initialized a `set` instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- #396: Removed generic `BackendState` from `graphix.sim` modules and methods in `graphix.pattern` and `graphix.simulator` modules.
+
 - #374: Adapted existing method `graphix.opengraph.OpenGraph.isclose` to the new API introduced in #358.
 
 - #375: Adapted existing method `graphix.opengraph.OpenGraph.compose` to the new API introduced in #358.
--- a/graphix/pattern.py
+++ b/graphix/pattern.py
@@ -14,7 +14,7 @@ from collections.abc import Iterable, Iterator
 from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
-from typing import TYPE_CHECKING, SupportsFloat, TypeVar
+from typing import TYPE_CHECKING, Literal, SupportsFloat, overload
 
 import networkx as nx
 from typing_extensions import assert_never
@@ -28,6 +28,7 @@ from graphix.measurements import Measurement, Outcome, PauliMeasurement, toggle_
 from graphix.opengraph import OpenGraph
 from graphix.pretty_print import OutputFormat, pattern_to_str
 from graphix.qasm3_exporter import pattern_to_qasm3_lines
+from graphix.sim import DensityMatrix, MBQCTensorNet, Statevec
 from graphix.simulator import PatternSimulator
 from graphix.states import BasicStates
 from graphix.visualization import GraphVisualizer
@@ -40,12 +41,19 @@ if TYPE_CHECKING:
     from numpy.random import Generator
 
     from graphix.flow.core import CausalFlow, GFlow, XZCorrections
-    from graphix.parameter import ExpressionOrSupportsFloat, Parameter
-    from graphix.sim import Backend, BackendState, Data
+    from graphix.parameter import ExpressionOrSupportsComplex, ExpressionOrSupportsFloat, Parameter
+    from graphix.sim import (
+        Backend,
+        Data,
+        DensityMatrixBackend,
+        StatevectorBackend,
+    )
+    from graphix.sim.base_backend import _StateT_co
+    from graphix.sim.tensornet import TensorNetworkBackend
+    from graphix.simulator import _BackendLiteral
     from graphix.states import State
 
-
-_StateT_co = TypeVar("_StateT_co", bound="BackendState", covariant=True)
+_BuiltinBackendState = DensityMatrix | Statevec | MBQCTensorNet
 
 
 class Pattern:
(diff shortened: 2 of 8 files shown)


graphix/flow/core.py
```diff
@@ -354,6 +355,38 @@ def check_well_formed(self) -> None:
...
+        multiline : bool (optional)
+            Flag to format the output. If ``True`` a new line is printed after each correction set. Defaults to ``False``.
+
+        Returns
+        -------
+        str
+        """
+        return xzcorr_to_str(self, output=OutputFormat.ASCII, multiline=multiline)
+
+    def to_latex(self, multiline: bool = False) -> str:
```
emlynsg, line 346: Can someone double check this command on a couple of tests on their machine? I am getting doubleslash (`\\`) before the `prec` commands.

matulni: Thanks for the review @emlynsg .
On VSCode 1.107.1 (MacOS) running a Jupyter notebook (plugin version 2025.9.1) the following code
```python
og = OpenGraph(
        graph=nx.Graph([(1, 3), (2, 4), (3, 4), (3, 5), (4, 6)]),
        input_nodes=[1, 2],
        output_nodes=[5, 6],
        measurements={
            1: Measurement(0.1, Plane.XY),
            2: Measurement(0.2, Plane.XY),
            3: Measurement(0.3, Plane.XY),
            4: Measurement(0.4, Plane.XY),
        },
    )
c = og.extract_causal_flow()
print(c.to_latex())
```

gives this output
```
c(3) = \{5\}, \;c(4) = \{6\}, \;c(1) = \{3\}, \;c(2) = \{4\},\\\{1, 2\} \prec \{3, 4\} \prec \{5, 6\}
```

which can be rendered in a Jupyter notebook as follows:

```python
from IPython.display import display, Math
display(Math(c.to_latex()))
```
Output:
image


I'm a bit puzzled by your example... do you have similar problems when using `Pattern.to_latex()`?

graphix/pretty_print.py
```diff
@@ -233,3 +238,129 @@ def pattern_to_str(
...
+        If ``True``, format each correction set on a separate line (or LaTeX line break).
+        If ``False``, format each correction set on a single line separated by commas.
+        Default is ``False``.
+
+    Returns
+    -------
+    str
+        A string representation of the flow object formatted according to the specified output format and layout.
+    """
+    separator = r",\\" if output == OutputFormat.LaTeX else "\n"
```
thierry-martinez: The trailing comma inserted in LaTeX does not appear in other output formats. Is it done on purpose?
Moreover, it is a bit surprising that the output is multiline even if `multiline = False`. I believe I would prefer it if `multiline = False` used for instance `,` to separate corrections (as it is currently done), and `;` to separate main components.

tests/test_pretty_print.py
```diff
@@ -83,3 +93,75 @@ def test_pattern_pretty_print_random(fx_bg: PCG64, jumps: int, output: OutputFor
...
+            3: Measurement(0.3, Plane.XY),
+            4: Measurement(0.4, Plane.XY),
+        },
+    )
+
+
+def test_cflow_str() -> None:
+    flow = example_og().extract_causal_flow()
+
+    assert str(flow) == "c(3) = {5}, c(4) = {6}, c(1) = {3}, c(2) = {4}\n{1, 2} < {3, 4} < {5, 6}"
```
thierry-martinez: It would be nice if there were at least one test that checks every output format.

```suggestion
    assert str(flow) == "c(3) = {5}, c(4) = {6}, c(1) = {3}, c(2) = {4}\n{1, 2} < {3, 4} < {5, 6}"
    assert flow.to_latex() == r"c(3) = \{5\}, \;c(4) = \{6\}, \;c(1) = \{3\}, \;c(2) = \{4\},\\\{1, 2\} \prec \{3, 4\} \prec \{5, 6\}"
    assert flow.to_unicode() == "c(3) = {5}, c(4) = {6}, c(1) = {3}, c(2) = {4}\n{1, 2} ≺ {3, 4} ≺ {5, 6}"
```

graphix/pretty_print.py
```diff
@@ -233,3 +238,129 @@ def pattern_to_str(
...
+
+    return separator.join(
+        (
+            correction_function_to_str(xzcorr.x_corrections, "x", output, multiline),
+            correction_function_to_str(xzcorr.z_corrections, "z", output, multiline),
+            partial_order_to_str(xzcorr.partial_order_layers, output),
+        )
+    )
```
thierry-martinez: The logic follows the same pattern as `flow_to_str`. Perhaps it is worth factorizing.

graphix/flow/core.py
thierry-martinez: I am also unable to reproduce the double slash reported by @emlynsg. @emlynsg, could you express your problem as a Python assertion that fails, or does it happen only when you interact with your IDE?

graphix/pretty_print.py
```diff
@@ -233,3 +238,129 @@ def pattern_to_str(
...
+        else (r", \;" if output == OutputFormat.LaTeX else ", ")
+    )
+
+    if output == OutputFormat.LaTeX:
+        return separator.join(
+            f"{cf_name}({node}) = " + r"\{{{}\}}".format(", ".join(map(str, cset)))
+            for node, cset in correction_function.items()
+        )
+
+    return separator.join(f"{cf_name}({node}) = { {*cset} }" for node, cset in correction_function.items())
```
thierry-martinez: I suggest introducing the following function:
```python
def set_to_str(objects: Iterable[object], output: OutputFormat) -> str:
    """Convert a set to a formatted string representation.

    Parameters
    ----------
    objects : Iterable[object]
        The set to format.
    output : OutputFormat
        The desired output format (ASCII, LaTeX or Unicode).
    """
    contents = ", ".join(str(item) for item in objects)
    if output == OutputFormat.LaTeX:
        return f"\\{{{contents}\\}}"
    return f"{{{contents}}}"
```

and then the code becomes:
```suggestion
    return separator.join(f"{cf_name}({node}) = {set_to_str(cset, output)}" for node, cset in correction_function.items())
```

graphix/pretty_print.py
```diff
@@ -233,3 +238,129 @@ def pattern_to_str(
...
+    output : OutputFormat
+        The desired output format (ASCII, LaTeX or Unicode).
+
+    Returns
+    -------
+    str
+    """
+    match output:
+        case OutputFormat.ASCII:
+            sep = " < "
```
thierry-martinez: This variable is called `separator` in the other functions.

graphix/pretty_print.py
```diff
@@ -233,3 +238,129 @@ def pattern_to_str(
...
+        case OutputFormat.Unicode:
+            sep = " ≺ "
+        case OutputFormat.LaTeX:
+            sep = r" \prec "
+        case _:
+            assert_never(output)
+
+    if output == OutputFormat.LaTeX:
+        return sep.join(r"\{{{}\}}".format(", ".join(map(str, layer))) for layer in partial_order_layers[::-1])
+    return sep.join(f"{ {*layer} }" for layer in partial_order_layers[::-1])
```
thierry-martinez: Using the `set_to_str` function proposed above:
```suggestion
    return sep.join(set_to_str(layer, output) for layer in partial_order_layers[::-1])
```

tests/test_pretty_print.py
matulni: Thanks for taking the time to write that down, added in [3c210d5](https://github.com/TeamGraphix/graphix/pull/412/commits/3c210d5ca0d236232fd8bba7709dd7901069a12e)

graphix/pretty_print.py
matulni: I added a comma because in papers (at least in the field of condensed matter) it's customary to add a comma between equation lines, but I thought it was not necessary in ASCII as it it's done for rapid debugging.  If you think it's best to be fully homogeneous, I'd remove it from the LaTeX output.

As for the `multiline` flag, I agree, I'll implement it. I thought it was a pity that `print(a_flow_instance)` would put everything on the same line because it's not very readable, but I suppose is less surprising this way, and better for logging.

graphix/pretty_print.py
matulni: Done in [3c210d5](https://github.com/TeamGraphix/graphix/pull/412/commits/3c210d5ca0d236232fd8bba7709dd7901069a12e)

graphix/pretty_print.py
matulni: Fixed in [3c210d5](https://github.com/TeamGraphix/graphix/pull/412/commits/3c210d5ca0d236232fd8bba7709dd7901069a12e)

graphix/pretty_print.py
matulni: Thanks for the suggestion, it looks much cleaner now. Fixed in [3c210d5](https://github.com/TeamGraphix/graphix/pull/412/commits/3c210d5ca0d236232fd8bba7709dd7901069a12e)

graphix/pretty_print.py
matulni: I did some partial refactor in [3c210d5](https://github.com/TeamGraphix/graphix/pull/412/commits/3c210d5ca0d236232fd8bba7709dd7901069a12e)

Unfortunately, the "name" of the `x_corrections` and `z_corrections` correction functions is hardcoded in the function body, so it's not obvious to me how to encapsulate both functions together. Did you have something else in mind ?

graphix/pretty_print.py
thierry-martinez: Alternatively, you could add a trailing comma for all output modes. In summary, a semicolon in single-line mode is replaced by a comma and a line-break in multi-line mode.

emlynsg (approved):
As discussed there may be problems on Windows but it isn't important - just use WSL.
Approved.

thierry-martinez:
@emlynsg I'm fine with the idea that Windows issues should not block this PR, but could you document how to reproduce these problems, either here or in a new issue?

PR #418, opened 2026-01-22
Fix #220: rename `get_`/`set_`-prefixed functions

### LINKED ISSUE #220: Avoid `get_` / `set_`

thierry-martinez:
This commit renames functions that use `get_` / `set_` prefixes in order to make function names more descriptive.

The use of `get_`/`set_` prefixes is widespread in Python codebases, including in external dependencies such as Qiskit (`get_memory`, `get_statevector`) and the Python standard library (`typing.get_origin`, `typing.get_args`).

The renamings in this commit aim to better reflect the semantics of the corresponding functions (e.g., whether they perform a computation, a lookup, or a transformation), without changing behavior.

There is a single behavioral change:

- `extract_measurement_commands` replaces `get_angles` and `get_meas_planes` and now returns a dictionary, instead of the subsequence of commands.

The other renamings are:

- `graph_to_fusion_network` instead of `get_fusion_network_from_graph`;

- `state_to_statevector` instead of `get_statevector`, with a `State` or a more general type as parameter;

- `to_statevector` instead of `get_statevector` and `to_densitymatrix` instead of `get_densitymatrix` in `State`;

- in resource graphs, `fusion_nodes` instead of `get_fusion_nodes`;

- in tensor networks, `basis_amplitude` instead of `get_basis_amplitude`;

- `basis_coefficient` instead of `get_basis_coefficient`;

- `open_tensor` instead of `get_open_tensor_from_index`;

- in flows, `measurement_label` instead of `get_measurement_label`;

- in graph sim, `isolates` instead of `get_isolates`; the new name matches the underlying `nx.isolate`;

- `_norm` instead of `_get_statevec_norm`;

- in pretty-printing, `component_separator_for` instead of `get_component_separator`;

- in measurement methods, `measurement_outcome` instead of `get_measure_result`;

- `store_measurement_outcome` instead of `set_measure_result`;

- `describe_measurement` instead of `get_measurement_description`;

- in tensor networkds, `prepare_graph_state` instead of `set_graph_state`;

- in type conversion for density matrices, `cast_row` instead of `get_row`;

- in patterns and graph sim, `extract_clifford` instead of `get_vops`;

- `place_flow`, `place_gflow`, `place_paths`, `place_edge_paths`, `place_wo_structure`, and `place_all_corrections` instead of `get_pos_from_flow`, `get_pos_from_gflow`, `get_paths`, `get_edge_path`, `get_pos_wo_structure`, and `get_pos_all_correction`;

- `determine_figsize` instead of `get_figsize`;

- `_outcome_to_op_mat` instead of `_op_mat_from_result`;

- `compute_op_mat0` instead of `get_op_mat0`; even though `compute` can be as vague as `get`, `compute_op_mat0` emphasizes in this context that the computation of the matrix operator is lazily deferred;

- `_extract_dependency` instead of `_get_dependency`;

- `_decompose_cz` instead of `_get_decomposed_cz`;

- `set_noise_model` is removed, as it merely encapsulated a direct assignment to the `noise_model` field.

### DIFF
--- a/graphix/pattern.py
+++ b/graphix/pattern.py
@@ -43,12 +43,7 @@ if TYPE_CHECKING:
 
     from graphix.flow.core import CausalFlow, GFlow, XZCorrections
     from graphix.parameter import ExpressionOrSupportsComplex, ExpressionOrSupportsFloat, Parameter
-    from graphix.sim import (
-        Backend,
-        Data,
-        DensityMatrixBackend,
-        StatevectorBackend,
-    )
+    from graphix.sim import Backend, Data, DensityMatrixBackend, StatevectorBackend
     from graphix.sim.base_backend import _StateT_co
     from graphix.sim.tensornet import TensorNetworkBackend
     from graphix.simulator import _BackendLiteral
@@ -113,7 +108,6 @@ class Pattern:
         else:
             self.__input_nodes = list(input_nodes)  # input nodes (list() makes our own copy of the list)
         self.__n_node = len(self.__input_nodes)  # total number of nodes in the graph state
-        self._pauli_preprocessed = False  # flag for `measure_pauli` preprocessing completion
 
         self.__seq = []
         # output nodes are initially a copy input nodes, since none are measured yet
@@ -135,7 +129,6 @@ class Pattern:
         cmd : :class:`graphix.command.Command`
             MBQC command.
         """
-        self._pauli_preprocessed = False
         if cmd.kind == CommandKind.N:
             self.__n_node += 1
             self.__output_nodes.append(cmd.node)
@@ -149,7 +142,6 @@ class Pattern:
 
         :param cmds: sequences of commands
         """
-        self._pauli_preprocessed = False
         for item in cmds:
             if isinstance(item, Iterable):
                 for cmd in item:
@@ -159,7 +151,6 @@ class Pattern:
 
     def clear(self) -> None:
         """Clear the sequence of pattern commands."""
-        self._pauli_preprocessed = False
         self.__n_node = len(self.__input_nodes)
         self.__seq = []
         self.__output_nodes = list(self.__input_nodes)
@@ -171,7 +162,6 @@ class Pattern:
 
         :param input_nodes: optional, list of input qubits (by default, keep the same input nodes as before)
         """
-        self._pauli_preprocessed = False
         if input_nodes is not None:
             self.__input_nodes = list(input_nodes)
         self.clear()
@@ -217,7 +207,6 @@ class Pattern:
         - Input (and, respectively, output) nodes in the returned pattern have the order of the pattern `self` followed by those of the pattern `other`. Merged nodes are removed.
         - If `preserve_mapping = True` and :math:`|M_1| = |I_2| = |O_2|`, then the outputs of the returned pattern are the outputs of pattern `self`, where the nth merged output is replaced by the output of pattern `other` corresponding to its nth input instead.
         """
-        self._pauli_preprocessed = False
         nodes_p1 = self.extract_nodes() | self.results.keys()  # Results contain preprocessed Pauli nodes
         nodes_p2 = other.extract_nodes() | other.results.keys()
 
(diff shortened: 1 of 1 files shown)


graphix/flow/_find_gpflow.py
```diff
@@ -178,7 +178,7 @@ def _og_matrices(self) -> tuple[MatGF2, MatGF2]:
 
         return flow_demand_matrix, order_demand_matrix
 
-    def _get_measurement_label(self, node: int) -> Plane | Axis:
+    def _measurement_label(self, node: int) -> Plane | Axis:
```
emlynsg: I think this function is still a little unclear since it operates on a node not the whole AOG. I suggest `node_measurement_label`.

graphix/sim/base_backend.py
```diff
@@ -464,17 +464,17 @@ def apply_noise(self, qubits: Sequence[int], noise: Noise) -> None:  # noqa: ARG
         raise NoiseNotSupportedError
 
 
-def _op_mat_from_result(
-    vec: tuple[ExpressionOrFloat, ExpressionOrFloat, ExpressionOrFloat], result: Outcome, symbolic: bool = False
+def _outcome_to_op_mat(
```
emlynsg: I would prefer to write `outcome_to_operator_matrix` in full in these cases, for readability.

graphix/graphsim.py
```diff
@@ -510,6 +510,6 @@ def to_statevector(self) -> Statevec:
                 gstate.evolve_single(Ops.H, i)
         return gstate
 
-    def get_isolates(self) -> list[int]:
+    def isolates(self) -> list[int]:
```
emlynsg: I know this is the nx terminology, but would `isolated_nodes` not be more descriptive?

graphix/graphsim.py
```diff
@@ -139,7 +139,7 @@ def apply_vops(self, vops: Mapping[int, Clifford]) -> None:
                 else:
                     raise RuntimeError
 
-    def get_vops(self) -> dict[int, Clifford]:
+    def extract_vops(self) -> dict[int, Clifford]:
```
emlynsg, line 142: I like that this matches the attribute name `vops`, but why don't we just call both `clifford_operations`?

graphix/visualization.py
```diff
@@ -134,43 +134,43 @@ def visualize(
...
+                def place_paths(
                     pos: Mapping[int, _Point],
                 ) -> tuple[Mapping[_Edge, Sequence[_Point]], Mapping[_Edge, Sequence[_Point]] | None]:
-                    return self.get_edge_path(cf, pos)
+                    return self.place_edge_paths(cf, pos)
 
             else:
                 print("No flow or gflow detected in the graph.")
-                pos = self.get_pos_wo_structure()
+                pos = self.place_wo_structure()
```
emlynsg: I assume `wo` here stands for without, but it is confusing. We could just use `no` instead if writing `without` in full is too long.

thierry-martinez: Replaced by `without`.

graphix/graphsim.py
thierry-martinez: Fixed.

graphix/sim/base_backend.py
thierry-martinez: Fixed.

graphix/flow/_find_gpflow.py
thierry-martinez: Fixed.

graphix/graphsim.py
thierry-martinez: As discussed in the Qode meeting, the terminology *vertex operators* comes from the fact that only a subset of the Clifford gates can appear in the graph simulator. Since the status of the `graphsim` module is unclear, I prefer to leave it untouched for now.

Issue #413, opened 2026-01-13
Should always `copy` in `pattern.py` `measure_pauli` correctly.


emlynsg:
- Mutating in-place a pattern using `measure_pauli` is unused.
- Line 1699 in `pattern.py` should be `pattern.results` not `pat.results`.
- `pauli_preprocessed` flag is unusued throughout `pattern.py`.
Recommend PR that corrects the above.

PR #421, opened 2026-01-30
Removed `_pauli_preprocessed` flag from `Pattern` class functions.

emlynsg:
Building on PR #392 and resolving the last of issue #413 , this PR removes the `_pauli_preprocessed` flag from Graphix entirely. Functions now will proceed with their logic whatever the preprocessed status.

For discussion:
* This means `extract_causal_flow` proceeds without checking if Pauli preprocessing has occurred. Is this a problem, given that some extraction will proceed without raising a `FlowError`?

### DIFF
--- a/graphix/pattern.py
+++ b/graphix/pattern.py
@@ -42,11 +42,7 @@ if TYPE_CHECKING:
     from numpy.random import Generator
 
     from graphix.flow.core import CausalFlow, GFlow, XZCorrections
-    from graphix.parameter import (
-        ExpressionOrSupportsComplex,
-        ExpressionOrSupportsFloat,
-        Parameter,
-    )
+    from graphix.parameter import ExpressionOrSupportsComplex, ExpressionOrSupportsFloat, Parameter
     from graphix.sim import Backend, Data, DensityMatrixBackend, StatevectorBackend
     from graphix.sim.base_backend import _StateT_co
     from graphix.sim.tensornet import TensorNetworkBackend


graphix/pattern.py
```diff
@@ -42,7 +42,11 @@
     from numpy.random import Generator
 
     from graphix.flow.core import CausalFlow, GFlow, XZCorrections
-    from graphix.parameter import ExpressionOrSupportsComplex, ExpressionOrSupportsFloat, Parameter
+    from graphix.parameter import (
+        ExpressionOrSupportsComplex,
+        ExpressionOrSupportsFloat,
+        Parameter,
+    )
```
thierry-martinez: `ruff` keeps formatting the list of symbols across multiple lines because of the trailing comma after `Parameter`. If you remove it, the list fits on a single line.

```suggestion
    from graphix.parameter import ExpressionOrSupportsComplex, ExpressionOrSupportsFloat, Parameter
```

The same comment applies to the next line (although it is not part of the changes in this PR), and on other occurrences in this PR (for instance, line 383 and below): what is the reason of these changes?

emlynsg: Hi Thierry. Sorry, I think whatever autoformat is on in my VSCode just went a bit crazy - I'll investigate and fix up the PR.

emlynsg: Confirming it's an issue where a new install of VSCode (in my case playing around with Docker containers) can default to format on save and use a different formatter to Ruff. Should be fixed now.

thierry-martinez:
There is currently a CI failure, but I suppose it is transient and will disappear once it is restarted. I suspect there is a test in veriphix that has a non-negligible probability of failing even if the code is correct. See https://github.com/TeamGraphix/graphix/actions/runs/21586522415/job/62195929339?pr=421#step:6:780

Thank you for having fixed the code formatting issue in the line I spotted, ~but the same issue has other occurrences in the same file. Can you fix them as well?~ **Edit: My bad, everything is properly fixed!**

PR #423, opened 2026-02-04
Fix #181: Structural separation of Pauli measurements

### LINKED ISSUE #181: Add Pauli plane measurements

thierry-martinez:
This commit introduces a structural separation between Pauli measurements and arbitrary measurements. Technically, the class `Measurement` is now abstract and has two concrete subclasses: `PauliMeasurement` and `BlochMeasurement`. In patterns,  `M` commands are now parameterized by an instance `Measurement` (instead of carrying a plane and an angle).

With these changes, an M command on an arbitrary measurement is written in the following form:
```python
M(node, BlochMeasurement(angle, plane), s_domain, t_domain)
```
and an M command on a Pauli measurement is written as:
```python
M(node, PauliMeasurement(axis, sign), s_domain, t_domain)
```

There are convenient notations for common cases:
- for an arbitrary measurement on a statically known plane, the user can use `Measurement.XY(angle)`, `Measurement.YZ(angle)`, or `Measurement.XZ(angle)`;
- for a positive Pauli measurement on a statically known axis, the user can use `Measurement.X`, `Measurement.Y`, `Measurement.Z`;
- for a negative Pauli measurement on a statically known axis, the user can use `-Measurement.X`, `-Measurement.Y`, `-Measurement.Z`.

By default, `M(node)` is `M(node, Measurement.X)`.

There is no implicit conversion between Pauli measurements represented as `PauliMeasurement` and measurements represented as `BlochMeasurement`.  `Measurement` class provides methods to perform these conversions explicitly:

- `to_bloch()` returns a `BlochMeasurement` equivalent to the given measurement; note that there are multiple possible Bloch descriptions for the same Pauli measurement. `to_bloch()` uses `Plane.XY` for `X` and `Y` and `Plane.YZ` for `Z`.
- `try_to_pauli()` returns a `PauliMeasurement` equivalent to the given measurement if it exists (up to a precision error specified by optional parameters `rel_tol` and `abs_tol`), or `None` otherwise.
- `to_pauli_or_bloch()` is similar to `try_to_pauli()`, but returns the original measurement instead of `None` in the non-Pauli case.

There are conversion functions at the level of `Pattern`, `StandardizedPattern`, and `OpenGraph` to convert all measurements:

- `to_bloch()` converts all measurements to `BlochMeasurement`;
- `infer_pauli_measurements()` converts all measurements that are Pauli (up to a precision error) into `PauliMeasurement`.

In particular, Pauli presimulation, flow extraction algorithms and visualization routines do not perform any implicit conversion:

- Pauli presimulation only applies on `M` commands carrying a `PauliMeasurement`;

- causal flows and gflows may only exist if all measurements are `BlochMeasurement`;

- Pauli flows considers a measurement to be a Pauli measurement only if it carries a `PauliMeasurement`;

- visualization routines now take an `OpenGraph` directly instead of taking planes and angles separately, and displayed Pauli nodes and labels reflect whether the measurement is a `BlochMeasurement` or a `PauliMeasurement`. In particular, since there is no Pauli-flow extraction from pattern, visualization of flows extracted from patterns only work if all measurements are `BlochMeasurement` (and no Pauli nodes will be displayed).

### DIFF
--- a/graphix/opengraph.py
+++ b/graphix/opengraph.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Generic, TypeVar
+from warnings import warn
 
 import networkx as nx
 
@@ -11,13 +12,12 @@ from graphix.flow._find_cflow import find_cflow
 from graphix.flow._find_gpflow import AlgebraicOpenGraph, PlanarAlgebraicOpenGraph, compute_correction_matrix
 from graphix.flow.core import GFlow, PauliFlow
 from graphix.fundamentals import AbstractMeasurement, AbstractPlanarMeasurement
-from graphix.measurements import Measurement
+from graphix.measurements import BlochMeasurement, Measurement
 
 if TYPE_CHECKING:
     from collections.abc import Callable, Collection, Iterable, Mapping, Sequence
 
     from graphix.flow.core import CausalFlow
-    from graphix.measurements import BlochMeasurement
     from graphix.parameter import ExpressionOrSupportsFloat, Parameter
     from graphix.pattern import Pattern
 
@@ -87,8 +87,14 @@ class OpenGraph(Generic[_AM_co]):
         if len(outputs) != len(self.output_nodes):
             raise OpenGraphError("Output nodes contain duplicates.")
 
-    def to_pattern(self: OpenGraph[Measurement]) -> Pattern:
+    def to_pattern(self: OpenGraph[Measurement], *, stacklevel: int = 1) -> Pattern:
         """Extract a deterministic pattern from an `OpenGraph[Measurement]` if it exists.
+
+        Parameters
+        ----------
+        stacklevel : int, optional
+            Stack level to use for warnings. Defaults to 1, meaning that warnings
+            are reported at this function's call site.
 
         Returns
         -------
@@ -118,7 +124,7 @@ class OpenGraph(Generic[_AM_co]):
         else:
             flow = bloch_case.find_causal_flow()
         if flow is None:
-            flow = self.find_pauli_flow()
+            flow = self.find_pauli_flow(stacklevel=stacklevel + 1)
         if flow is not None:
             return flow.to_corrections().to_pattern()
         raise OpenGraphError("The open graph does not have flow. It does not support a deterministic pattern.")
@@ -392,10 +398,16 @@ class OpenGraph(Generic[_AM_co]):
             raise OpenGraphError("The open graph does not have a gflow.")
         return gf
 
-    def extract_pauli_flow(self: OpenGraph[_AM_co]) -> PauliFlow[_AM_co]:
+    def extract_pauli_flow(self: OpenGraph[_AM_co], *, stacklevel: int = 1) -> PauliFlow[_AM_co]:
         r"""Try to extract a maximally delayed Pauli on the open graph.
 
         This method is a wrapper over :func:`OpenGraph.find_pauli_flow` with a single return type.
+
+        Parameters
+        ----------
+        stacklevel : int, optional
+            Stack level to use for warnings. Defaults to 1, meaning that warnings
+            are reported at this function's call site.
 
         Returns
         -------
@@ -411,7 +423,7 @@ class OpenGraph(Generic[_AM_co]):
         --------
         :func:`OpenGraph.find_pauli_flow`, :func:`OpenGraph.infer_pauli_measurements`
         """
-        pf = self.find_pauli_flow()
+        pf = self.find_pauli_flow(stacklevel=stacklevel + 1)
         if pf is None:
             raise OpenGraphError("The open graph does not have a Pauli flow.")
         return pf
@@ -476,8 +488,14 @@ class OpenGraph(Generic[_AM_co]):
             correction_matrix
         )  # The constructor returns `None` if the correction matrix is not compatible with any partial order on the open graph.
 
-    def find_pauli_flow(self: OpenGraph[_AM_co]) -> PauliFlow[_AM_co] | None:
+    def find_pauli_flow(self: OpenGraph[_AM_co], *, stacklevel: int = 1) -> PauliFlow[_AM_co] | None:
         r"""Return a maximally delayed Pauli on the open graph if it exists.
+
+        Parameters
+        ----------
+        stacklevel : int, optional
+            Stack level to use for warnings. Defaults to 1, meaning that warnings
+            are reported at this function's call site.
 
         Returns
         -------
@@ -514,6 +532,7 @@ class OpenGraph(Generic[_AM_co]):
         >>> str(og.infer_pauli_measurements().extract_pauli_flow())
         'p(0) = {1}, p(1) = {2}; {0, 1} < {2}'
         """
+        self._warn_non_inferred_pauli_measurements(stacklevel=stacklevel + 1)
         aog = AlgebraicOpenGraph(self)
         correction_matrix = compute_correction_matrix(aog)
         if correction_matrix is None:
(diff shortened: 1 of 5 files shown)


graphix/simulator.py
```diff
@@ -302,7 +308,10 @@ def run(self, input_state: Data = BasicStates.PLUS, rng: Generator | None = None
         # to catch these errors before starting the simulation.
         self.pattern.check_runnability()
 
+        logger.debug("Initial state: %s", self.backend.state)
```
shinich1, line 324: I guess these logger-related lines are to be removed?

graphix/opengraph.py
```diff
@@ -110,19 +112,71 @@ def to_pattern(self: OpenGraph[Measurement]) -> Pattern:
...
+        else:
+            flow = bloch_case.find_causal_flow()
+        if flow is None:
+            flow = self.find_pauli_flow()
+        if flow is not None:
+            return flow.to_corrections().to_pattern()
         raise OpenGraphError("The open graph does not have flow. It does not support a deterministic pattern.")
 
-    def isclose(self, other: OpenGraph[_M_co], rel_tol: float = 1e-09, abs_tol: float = 0.0) -> bool:
+    def map(self: OpenGraph[_A], f: Callable[[_A], _B]) -> OpenGraph[_B]:
```
shinich1, line 132: are the `map` methods tested?

graphix/simulator.py
thierry-martinez: I suggest keeping them and, more generally, equipping the code with more logging. I think it is a better practice than ad-hoc print debugging: the logging API is convenient for developers to enable the logs they want when debugging, it should be close to zero cost when logging is disabled, and it is less error prone than adding and removing `print` lines repeatedly.

graphix/opengraph.py
thierry-martinez: They were indirectly tested because they were used by `to_bloch`, `downcast_bloch`, and `infer_pauli_measurements`, which are tested. However, I have just added more documentation in the docstrings, including examples that are tested with `pytest --doctest-modules`.

thierry-martinez:
I added more documentation in `measurements.py` and other impacted modules, and the examples are tested with `pytest --doctest-modules`. In particular, test coverage for `measurements.py` is now 100%.

graphix/opengraph.py
```diff
@@ -2,33 +2,35 @@
...
     from graphix.parameter import ExpressionOrSupportsFloat, Parameter
     from graphix.pattern import Pattern
 
 # TODO: Maybe move these definitions to graphix.fundamentals and graphix.measurements ? Now they are redefined in graphix.flow._find_gpflow, not very elegant.
-_M_co = TypeVar("_M_co", bound=AbstractMeasurement, covariant=True)
+_AM_co = TypeVar("_AM_co", bound=AbstractMeasurement, covariant=True)
 _PM_co = TypeVar("_PM_co", bound=AbstractPlanarMeasurement, covariant=True)
+_M = TypeVar("_M", bound=Measurement)
+_A = TypeVar("_A", bound=AbstractMeasurement)
+_B = TypeVar("_B", bound=AbstractMeasurement)
```
emlynsg, line 29: Should we centrally define and import these TypeVars that are used across modules?

graphix/visualization.py
```diff
@@ -146,20 +107,22 @@ def place_paths(
                 return self.place_edge_paths(cf, pos)
 
         else:
-            g_flow = og.find_gflow()
-            if g_flow is not None:
-                print("Gflow detected in the graph. (flow not detected)")
-                pos = self.place_gflow(g_flow)
-                cf = g_flow.correction_function
-                l_k = {node: layer_idx for layer_idx, layer in enumerate(g_flow.partial_order_layers) for node in layer}
+            pauli_flow = self.og.find_pauli_flow()
```
emlynsg, line 110: Can you explain why we have switched from gflow to Pauli flow? Would it not be better to specify the strongest present flow?

tests/baseline/test_draw_graph_reference_True.png
emlynsg, line 1: I think it was clearer when the axis measurements were labelled as such in these plots.

tests/test_pyzx.py
```diff
@@ -74,15 +74,15 @@ def test_random_clifford_t() -> None:
         assert_reconstructed_pyzx_graph_equal(g)
 
 
-@pytest.mark.parametrize("jumps", range(1, 11))
+@pytest.mark.parametrize("jumps", range(1, 2))
```
emlynsg: Is there a reason to reduce the number of jumps?

graphix/opengraph.py
thierry-martinez: These type variables are defined in an ad hoc manner. For instance, there are two type variables bounded by `AbstractMeasurement` because `map()` needs them. Therefore, I prefer to define them locally. Ideally, I would prefer to use the type parameter syntax introduced by PEP 695, but we will have to wait until October 2027 for Python 3.11 to be deprecated.

graphix/visualization.py
thierry-martinez: Actually, now that Pauli measurements are separated from the others, `find_pauli_flow` is just an extension of the `find_gflow` function for graphs with Pauli measurements. On graphs where all measurements are represented as `BlochMeasurement` instances, they behave the same, and `find_gflow` does not support graphs with Pauli measurements. Therefore, on graphs where all measurements are represented as `BlochMeasurement` instances, the visualization will behave as before, and on graphs with Pauli measurements, the visualization will now find a Pauli flow. Previously, the visualization only looked for a gflow, and Pauli measurements were interpreted as planar measurements. We can still reproduce this behavior now by calling `to_bloch()` before visualizing.

tests/baseline/test_draw_graph_reference_True.png
thierry-martinez: I agree, but these labels were misleading: the nodes appeared as Pauli nodes, whereas the flow was extracted as gflow (which is also a correct Pauli flow, but the flow extraction works only on graphs with gflow). I opened issue #432 to implement Pauli flow extraction and visualize the graph with Pauli nodes and Pauli flow.

tests/test_pyzx.py
thierry-martinez: Oops, nice catch: I just forgot to restore the number of jumps after debugging. Fixed in a231b91.

graphix/fundamentals.py
```diff
@@ -294,6 +294,10 @@ def to_plane(self) -> Plane:
         Plane
         """
 
+    @override
+    def to_plane_or_axis(self) -> Plane | Axis:
```
matulni: I'm confused, why does the abstract class `AbstractPlanarMeasurement` need to implement this method? 
Also, shouldn't be the return type just `Plane` ?

graphix/flow/core.py
matulni: Would it be possible to rename this variable as `_AM_co` for consistency with the `opengraph.py` module (it's a bit of nitpicking, but I think it would improve readability) ? (Or alternatively to define it locally if you think it's more appropriate following [@emlynsg 's comment](https://github.com/TeamGraphix/graphix/pull/423#discussion_r2804634116) ).

graphix/fundamentals.py
thierry-martinez: Yes, the return type can be narrowed to `Plane`: fixed in a14cb59. There is no need for an abstract class to implement any method, but in this case, I think it makes sense to have a default implementation of `to_plane_or_axis` in `AbstractPlanarMeasurement` that just calls `to_plane`. It is more convenient for derived classes just to implement `to_plane` and have `to_plane_or_axis` for free. For instance, in `BlochMeasurement`, I did not reimplement `to_plane_or_axis`. I didn't remove `to_plane_or_axis` from `Plane` because it was not incorrect to leave it there, but I could remove it.

graphix/flow/core.py
thierry-martinez: Done in a14cb59, with local redefinition.

graphix/fundamentals.py
matulni: Thanks!

PR #425, opened 2026-02-05
Make tests in `test_noisy_density_matrix.py` use the BranchSelector feature

### LINKED ISSUE #416: Make the noisy density matrix backend tests deterministic with the `BranchSelector` feature
**Problem statement**

The tests for the noisy density matrix in `test_noisy_density_matrix.py` were written before the introduction of the `BranchSelector` feature. As a consequence since the branch is not fixed and we don't know what the output is, many cases have to be tested (see e.g. [here](https://github.com/TeamGraphix/graphix/blob/53c4fde5d5be29421640b15b357deff412b883ad/tests/test_noisy_density_matrix.py#L313) )

**Fix**

Use the `BranchSelector` feature (introduced in [PR #300](https://github.com/TeamGraphix/graphix/pull/300) to make the tests deterministic.

tibisabau:

- Format added code by `ruff`

Then, please fill in below:

**Context (if applicable):**

The noisy density matrix backend tests in `test_noisy_density_matrix.py` were non-deterministic because they relied on random measurement outcomes. These tests used `or` conditions to check multiple possible outcomes, making them fail intermittently. The BranchSelector feature (introduced in PR #300) allows fixing measurement outcomes deterministically.

Modified all affected tests in `test_noisy_density_matrix.py` to use `BranchSelector` (`ConstBranchSelector` and `FixedBranchSelector`) to fix measurement outcomes, making the tests fully deterministic. Updated test logic based on understanding that noise channels are applied unconditionally after correction commands in the depolarising noise model.

**Related issue:**

Fixes #416

### DIFF
--- a/graphix/flow/_find_gpflow.py
+++ b/graphix/flow/_find_gpflow.py
@@ -258,7 +258,7 @@ class CorrectionMatrix(Generic[_AM_co]):
         correction_function: dict[int, frozenset[int]] = {}
         for node in col_tags:
             i = col_tags.index(node)
-            correction_set = {row_tags[j] for j in np.flatnonzero(self.c_matrix[:, i])}
+            correction_set = {row_tags[int(j)] for j in np.flatnonzero(self.c_matrix[:, i])}
             correction_function[node] = frozenset(correction_set)
         return correction_function
 
--- a/tests/test_density_matrix.py
+++ b/tests/test_density_matrix.py
@@ -302,7 +302,7 @@ class TestDensityMatrix:
         edge = (0, 1)
         dm.cnot(edge)
         psi = psi.reshape((2, 2))
-        psi = np.tensordot(CNOT_TENSOR, psi, ((2, 3), edge))  # type: ignore[assignment]
+        psi = np.tensordot(CNOT_TENSOR, psi, ((2, 3), edge))
         psi = np.moveaxis(psi, (0, 1), edge)
         expected_matrix2 = np.outer(psi, psi.conj())
         assert np.allclose(dm.rho, expected_matrix2)
@@ -320,7 +320,7 @@ class TestDensityMatrix:
         edge = (u, v)
         dm.cnot(edge)
         psi = psi.reshape((2,) * n)
-        psi = np.tensordot(CNOT_TENSOR, psi, ((2, 3), edge))  # type: ignore[assignment]
+        psi = np.tensordot(CNOT_TENSOR, psi, ((2, 3), edge))
         psi = np.moveaxis(psi, (0, 1), edge)
         expected_matrix3 = np.outer(psi, psi.conj())
         assert np.allclose(dm.rho, expected_matrix3)
@@ -354,7 +354,7 @@ class TestDensityMatrix:
         dm.swap(edge)
         rho = dm.rho
         psi = psi.reshape((2, 2))
-        psi = np.tensordot(SWAP_TENSOR, psi, ((2, 3), edge))  # type: ignore[assignment]
+        psi = np.tensordot(SWAP_TENSOR, psi, ((2, 3), edge))
         psi = np.moveaxis(psi, (0, 1), edge)
         expected_matrix2 = np.outer(psi, psi.conj())
         assert np.allclose(rho, expected_matrix2)
@@ -392,7 +392,7 @@ class TestDensityMatrix:
         dm.entangle(edge)
         rho = dm.rho
         psi = psi.reshape((2, 2))
-        psi = np.tensordot(CZ_TENSOR, psi, ((2, 3), edge))  # type: ignore[assignment]
+        psi = np.tensordot(CZ_TENSOR, psi, ((2, 3), edge))
         psi = np.moveaxis(psi, (0, 1), edge)
         expected_matrix2 = np.outer(psi, psi.conj())
         assert np.allclose(rho, expected_matrix2)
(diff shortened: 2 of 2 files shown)


tests/test_noisy_density_matrix.py
```diff
@@ -310,26 +319,36 @@ def test_noisy_measure_channel_rz(self, fx_rng: Generator) -> None:
             ),
         )
 
-    def test_noisy_x_rz(self, fx_rng: Generator) -> None:
+    @pytest.mark.parametrize("outcome_z,outcome_x", [(0, 0), (0, 1), (1, 0), (1, 1)])
```
thierry-martinez: `ruff` recommends
```suggestion
    @pytest.mark.parametrize(("outcome_z", "outcome_x"), [(0, 0), (0, 1), (1, 0), (1, 1)])
```
But since it is a Cartesian product, we could even do
```suggestion
    @pytest.mark.parametrize("z_outcome", [0, 1])
    @pytest.mark.parametrize("x_outcome", [0, 1])
```

tests/test_noisy_density_matrix.py
```diff
@@ -69,14 +70,22 @@ def test_noisy_measure_confuse_hadamard(self, fx_rng: Generator) -> None:
         assert isinstance(res, DensityMatrix)
         assert np.allclose(res.rho, np.array([[0.0, 0.0], [0.0, 1.0]]))
 
-        # arbitrary probability
+    @pytest.mark.parametrize("outcome", [0, 1])
+    def test_noisy_measure_confuse_hadamard_arbitrary(self, fx_rng: Generator, outcome: int) -> None:
```
thierry-martinez: `mypy` complains on line 79 because `ConstBranchSelector` expects an argument of type `Outcome`.

```suggestion
    def test_noisy_measure_confuse_hadamard_arbitrary(self, fx_rng: Generator, outcome: Outcome) -> None:
```

tests/test_noisy_density_matrix.py
```diff
@@ -310,26 +319,36 @@ def test_noisy_measure_channel_rz(self, fx_rng: Generator) -> None:
...
         # x error only
         x_error_pr = fx_rng.random()
-        print(f"x_error_pr = {x_error_pr}")
+        print(f"x_error_pr = {x_error_pr}, outcome_z = {outcome_z}, outcome_x = {outcome_x}")
+
+        # M(0) determines Z, M(1) determines X
+        results = {}
+        cmd_count = 0
+        for cmd in rzpattern:
+            if cmd.kind.name == "M":
```
thierry-martinez: `mypy` does not deduce from this equality that `cmd` is `M`, which is required to access `cmd.node`. The type of `cmd` can be narrowed if we compare `cmd.kind` directly instead.

```suggestion
            if cmd.kind == CommandKind.M:
```

tests/test_noisy_density_matrix.py
```diff
@@ -310,26 +319,36 @@ def test_noisy_measure_channel_rz(self, fx_rng: Generator) -> None:
...
+                if cmd_count == 0:
+                    results[cmd.node] = outcome_z
+                elif cmd_count == 1:
+                    results[cmd.node] = outcome_x
+                cmd_count += 1
```
thierry-martinez: There is no need to loop over all the measurements to modify only the first two of them.

thierry-martinez: There is no need to loop over all the pattern to modify only the two first (and only, in this case) measurements.

```suggestion
        # M(0) determines Z, M(1) determines X
        m_nodes = (cmd.node for cmd in rzpattern if cmd.kind == CommandKind.M)
        results = {next(m_nodes): outcome_z, next(m_nodes): outcome_x}
```

But as your comment says, `rzpattern` is fixed and we know that the two first measurement nodes are 0 and 1, therefore we can just simply refer to them.


```suggestion
        # M(0) determines Z, M(1) determines X
        results = {0: outcome_z, 1: outcome_x}
```

tests/test_noisy_density_matrix.py
```diff
@@ -340,26 +355,36 @@ def test_noisy_x_rz(self, fx_rng: Generator) -> None:
...
+        # M(0) determines Z, M(1) determines X
+        results: dict[int, Outcome] = {}
+        cmd_count = 0
+        for cmd in rzpattern:
+            if cmd.kind.name == "M":
+                if cmd_count == 0:
+                    results[cast("M", cmd).node] = outcome_z
+                elif cmd_count == 1:
+                    results[cast("M", cmd).node] = outcome_x
+                cmd_count += 1
```
thierry-martinez: My previous comment (https://github.com/TeamGraphix/graphix/pull/425#discussion_r2772853098) applies here and lines 410 and below as well: in particular, we don't need to use `cast` here.

tests/test_noisy_density_matrix.py
```diff
@@ -340,26 +355,36 @@ def test_noisy_x_rz(self, fx_rng: Generator) -> None:
             ),
         )
 
-    def test_noisy_z_rz(self, fx_rng: Generator) -> None:
+    @pytest.mark.parametrize(("outcome_z", "outcome_x"), [(0, 0), (0, 1), (1, 0), (1, 1)])
```
thierry-martinez: For homogeneity, I would prefer if we do the same here as for line 324 (and this applies to line 398 and 442 too).

mgarnier59 (approved):
A few extra commits remain for some reason but looks good. Thanks for the good work!

Issue #428, opened 2026-02-09
[Bug]: `ApplyNoise` should be conditional with domain


thierry-martinez:
In #425, @tibisabau noticed that `ApplyNoise` is unconditional, even when applied to $X$ and $Z$ commands that are conditioned by a domain.

We should be able to express that some `ApplyNoise` only apply for certain outcomes by adding a domain to the command.

In particular, that means that the noise model can break the equivalence between the expanded form $MZ^tX^s$ and the form $_t[M]^s$  where the corrections are absorbed in the measurement. Indeed, the noise applied to $Z$ and $X$ will not be applied in the absorbed form (and that corresponds to what would happen on a concrete machine, where the measurement will not apply the gates $Z$ and $X$ but only change the measurement angles).

PR #429, opened 2026-02-10
Fix conditionality of `ApplyNoise` commands

mgarnier59:
As described in Issue #428, the application of the noise after a condition command (`command.X` and `command.Z`) was not conditional.

This PR fixes this issue by introducing an optional field to `ApplyCommand`. This field can take the values `set[Node]` to match the type of `domains` or `None` if the noise in unconditional. The field defaults to `None`. Thus the conditionality mimicks those of correction commands.

Note that this now makes patterns with explicit X and Z corrections before measurements different from those where they have been absorbed in the measurement.

We will have to see if this precludes more advanced usage of the noise model API but it seems fine for now.

We took advantage of this PR to remove code duplication by removing the domain checking of `correct_byproduct` and `apply_noise` from the backends and doing it only at the level of the simulator in a backend-independent way.

We removed a test that somehow made no sense anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+### Fixed
+
+- #429
+  - Modify `graphix.noise_models.noise_model.ApplyNoise` to handle conditionality based on a `domain` attribute (like `command.X` and `command.Z`).
+  - Moved the conditional logic to `graphix.simulator` to remove code duplication in the backends.
+  - Solves [#428](https://github.com/TeamGraphix/graphix/issues/428).
+
+### Changed
+
 ## [0.3.4] - 2026-02-05
 
 ### Added
@@ -44,6 +57,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - #412: Added pretty-print methods (`to_ascii`, `to_latex` and `to_unicode`) for `PauliFlow` and `XZCorrections` classes. Implemented their `__str__` method as a call to `self.to_ascii`.
 
 ### Fixed
+
+-
 
 - #363, #392: `Pattern.remove_input_nodes` is required before the `Pattern.perform_pauli_measurements` method to ensure input nodes are removed and fixed in the |+> state.
 
@@ -216,10 +231,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - #322: Added a new `optimization` module containing:
 
-  * a functional version of `standardize` that returns a standardized
+  - a functional version of `standardize` that returns a standardized
     pattern as a new object;
 
-  * a function `incorporate_pauli_results` that returns an equivalent
+  - a function `incorporate_pauli_results` that returns an equivalent
     pattern in which the `results` are incorporated into measurement
     and correction domains.
     The resulting pattern is suitable for flow analysis. In


graphix/sim/base_backend.py
```diff
@@ -782,25 +781,25 @@ def f_expectation0() -> float:
         return outcome
 
     @override
-    def correct_byproduct(self, cmd: command.X | command.Z, measure_method: MeasureMethod) -> None:
+    def correct_byproduct(self, cmd: command.X | command.Z) -> None:
         """Byproduct correction correct for the X or Z byproduct operators, by applying the X or Z gate."""
-        if np.mod(sum(measure_method.measurement_outcome(j) for j in cmd.domain), 2) == 1:
-            op = Ops.X if cmd.kind == CommandKind.X else Ops.Z
-            self.apply_single(node=cmd.node, op=op)
+        # conditional ligic taken into accoutn in simulator.py
```
thierry-martinez: This comment can be removed if the responsibility of the logic is documented in the docstring. Moreover the docstring does not have to be repeated in the overridden method; only the docstring in the base (abstract) method is necessary.

graphix/sim/base_backend.py
```diff
@@ -782,25 +781,25 @@ def f_expectation0() -> float:
...
+        op = Ops.X if cmd.kind == CommandKind.X else Ops.Z
+        self.apply_single(node=cmd.node, op=op)
 
     @override
-    def apply_noise(self, nodes: Sequence[int], noise: Noise) -> None:
-        """Apply noise.
+    def apply_noise(self, cmd: ApplyNoise) -> None:
+        """Apply noise based on the attributes of `:class: graphix.noise_model.ApplyNoise`.
```
thierry-martinez:
```suggestion
        """Apply noise for the command `:class: graphix.noise_model.ApplyNoise`.
```

graphix/sim/base_backend.py
```diff
@@ -782,25 +781,25 @@ def f_expectation0() -> float:
...
 
         Parameters
         ----------
-        nodes : sequence of ints.
-            Target qubits
-        noise : Noise
-            Noise to apply
+        cmd : ApplyNoise
+            command ApplyNoise
+        measure_method : MeasureMethod
```
thierry-martinez: This parameter has been removed.

graphix/sim/base_backend.py
```diff
@@ -782,25 +781,25 @@ def f_expectation0() -> float:
...
         """
-        indices = [self.node_index.index(i) for i in nodes]
-        self.state.apply_noise(indices, noise)
+        # conditional logic done at the level above
```
thierry-martinez: This comment can be removed too.

graphix/sim/base_backend.py
```diff
@@ -642,7 +641,7 @@ def apply_clifford(self, node: int, clifford: Clifford) -> None:
         """Apply single-qubit Clifford gate, specified by vop index specified in graphix.clifford.CLIFFORD."""
 
     @abstractmethod
-    def correct_byproduct(self, cmd: command.X | command.Z, measure_method: MeasureMethod) -> None:
+    def correct_byproduct(self, cmd: command.X | command.Z) -> None:
         """Byproduct correction correct for the X or Z byproduct operators, by applying the X or Z gate."""
```
thierry-martinez: I would add the following note to document which part of the code is responsible for the domain logic.

```suggestion
        """Byproduct correction correct for the X or Z byproduct operators, by applying the X or Z gate.
        
        Note: the simulator is responsible for checking that the measurement outcomes match the domain condition before calling this method."""
```

graphix/sim/base_backend.py
```diff
@@ -619,7 +618,7 @@ def add_nodes(self, nodes: Sequence[int], data: Data = BasicStates.PLUS) -> None
         Previously existing nodes remain unchanged.
         """
 
-    def apply_noise(self, nodes: Sequence[int], noise: Noise) -> None:  # noqa: ARG002,PLR6301
+    def apply_noise(self, cmd: ApplyNoise) -> None:  # noqa: ARG002,PLR6301
```
thierry-martinez, line 621: I would add the following note in the docstring to document which part of the code is responsible for the domain logic.

```
        Note: the simulator is responsible for checking that the measurement outcomes match the domain condition before calling this method.
```

graphix/sim/tensornet.py
```diff
@@ -768,7 +767,7 @@ def measure(self, node: int, measurement: Measurement, rng: Generator | None = N
         return result
 
     @override
-    def correct_byproduct(self, cmd: command.X | command.Z, measure_method: MeasureMethod) -> None:
+    def correct_byproduct(self, cmd: command.X | command.Z) -> None:
```
thierry-martinez, line 770: The docstring is outdated and can be removed, since only the base method needs to be documented.

graphix/simulator.py
```diff
@@ -311,7 +312,8 @@ def run(self, input_state: Data = BasicStates.PLUS, rng: Generator | None = None
                 self.__measure_method.measure(self.backend, cmd, noise_model=self.noise_model, rng=rng)
             # Use of `==` here for mypy
             elif cmd.kind == CommandKind.X or cmd.kind == CommandKind.Z:  # noqa: PLR1714
-                self.backend.correct_byproduct(cmd, self.__measure_method)
+                if np.mod(sum(self.__measure_method.measurement_outcome(j) for j in cmd.domain), 2) == 1:
```
thierry-martinez:
```suggestion
                if self.__measure_method.check_domain(cmd.domain):
```
by defining the following method in `MeasureMethod`:
```python
def check_domain(self, domain: Iterable[int]) -> bool:
    """Check that the measurement outcomes match the domain condition."""
    return sum(self.measurement_outcome(j) for j in domain) % 2 == 1
```

graphix/simulator.py
```diff
@@ -321,7 +323,11 @@ def run(self, input_state: Data = BasicStates.PLUS, rng: Generator | None = None
                 # handling of ticks during noise transpilation.
                 pass
             elif cmd.kind == CommandKind.ApplyNoise:
-                self.backend.apply_noise(cmd.nodes, cmd.noise)
+                if (
+                    cmd.domain is None
+                    or np.mod(sum(self.__measure_method.measurement_outcome(j) for j in cmd.domain), 2) == 1
+                ):
```
thierry-martinez:
```suggestion
                if cmd.domain is None or self.__measure_method.check_domain(cmd.domain):
```

tests/test_density_matrix.py
```diff
@@ -926,33 +926,3 @@ def test_measure(self, pr_calc: bool) -> None:
...
-        sv_backend.entangle_nodes((1, 2))
-        measure_method.measure(sv_backend, command.M(0))
-        measure_method.measure(sv_backend, command.M(1, angle=-ANGLE_PI / 2, s_domain={0}))
-        sv_backend.correct_byproduct(command.X(2, {1}), measure_method)
-        sv_backend.correct_byproduct(command.Z(2, {0}), measure_method)
-        psi = sv_backend.state.psi
-
-        assert np.allclose(
-            rho, np.outer(psi.astype(np.complex128, copy=False), psi.conj().astype(np.complex128, copy=False))
-        )
```
thierry-martinez, line 958: If we want to keep a unit-test for `correct_byproduct`, I propose the following change, where we ensure to cover every outcome branch.

```python
    @pytest.mark.parametrize("outcome0", [0, 1])
    @pytest.mark.parametrize("outcome1", [0, 1])
    def test_correct_byproduct(self, outcome0: Outcome, outcome1: Outcome) -> None:
        measure_method = DefaultMeasureMethod()
        branch_selector = FixedBranchSelector({0: outcome0, 1: outcome1})
        dm_backend = DensityMatrixBackend(branch_selector=branch_selector)
        dm_backend.add_nodes([0])
        # node 0 initialized in Backend
        dm_backend.add_nodes([1, 2])
        dm_backend.entangle_nodes((0, 1))
        dm_backend.entangle_nodes((1, 2))
        measure_method.measure(dm_backend, command.M(0))
        measure_method.measure(dm_backend, command.M(1, angle=-ANGLE_PI / 2, s_domain={0}))
        if outcome1:
            dm_backend.correct_byproduct(command.X(2, {1}))
        if outcome0:
            dm_backend.correct_byproduct(command.Z(2, {0}))
        rho = dm_backend.state.rho

        sv_backend = StatevectorBackend(branch_selector=branch_selector)
        sv_backend.add_nodes([0])
        # node 0 initialized in Backend
        sv_backend.add_nodes([1, 2])
        sv_backend.entangle_nodes((0, 1))
        sv_backend.entangle_nodes((1, 2))
        measure_method.measure(sv_backend, command.M(0))
        measure_method.measure(sv_backend, command.M(1, angle=-ANGLE_PI / 2, s_domain={0}))
        if outcome1:
            sv_backend.correct_byproduct(command.X(2, {1}))
        if outcome0:
            sv_backend.correct_byproduct(command.Z(2, {0}))
        psi = sv_backend.state.psi

        assert np.allclose(
            rho, np.outer(psi.astype(np.complex128, copy=False), psi.conj().astype(np.complex128, copy=False))
        )
```

mgarnier59: Indeed. But I wanted to separate the introduction of `BranchSelector` for the next PR (updating noisy tests). If we think we need this test, we can reintroduce it then.

graphix/sim/base_backend.py
mgarnier59: Fixed in dd30c2055a786f5bfd9b4c908fc64d791c8ba56c.

graphix/sim/base_backend.py
mgarnier59: Fixed in dd30c2055a786f5bfd9b4c908fc64d791c8ba56c.

graphix/sim/base_backend.py
mgarnier59: Fixed in dd30c2055a786f5bfd9b4c908fc64d791c8ba56c.

graphix/sim/base_backend.py
mgarnier59: Fixed in dd30c2055a786f5bfd9b4c908fc64d791c8ba56c.

graphix/sim/base_backend.py
mgarnier59: Fixed in dd30c2055a786f5bfd9b4c908fc64d791c8ba56c.

graphix/sim/base_backend.py
mgarnier59: Fixed in dd30c2055a786f5bfd9b4c908fc64d791c8ba56c.

graphix/sim/tensornet.py
mgarnier59: I removed it in dd30c2055a786f5bfd9b4c908fc64d791c8ba56c.

graphix/simulator.py
mgarnier59: Done in dd30c2055a786f5bfd9b4c908fc64d791c8ba56c. I have added it as an abstract method for `MeasureMethod` and gave a specific implementation for `DefaultMeasureMethod'. I think this would allow custom logic on measurement outcomes if needed. 

Though it might break what has been done in veriphix where a custom `measure_method` is used. CI will tell anyways.

thierry-martinez: I would prefer to have the MBQC logic implemented in `MeasureMethod`, because it will still be possible for custom measurement methods to override it if needed, but it will be simpler for `veriphix` just to have to override nothing and have the default behavior.

mgarnier59: Fixed in ac681a8b74706de90155f5703c25342c3b3e0dec.

tests/test_density_matrix.py
thierry-martinez: Agreed.

graphix/noise_models/noise_model.py
```diff
@@ -47,6 +47,7 @@ class ApplyNoise(_KindChecker):
     kind: ClassVar[Literal[CommandKind.ApplyNoise]] = dataclasses.field(default=CommandKind.ApplyNoise, init=False)
     noise: Noise
     nodes: list[Node]
+    domain: set[Node] | None = None
```
thierry-martinez, line 66:
```suggestion
    domain: set[Node] | None = None
    """Optional domain for conditional noise.
    
    If `None`, the noise is applied unconditionally.
    Otherwise, the noise is applied if there is an odd number of nodes among `domain` that have been measured with outcome 1 (as for `X` and `Z` commands).
    Note that the noise is never applied if `domain` is the empty set."""
```

mgarnier59: Done in acb924c98d05e3608ec140ff95cbac9b655a19aa.

thierry-martinez:
As you expected, this commit should break `graphix-stim-backend`, but reverse dependency checks don't spot it because the graphix version is pinned in the `graphix-stim-backend` package. I already checked that this would have been spotted by #431. Either we merge this PR (#429) first, and we will have to fix `graphix-stim-backend` for #431, or we merge #431 first and we will have to fix it for this PR.

emlynsg (approved):
LGTM too!

emlynsg:
Also I would add it would be good to update `graphix-stim-backend` to remove the old dependency on `galois`/`numba`.

PR #430, opened 2026-02-11
Enable type-checking for examples

thierry-martinez:
This commit adds type annotations in all example modules, and refines the type of `PatternSimulator` constructor, so that the type parameter for states is automatically inferred from the `backend` argument.

It enables full type-checking with mypy and pyright while preserving existing functionality across the entire code base. No files are ignored by the typers anymore.

**Related issue:**
This commit continues the work started in #302, #308, #312, #347, and #414.

### DIFF
--- a/graphix/simulator.py
+++ b/graphix/simulator.py
@@ -385,31 +385,31 @@ class PatternSimulator(Generic[_StateT_co]):
 
         for cmd in pattern:
             logger.debug("Command: %s", cmd)
-            if cmd.kind == CommandKind.N:
-                self.__prepare_method.prepare(self.backend, cmd, rng=rng)
-            elif cmd.kind == CommandKind.E:
-                self.backend.entangle_nodes(edge=cmd.nodes)
-            elif cmd.kind == CommandKind.M:
-                self.__measure_method.measure(self.backend, cmd, noise_model=self.noise_model, rng=rng)
-            # Use of `==` here for mypy
-            elif cmd.kind == CommandKind.X or cmd.kind == CommandKind.Z:  # noqa: PLR1714
-                if self.__measure_method.check_domain(cmd.domain):
-                    self.backend.correct_byproduct(cmd)
-            elif cmd.kind == CommandKind.C:
-                self.backend.apply_clifford(cmd.node, cmd.clifford)
-            elif cmd.kind == CommandKind.T:
-                # The T command is a flag for one clock cycle in a simulated
-                # experiment, added via a hardware-agnostic
-                # pattern modifier. Noise models can perform special
-                # handling of ticks during noise transpilation.
-                pass
-            elif cmd.kind == CommandKind.ApplyNoise:
-                if cmd.domain is None or self.__measure_method.check_domain(cmd.domain):
-                    self.backend.apply_noise(cmd)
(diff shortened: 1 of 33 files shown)


graphix/simulator.py
```diff
@@ -235,12 +243,64 @@ class PatternSimulator(Generic[_StateT_co]):
...
+        branch_selector: BranchSelector | None = None,
+        graph_prep: str | None = None,
+        symbolic: bool = False,
+    ) -> None: ...
+
+    def __init__(
+        self: PatternSimulator[_StateT | _BuiltinBackendState],
         pattern: Pattern,
-        backend: Backend[_StateT_co] | _BackendLiteral = "statevector",
+        backend: Backend[_StateT] | _BackendLiteral = "statevector",
```
matulni, line 303: For my understanding: since the dataclass attribute of `PatternSimulator` is typed with `_StateT_co` why don't we use `Backend[_StateT_co]` here ?

Do I understand correctly that we use `_StateT_co` because we want, for instance, `PatternSimulator[Statevec]` to be a subtype of `PatternSimulator[DenseState]` (since `Statevec < DenseState`) ?

thierry-martinez: If `_StateT_co` is used here, pyright emits the following diagnostic (a warning by default, which I have configured as an error):
```
error: Type annotation for "self" parameter of "__init__" method cannot contain class-scoped type variables (reportInvalidTypeVarUse)
```
There is no such diagnostic if `_StateT_co` is used in the type of a parameter other than `self`, nor if it is used in the type of `self` for a method other than `__init__`.

I am not sure I fully understand the rationale for this restriction. By definition, `self` in all instance methods has type `PatternSimulator[_StateT_co]`, where `_StateT_co` is universally quantified at the class level. If we wished to refine the type of `self`, it would be clearer to introduce a new type variable quantified at the function level. However, this does not explain why using `_StateT_co` directly is permitted for other methods and why `__init__` is treated specially.

Alternatively, one might argue that at construction time `_StateT_co` is not yet bound, and therefore `__init__` would need to use locally quantified type variables. But that does not explain why only the annotation of `self` is rejected, while `_StateT_co` is permitted in other parameter.

thierry-martinez: Anyway, even if it is somewhat obscure why `_StateT_co` is forbidden here while it is permitted elsewhere, I think that using another type variable here is clearer.

First, the variance of a type variable is meaningful only at the level of the class, not at the level of a function or method. As you mentioned it, saying that `PatternSimulator[_StateT_co]` is covariant in `_StateT_co` means that `A` being a subtype of `B` implies `PatternSimulator[A]` is a subtype of `PatternSimulator[B]`.

At the method level, `self` is of type `PatternSimulator[_StateT_co]` by definition, and using`_StateT_co` for another parameter simply refers to the same type parameter as that of `self`. Here `self: PatternSimulator[_StateT | _BuiltinBackendState]` means that the definition of `__init__` is valid for a type parameter `_StateT_co` that is larger than `_BuiltinBackendState`, i.e. `_BuiltinBackendState` union another type `_StateT`, where `_StateT` is local to the method definition.

matulni: Hi Thierry, thanks a lot for the detailed explanation. I think it's clear now :)

PR #431, opened 2026-02-11
Force test reverse dependencies with the current graphix codebase

thierry-martinez:
This commit installs graphix after the reverse dependency to ensure that the test uses the current graphix codebase even if another version is pinned in the reverse dependency.


emlynsg (approved):
CI logs show graphix install occurs last, so LGTM!

PR #434, opened 2026-02-15
Fix #324: use `match` for better readability and typing

### LINKED ISSUE #324: Disable `PLR1714` while it is incompatible with `mypy`
`PLR1714` suggests rewriting code like:
```python
if cmd.kind == CommandKind.X or cmd.kind == CommandKind.Z: ...
```
into
```python
if cmd.kind in {CommandKind.X, CommandKind.Z}: ...
```

While it can be considered to legitimately improve readability, `mypy` currently does not understand that `cmd` is an instance of `X` or `Z` in the `if`-block, whereas `mypy` is able to understand it in the original code with `or`.

@EarlMilktea [suggests](https://github.com/TeamGraphix/graphix/pull/322#discussion_r2222845464) that we disable PLR1714 while it is incompatible with `mypy`.

A caveat is that we won't benefit from such suggestions in case where we don't need `mypy`'s type narrowing, but I guess most of the time we are in a context where we need such narrowing, so it makes sense to disable the rule entirely.

thierry-martinez:
This commit replaces `if`/`elif`/`else` chains with `match` when the conditions perform equality comparisons on the same subject. These changes improve readibility and avoid `mypy` shortcomings related to type narrowing.

Before this commit, PLR1714 suggested rewriting code such as:

```python
if cmd.kind == CommandKind.X or cmd.kind == CommandKind.Z: ...
```

into:

```python
if cmd.kind in {CommandKind.X, CommandKind.Z}: ...
```

However, `mypy` does not currently infer that `cmd` is an instance of `X` or `Z` within the `if` block when set membership is used.

With this commit, the condition is expressed using an or-pattern, and type checking behaves as expected:

```python
match cmd.kind:
    case CommandKind.X | CommandKind.Z: ...
```

### DIFF
--- a/.github/workflows/typecheck.yml
+++ b/.github/workflows/typecheck.yml
@@ -32,7 +32,7 @@ jobs:
 
       - run: |
           python -m pip install --upgrade pip
-          pip install -e .[dev,extra]
+          pip install -e .[dev,extra,typing]
           pip install --no-deps -e .[no-deps]
 
       - run: mypy
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,6 +39,7 @@ dependencies = { file = ["requirements.txt"] }
 dev = { file = ["requirements-dev.txt"] }
 extra = { file = ["requirements-extra.txt"] }
 doc = { file = ["requirements-doc.txt"] }
+typing = { file = ["requirements-typing.txt"] }
 
 [tool.setuptools.packages.find]
 include = ["graphix", "stubs"]
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -4,10 +4,6 @@ pre-commit # for language-agnostic hooks
 pyright
 ruff==0.15.0
 
-# numpy and numba versions are sensible
-numpy==2.4.2
-numba==0.64.0
-
 # Stubs
 types-networkx==3.6.1.20251220
 types-psutil
--- a/requirements-typing.txt
+++ b/requirements-typing.txt
@@ -1,0 +1,13 @@
+# Pinning for typing
+
+# Packages listed in this file should also appear in
+# `requirements.txt`, `requirements-dev.txt`, or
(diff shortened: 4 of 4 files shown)


graphix/transpiler.py
emlynsg: Minor but slight readability improvement:

```
match instr.axis:
    case Axis.X:
        circuit.h(instr.target)
        circuit.m(instr.target, Axis.Z)
    case Axis.Y:
        circuit.rx(instr.target, ANGLE_PI / 2)
        circuit.m(instr.target, Axis.Z)
    case _:
        circuit.add(instr)
```
Could possibly do the same to the outer if else.

graphix/qasm3_exporter.py
```diff
@@ -184,61 +177,57 @@ def command_to_qasm3_lines(cmd: Command) -> Iterator[str]:
...
-        yield from domain_to_qasm3_lines(cmd.domain, f"z q{cmd.node}")
-
-    elif cmd.kind == CommandKind.C:
-        for op in cmd.clifford.qasm3:
-            yield str(op) + " q" + str(cmd.node) + ";\n"
-
-    else:
-        raise ValueError(f"invalid command {cmd}")
+                yield f"h q{cmd.node};\n"
+            if cmd.angle != 0:
```
emlynsg: Similar here:
```
match cmd.plane:
    case Plane.XY:
        gate = "rx"
        angle = -cmd.angle
    case Plane.XZ:
        gate = "ry"
        angle = -cmd.angle
    case Plane.YZ:
        gate = "rx"
        angle = cmd.angle
    case _:
        assert_never(cmd.plane)
```

graphix/optimization.py
emlynsg, line 275: Can change to match here

graphix/pattern.py
emlynsg: Can replace these with match too

emlynsg: Replace with match.

Issue #387, opened 2025-12-08
Improve method `Pattern.draw_graph`


matulni:
_This issue is part of **unitaryDESIGN 2026**. You have to be [registered](https://unitary.design/register/) to complete this issue._
_Learn more about the PR review process [here](https://unitary.design/participant-guide/) and about unitaryDESIGN rules and opportunities [here](https://unitary.design/rules/)!_ 

_A note about [AI Slop](https://en.wikipedia.org/wiki/AI_slop): while we are open to collaboration with LLMs for unitaryDESIGN, fully AI-generated PRs are not acceptable. It is at the maintainers' discretion whether or not LLM-generated PRs are the right fit for the issues, and those that appear fully AI-generated may be immediately rejected._

This issue consists of two related but independent tasks.

**Context**

Graphix offers the possibility to graphically represent an MBQC pattern by calling `:func: graphix.Pattern.draw_graph`. This method produces a plot of the underlying open graph resource state of the pattern and its flow (when it exists). Example:
```python
>>> p = Pattern(input_nodes=[1], cmds=[N(0), N(2), N(3), N(4), N(5), N(6), N(7), E((0, 1)), E((0, 3)), E((1, 4)), E((3, 4)), E((3, 2)), E((3, 6)), E((4, 7)), E((2, 5)), M(0, Plane.XZ, 0.1), Z(2, {0}), Z(6, {0}), Z(7, {0}), X(4, {0}), X(7, {0}), M(1, angle=0.1), Z(5, {1}), Z(7, {1}), X(2, {1}), X(4, {1}), M(3), X(6, {3}), M(4, angle=0.1), X(7, {4}), M(5, Plane.YZ, 0.1), Z(2, {5})])
>>> p.draw_graph(flow_from_pattern=False, show_measurement_planes=True)
```
Output:

<img width="206" height="405" alt="Image" src="https://github.com/user-attachments/assets/01c8766d-25cb-4236-828b-76b94bc7cb19" />

The output plot contains the following information:

- Nodes and edges of the underlying open graph resource state. Nodes (circles) represent qubits and edges (thin dashed lines) represent the application of an entanglement command ($CZ$ gate) between the linked nodes.
- Qubit number (as node labels).
- Input qubits (red-border nodes). In the example, node 1.
- Output qubits (grey nodes). In the example, nodes {2, 6, 7}.
[...]

PR #444, opened 2026-02-17
Improved Pattern.draw_graph visualization

CodeMaverick2:

- Format added code by `ruff`

Then, please fill in below:

**Context (if applicable):**
 - The current draw_graph uses color-only node differentiation, dashed lines for both edges and layer separators (causing visual confusion), and has no angle display or legend.                                                                                                                                     

  - Node shapes follow MBQC literature conventions: squares for inputs, filled circles for measured, empty circles for outputs
  - Replaced vertical dashed layer separators with gray labels and a "Measurement order" arrow
  - Graph edges are light dashed lines; correction arrows remain solid — clear visual hierarchy
  - Added show_measurements parameter to display measurement labels (plane + angle in units of π)
  - Added show_legend parameter for node/edge type legend
  - Added _format_measurement_label static method with unit tests
  - Regenerated baseline images
  
###  From issue - 
523657674-01c8766d-25cb-4236-828b-76b94bc7cb19

### Generated with PR changes - 
after_issue_legend

The first image (before) is the example from the issue; the second (after) is the same pattern rendered with this PR.
Inputs (node 1) are now squares instead of red circles, measured nodes are filled black, Pauli-measured nodes (node 3) are blue, and outputs (2, 6, 7) are empty circles. 
Dashed layer separators are replaced by a horizontal arrow with layer labels. 
Measurement labels show plane and angle (e.g. XZ(π/10)). 
Graph edges (light dashed) are visually distinct from correction arrows (solid). A legend explains all visual elements.

**Related issue:**
#387

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - #440, #441: `Statevec.nqubits` now returns the correct value.
 
+- #455: Causal-flow finding algorithm (`graphix.flow._find_cflow.py`) does not raise `RecursionError` now.
+
 ### Changed
 
 - #181, #423: Structural separation of Pauli measurements
--- a/graphix/flow/_find_cflow.py
+++ b/graphix/flow/_find_cflow.py
@@ -16,8 +16,6 @@ from graphix.flow.core import CausalFlow
 from graphix.fundamentals import Plane
 
 if TYPE_CHECKING:
-    from collections.abc import Set as AbstractSet
-
     from graphix.opengraph import OpenGraph, _PM_co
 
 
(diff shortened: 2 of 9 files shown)


thierry-martinez:
Veriphix CI is now fixed. It's all green now!

graphix/visualization.py
```diff
@@ -414,39 +486,203 @@ def visualize_graph(
...
+        str
+            Formatted label string, or empty string if nothing to show.
+        """
+        if isinstance(meas, PauliMeasurement):
+            if show_planes:
+                sign = "" if meas.sign == Sign.PLUS else "-"
+                return f"{sign}{meas.axis.name}"
+            if show_angles:
+                bloch = meas.to_bloch()
+                if isinstance(bloch.angle, (int, float)):
```
thierry-martinez: This condition should always be `True`. I think it should be replaced by an `assert`.

graphix/visualization.py
```diff
@@ -414,39 +486,203 @@ def visualize_graph(
...
+        Returns
+        -------
+        str
+            Formatted label string, or empty string if nothing to show.
+        """
+        if isinstance(meas, PauliMeasurement):
+            if show_planes:
+                sign = "" if meas.sign == Sign.PLUS else "-"
+                return f"{sign}{meas.axis.name}"
+            if show_angles:
```
thierry-martinez: I believe we should rethink if `show_planes` and `show_angles` are the flags we want: for instance, for a Pauli measurement, showing only the angles basically reveals the sign but not the axis, and I don't know in which context it could be something we want. As a proposal, I would suggest to keep only one flag `show_measurements` and use `str(meas)` for Pauli measurements.

graphix/visualization.py
```diff
@@ -414,39 +486,203 @@ def visualize_graph(
...
+                else:
+                    angle_str = str(meas.angle)
+                if plane_str:
+                    return f"{plane_str}({angle_str})"
+                return angle_str
+            return plane_str
+        # Generic Measurement fallback
+        if show_planes:
+            return meas.to_plane_or_axis().name
+        return ""
```
thierry-martinez: I would prefer `None` if there is nothing to show.

graphix/visualization.py
```diff
@@ -252,35 +269,83 @@ def _shorten_path(path: Sequence[_Point]) -> list[_Point]:
...
-            color = "black"  # default color for 'other' nodes
-            inner_color = "white"
-            if node in self.og.input_nodes:
-                color = "red"
-            if node in self.og.output_nodes:
-                inner_color = "lightgray"
-            elif show_pauli_measurement and isinstance(self.og.measurements[node], PauliMeasurement):
-                inner_color = "lightblue"
+            is_input = node in input_set
+            is_output = node in output_set
```
thierry-martinez: These conditions seem to be short and readable enough to be inlined.

graphix/visualization.py
```diff
@@ -252,35 +269,83 @@ def _shorten_path(path: Sequence[_Point]) -> list[_Point]:
...
+            is_pauli = (
+                show_pauli_measurement
+                and node in self.og.measurements
+                and isinstance(self.og.measurements[node], PauliMeasurement)
+            )
```
thierry-martinez: `is_pauli` is independent of `show_pauli_measurement`. I think it would be clearer if the logic to decide how to color a node based on whether it is Pauli or not was inside an `if show_pauli_measurement:` statement.

graphix/visualization.py
```diff
@@ -252,35 +269,83 @@ def _shorten_path(path: Sequence[_Point]) -> list[_Point]:
...
-    def _draw_labels(self, pos: Mapping[int, _Point]) -> None:
+    def _draw_labels(self, pos: Mapping[int, _Point], dark_nodes: set[int] | None = None) -> None:
+        """Draw node number labels with appropriate text color.
+
+        Parameters
+        ----------
+        pos : Mapping[int, tuple[float, float]]
+            Dictionary of node positions.
+        dark_nodes : set[int] | None
+            Nodes with dark backgrounds that need white text. If None, all labels are black.
```
thierry-martinez: It looks very ad hoc, complicates the logic inside this function, and makes the color logic spread across the module. I would prefer to expose the `font_color` parameter of `draw_networkx_labels`, which can be either a color or a mapping from nodes to colors, and compute such a dictionary in `__draw_nodes_role`.

CodeMaverick2:
@emlynsg Thanks for the feedback. This time I explored multiple visual variants systematically testing different combinations of label positions, layer indicators, and edge styles side-by-side against the original to find the best approach.

  **What changed in this update:**

  1. **Dotted vertical layer separators** - Instead of removing separators entirely (which lost layer clarity) or keeping dashed ones (which conflicted with  dashed edges), I switched to **dotted lines** - clearly distinct from dashed edges while restoring the layer structure visibility you mentioned was better in   the original.

  2. **Measurement labels moved to lower-right** - Moved from above-node placement (which cluttered the graph) to lower-right, matching the original's label      
  position. Adaptive placement tries lower-right first, falls back to lower-left or above only when a neighbor would overlap.

  3. **Restored edge visibility** - Edges changed back from gray to black dashed, matching the original's visibility.

  4. **Subtler label styling** - Removed bold, thinner white stroke, keeping labels unobtrusive like the original.

  5. **Angle label truncation** - Non-rational angles now display compactly (e.g. `XY(1.72π)` instead of long decimals).

  6. **Code review fixes** (from @thierry-martinez's inline comments) - Replaced redundant None check with assert, inlined `is_pauli` into the condition.

  **Side-by-side comparison:**

  Small graph - Original:

small_original

  Small graph - Updated:

final_small


  Large graph - Original:

large_original

  Large graph -Updated:

final_large


  **On arrowhead placement:** The arrow routing code is identical to master arrows behave exactly the same as the original. Modifying the pathfinding algorithm 
  would be a separate concern outside the scope of issue #387.

CodeMaverick2:
@emlynsg Small follow-up - noticed nodes 5-6 had overlapping measurement labels in the small graph from my previous comment. Fixed this by making the adaptive    label placement also track already-placed labels as obstacles, so adjacent nodes no longer merge their labels together

Updated image:
                                                                                                                                                                  
  **Small graph - Updated (label overlap fixed):**
  
tw_v1_small

**Large graph - Updated:**
final_large

tests/baseline/test_draw_graph_reference_True.png
matulni, line 1: Could you investigate why the plot is so compressed ?  Labels overlap nodes and it's a bit difficult to read.
Additionally, removing the inner ticks in the generated plots would help lighten the plots.

graphix/visualization.py
```diff
@@ -252,35 +269,82 @@ def _shorten_path(path: Sequence[_Point]) -> list[_Point]:
...
         if max(self.og.graph.nodes(), default=0) >= 100:
             fontsize = int(fontsize * 2 / len(str(max(self.og.graph.nodes()))))
-        nx.draw_networkx_labels(self.og.graph, pos, font_size=fontsize)
-
-    def __draw_nodes_role(self, pos: Mapping[int, _Point], show_pauli_measurement: bool = False) -> None:
-        """
-        Draw the nodes with different colors based on their role (input, output, or other).
+        for node in self.og.graph.nodes():
+            x, y = pos[node]
+            color = font_color[node] if isinstance(font_color, dict) else font_color
```
matulni: It's better to use  [`.get` method](https://docs.python.org/3/library/stdtypes.html#dict.get) with default value

graphix/visualization.py
```diff
@@ -414,39 +490,217 @@ def visualize_graph(
         y_min = min((pos[node][1] for node in self.og.graph.nodes()), default=0)  # Get the minimum y coordinate
         y_max = max((pos[node][1] for node in self.og.graph.nodes()), default=0)  # Get the maximum y coordinate
 
-        if l_k is not None and l_k:
-            # Draw the vertical lines to separate different layers
-            for layer in range(min(l_k.values()), max(l_k.values())):
+        has_layers = l_k is not None and len(l_k) > 0
+        show_layers = show_measurement_order and has_layers
+        if show_layers:
+            assert l_k is not None
```
matulni: I think this assertion is not necessary, already checked by the conditional on `show_layers`

CodeMaverick2:
@matulni Thanks for the detailed feedback! Here are the changes addressing each point:

  **Compressed reference plot + inner ticks** - Increased the figsize width multiplier from 0.7 to 1.0 so the plot renders at a readable size with
  `node_distance=(0.7, 0.6)`. Added `plt.gca().set_axis_off()` to remove matplotlib's tick marks and frame.

  **`.get()` method** - Changed `font_color[node]` to `font_color.get(node, "black")` in `_draw_labels`.

  **Redundant assert** - Removed `assert l_k is not None` inside the `show_layers` block — the condition is already guaranteed by `has_layers`.

  **Labels outside the plot box** - The adaptive placement algorithm now computes the graph's x-boundary and excludes left-side placement for
  leftmost-column nodes and right-side placement for rightmost-column nodes. Also expanded the left/right and top margins when `show_measurements=True`.

  **Labels overlapping edges + fraction format** - Added a semi-transparent white `bbox` behind each measurement label so text stays readable over arrows.
  Reverted the fraction format from stacked LaTeX `\frac{π}{n}` back to inline Unicode `π/n` as requested.

  Updated images using your exact example graphs:


fix1_boundary


fix2_whitebox

matulni:
Subsumed by #479

PR #445, opened 2026-02-18
Circuit extraction

matulni:
This PR introduces functionality to perform the circuit extraction algorithm presented in Ref. [1] which allows to extract a circuit without ancillas from a strongly deterministic pattern.

## Context

The main result of the paper is a procedure to perform the transformation

$$U_{\mathcal{P}} \rightarrow C V(\mathbf{\theta}),$$

where $U_{\mathcal{P}}$ is the isometry implemented by the pattern $\mathcal{P}$ (a unitary when the pattern has the same number of inputs and outputs), $C$ is a Clifford map, and $V(\mathbf{\theta})$ is an ordered product of Pauli exponentials. 

The structure of the extraction process reads as follows:

```mermaid
flowchart LR

n0["Pattern"]
n1["OpenGraph"]
n2("Focused PauliFlow")

subgraph s1["ExtractionResult"]
    s10("PauliExponentialDAG")
    s11("CliffordMap")
end

%% Junction node to merge arrows
j1(( ))

n3("Graphix circuit")

n0 --> n1
n1 --> n2
n2 --> s10
n2 --> s11

%% Merging into the junction
s10 --> j1
s11 --> j1
j1 --> n3

%% Styling the junction to be small
style j1 fill:#000,stroke-width:0px,width:10px
``` 

- The transformation `Pattern` -> `OpenGraph` -> `Focused PauliFlow` already exists in the current implementation of Graphix. The $O(N^3)$ Pauli-flow extraction algorithm in Ref. [2] always returns focused Pauli flows, contrary to the $O(N^5)$ version in Ref. [1] which requires an additional post-processing to focus the correction function.

- `ExtractionResult` is the output of the algorithm which is denoted "PdDAG" in [1]. In particular:
    - `PauliExponentialDAG` is a mapping between measured nodes in the open graph and Pauli exponentials, and a partial order between the measured node (the flow's partial order). Pauli exponentials are a pair of an `Angle` (the node's measurement angle) and a `PauliString` acting on the output nodes of the open graph. The corresponding Pauli string is obtained from the focused flow's correction function. 
    - `Clifford Map` describes a linear transformation between the space of input qubits and the space of output qubits. It is encoded as a map from the Pauli-group generators ($X$ and $Z$) over the input nodes to Pauli strings over the output nodes.

        The $Z$-map is also obtained from the focused flow's correction function. The $X$-map is obtained from the focused flow's correction function of the _extended_ open graph (see Def. C.3 in [1]).

## Summary of changes

- Added the following to `graphix.flow.core.PauliFlow`: 

    - `is_focused` : (method) Verifies if a Pauli flow is focused according to definition 4.3 in Ref. [1].
    - `extract_circuit` : (method) Returns a `ExtractionResult` object.
    - `pauli_strings` : (cached_property) Associates a Pauli string to every corrected node in the flow's correction function.

- Added new module `graphix.circ_ext.extraction`
    - This module contains the machinery to perfom the transformation `Focused PauliFlow` -> `ExtractionResult`. 
    - We introduce the following (self-descriptive) classes:
        - `ExtractionResult`
        - `PauliString`
        - `PauliExponential`
        - `PauliExponentialDAG`
        - `CliffordMap`

- Added new module `graphix.circ_ext.compilation`
    - This module contains the machinery to perform the transformation `ExtractionResult` -> `Graphix circuit`.
    - It lays out a structure of abstract compilation passes which will allow in the future to experiment with different strategies. Currently:
        - The transformation `PauliExponentialDAG` -> `Graphix circuit` is done with the naive "star" construction (see [3]).
        - The transformation `CliffordMap` -> `Graphix circuit` relies on `stim` and currently only supports unitaries (i.e., patterns with the same number of inputa and outputs). To avoid introducing a dependency on `stim`, the `StimCliffordPass` is implemented in the external plugin [`graphix-stim-compiler`](https://github.com/qat-inria/graphix-stim-compiler). This compilation pass together with the tooling introduced in this PR allow to do a round-trip conversion circuit -> MBQC pattern -> circuit.

We note that all the transformations in the figure work with parametric objects as well.

**References**

[1] Simmons, 2021, arXiv:2109.05654.

[2] Mitosek and Backens, 2024, arXiv:2410.23439.

[3] https://quantumcomputing.stackexchange.com/questions/5567/circuit-construction-for-hamiltonian-simulation/11373#11373

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -126,7 +126,7 @@ class ReverseDependency:
             branch="python_314",
             doctest_modules=False,
         ),
-        ReverseDependency("https://github.com/qat-inria/graphix-stim-compiler", branch="branch_review"),
+        ReverseDependency("https://github.com/qat-inria/graphix-stim-compiler"),
     ],
 )
 def tests_reverse_dependencies(session: Session, package: ReverseDependency) -> None:


graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,420 @@
...
+    z_nodes : AbstractSet[int]
+        Nodes on which a Pauli Z operator is applied.
+    negative_sign : bool
+        Boolean flag indicating a -1 phase in the Pauli string if ``True``.
+    """
+
+    x_nodes: AbstractSet[int] = dataclasses.field(default_factory=frozenset)
+    y_nodes: AbstractSet[int] = dataclasses.field(default_factory=frozenset)
+    z_nodes: AbstractSet[int] = dataclasses.field(default_factory=frozenset)
+    negative_sign: bool = dataclasses.field(default_factory=lambda: False)
```
thierry-martinez: We could replace this field with `sign: Sign`, with `Sign` from `graphix.fundamentals`.

graphix/circ_ext/compilation.py
```diff
@@ -0,0 +1,288 @@
...
+        -----
+        It is assumed that the ``x``, ``y``, and ``z`` node sets of the Pauli string in the exponential are well-formed, i.e., contain only output nodes and are pairwise disjoint.
+        """
+        if pexp.angle == 0:  # No rotation
+            return
+
+        nodes = sorted(
+            pexp.pauli_string.x_nodes | pexp.pauli_string.y_nodes | pexp.pauli_string.z_nodes,
+            key=outputs_mapping.index,
+        )
```
thierry-martinez: I think the following computation is equivalent but a bit simpler, conceptually and algorithmically (filtering rather than sorting). What do you think?

```suggestion
        nodes = [
            node
            for node in outputs_mapping
            if node in pexp.pauli_string.x_nodes | pexp.pauli_string.y_nodes | pexp.pauli_string.z_nodes
        ]
```

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,420 @@
...
+
+    Attributes
+    ----------
+    angle : Angle | Expression
+        The Pauli exponential angle :math:`\alpha` in units of :math:`\pi`. When extracted from a corrected node, it corresponds to the node's measurement divided by two.
+    pauli_string : PauliString
+        The signed Pauli string :math:`P` specifying the tensor product of Pauli operators acting on the corresponding MBQC nodes.
+    """
+
+    angle: Angle | Expression
```
thierry-martinez: Couldn't we use `ParameterizedAngle` here?

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,420 @@
...
+        See Eq. (13) and Lemma 4.4 in Ref. [1]. The phase of the Pauli string is given by Eq. (37).
+
+        References
+        ----------
+        [1] Simmons, 2021 (arXiv:2109.05654).
+        """
+        pauli_string = flow.pauli_strings[node]
+        meas = flow.og.measurements[node]
+        # We don't extract any rotation from Pauli Measurements. This is equivalent to setting the angle to 0.
+        angle = 0 if isinstance(meas, PauliMeasurement) else meas.downcast_bloch().angle / 2
```
thierry-martinez: If you check for `BlochMeasurement` instead, you don't have to `downcast`:

```suggestion
        angle = meas.angle / 2 if isinstance(meas, BlochMeasurement) else 0
```

graphix/circ_ext/compilation.py
```diff
@@ -0,0 +1,288 @@
...
+        circuit = self.cm_cp.add_to_circuit(er.clifford_map)
+        return self.pexp_cp.add_to_circuit(er.pexp_dag, circuit)
+
+
+class PauliExponentialDAGCompilationPass(ABC):
+    """Abstract base class to implement a compilation procedure for a Pauli Exponential DAG."""
+
+    @staticmethod
+    @abstractmethod
+    def add_to_circuit(pexp_dag: PauliExponentialDAG, circuit: Circuit | None = None, copy: bool = False) -> Circuit:
```
thierry-martinez: I think that since the compilation APIs `PauliExponentialDAGCompilationPass` and `CliffordMapCompilationPass` are meant to be implemented by users and called from graphix internals, it is better to make them simpler to implement rather than simpler to use. In this respect, I would not give the caller the freedom to pass a circuit or not and to choose between in-place modification and functional update. I think the simplest approach is to decide that we always expect a circuit to be passed with the correct size, and we only update it in place, and we adapt `er_to_circuit` to follow this convention.

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,420 @@
...
+        # One phase flip per edge between adjacent vertices in the correction set.
+        for edge in combinations(c_set, 2):
+            negative_sign ^= edge in og.graph.edges()
+
+        # One phase flip per two Ys in the graph state stabilizer.
+        negative_sign ^= bool(len(inter_c_odd_set) // 2 % 2)
+
+        # One phase flip per node in the graph state stabilizer that is absorbed from a Pauli measurement with angle π.
+        for n, meas in og.measurements.items():
+            if isinstance(meas, PauliMeasurement) and n in (c_set | odd_c_set):
```
thierry-martinez: If `c_set | odd_c_set` is expected to be small compared to `og.measurements` (which I think it is likely to be the case for large open graphs), I think it is better to iterate over the neighborhood (in terms of corrections) instead of iterating over all the measurements.

```suggestion
        for n in c_set | odd_c_set:
            meas = og.measurements[n]:
            if isinstance(meas, PauliMeasurement):
```

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,420 @@
...
+        x_corrections = frozenset((c_set - odd_c_set).intersection(og.output_nodes))
+        y_corrections = frozenset(inter_c_odd_set.intersection(og.output_nodes))
+        z_corrections = frozenset((odd_c_set - c_set).intersection(og.output_nodes))
+
+        # Sign computation.
+        negative_sign = False
+
+        # One phase flip per edge between adjacent vertices in the correction set.
+        for edge in combinations(c_set, 2):
+            negative_sign ^= edge in og.graph.edges()
```
thierry-martinez: The following code is a bit simpler and faster if the graph is sparse (but slower if the graph is dense; however, it would be much faster in any case if we used `rustworkx`!):

```suggestion
        negative_sign ^= graph.subgraph(c_set).number_of_edges() % 2 == 1
```

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,420 @@
...
+    z_map: dict[int, PauliString] = {}
+    iset = set(flow.og.input_nodes)
+
+    for node in iset.intersection(flow.og.measurements.keys()):
+        z_map[node] = flow.pauli_strings[node]
+
+    for node in iset.intersection(flow.og.output_nodes):
+        z_map[node] = PauliString(z_nodes=frozenset({node}))
+
+    return z_map
```
thierry-martinez: All nodes are either measured or output nodes, so I think we can implement this somewhat more simply.

```suggestion
    return {
        node: flow.pauli_strings[node] if node in flow.og.measurements else PauliString(z_nodes=frozenset({node}))
        for node in flow.og.input_nodes
    }
```

graphix/circ_ext/compilation.py
```diff
@@ -0,0 +1,288 @@
...
+
+    @staticmethod
+    def add_to_circuit(pexp_dag: PauliExponentialDAG, circuit: Circuit | None = None, copy: bool = False) -> Circuit:
+        """Add a Pauli exponential DAG to a circuit.
+
+        See documentation in :meth:`PauliExponentialDAGCompilationPass.add_to_circuit` for additional information.
+        """
+        circuit = initialize_circuit(pexp_dag.output_nodes, circuit, copy)  # May raise value error
+        outputs_mapping = NodeIndex()
+        outputs_mapping.extend(pexp_dag.output_nodes)
```
thierry-martinez: Is there a particular reason for the Pauli strings of `pexp_dag` to use the pattern numbering instead of the `output_nodes` numbering? I would prefer Pauli strings to use the `output_nodes` numbering directly, so that `add_pexp` can use these indices directly instead of having to take care of `outputs_mapping` (either by changing the way the Pauli strings are computed, or by remapping them once for all before calling `add_to_circuit`).

graphix/circ_ext/extraction.py
matulni: Yes, I forgot about that `TypeAlias`. Done in [f99c285](https://github.com/TeamGraphix/graphix/pull/445/commits/f99c285354cacdc714c36d7f8b77da2274b86331)

graphix/circ_ext/extraction.py
matulni: Thanks, done in [f99c285](https://github.com/TeamGraphix/graphix/pull/445/commits/f99c285354cacdc714c36d7f8b77da2274b86331)

graphix/circ_ext/extraction.py
matulni: Indeed. Done in [f99c285](https://github.com/TeamGraphix/graphix/pull/445/commits/f99c285354cacdc714c36d7f8b77da2274b86331). (Small comment, elements in `c_set | odd_c_set` may be outputs, so we need `og.measurements.get(n, None)`.

graphix/circ_ext/extraction.py
matulni: Nice, thanks. Done in [f99c285](https://github.com/TeamGraphix/graphix/pull/445/commits/f99c285354cacdc714c36d7f8b77da2274b86331). Do you suggest moving to `rustworks` (adding this to the backlog)?

graphix/circ_ext/extraction.py
matulni: Indeed, done in [f99c285](https://github.com/TeamGraphix/graphix/pull/445/commits/f99c285354cacdc714c36d7f8b77da2274b86331)

graphix/circ_ext/extraction.py
matulni: Much nicer, done in [8dd3ebc](https://github.com/TeamGraphix/graphix/pull/445/commits/8dd3ebcf402b58a45373ccec70a8aae8007e83bd)

graphix/circ_ext/compilation.py
matulni: I'm not sure if I understand this comment.
Pauli strings are defined on output nodes (they are represented as a collection of sets of output nodes + a sign). The mapping `outputs_mapping` is required because in the current implementation Pauli exponentials (which are just a Pauli string + an angle) don't have information on the order of the output nodes of the original pattern. This information is stored in the parent object, namely, the PauliExponentialDAG. When we compile a Pauli exponential via the ladder pass, we need to iterate over the nodes of the Pauli string in the order defined by the output nodes list.

graphix/circ_ext/compilation.py
matulni: You are right! Thanks, done in [95fd393](https://github.com/TeamGraphix/graphix/pull/445/commits/95fd39339538785f36dc0fa84da6771b58e2d52e)

graphix/circ_ext/compilation.py
matulni: Agree, implemented in [575a9ac](https://github.com/TeamGraphix/graphix/pull/445/commits/575a9ac78aa8b4eb6863d199d033b1161bd249d2).
Since we modify the circuit in place I chose to return `None` for `add_to_circuit` to follow Python convention (we can't chain this methods then, but as you say they are be meant to be internals so I think this is clearer).

graphix/circ_ext/compilation.py
thierry-martinez: I was thinking about something like this: https://github.com/thierry-martinez/graphix/commit/d6a14641b02ed0265eb571892eb66205286567e4
The idea is that the mapping of indices can be performed before the compilation so that it separates concerns: the ladder pass (or any other pass implementation) does not have to reimplement the mapping. I'm still not sure whether it would make sense to integrate `remap` into `from_measured_node`.

thierry-martinez: Sorry, I forgot to update some tests: https://github.com/thierry-martinez/graphix/commit/e971fdedccdbf77b31cdc204fb9c21761b0915c7

matulni: Hi, thanks for the comment. I pushed some mods to your suggestion in 5903cf8.

I agree that it makes sense to manage the node-qubit mapping internally so that potential users don't have to take care about that when writing their compilation passes. However, I think that the object `PauliExponentialDAG` is a "pure MBQC" construct. It's derived from a pattern, and the nodes remap only makes sense when (and if) we want to synthesise a circuit. Therefore, I think it's best that `PauliExponentialDAG.from_focused_flow` returns a `PauliExponentialDAG` defined on output nodes, and that the remapping is done at the level of `CompilationPass.er_to_circuit`. (This is the mod in my last commit).
By the same logic, it may make sense to remap the Pauli strings in the CliffordMap as well, even if it's not needed for the current implementation of the `StimCliffordPass`. What do you think?

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,475 @@
...
+        """
+        x_map = {inputs_mapping(node): ps.remap(outputs_mapping) for node, ps in self.x_map.items()}
+        z_map = {inputs_mapping(node): ps.remap(outputs_mapping) for node, ps in self.z_map.items()}
+        return replace(self, x_map=x_map, z_map=z_map)
+
+
+def extend_input(og: OpenGraph[Measurement]) -> tuple[OpenGraph[Measurement], dict[int, int]]:
+    r"""Extend the inputs of a given open graph.
+
+    For every input node :math:`v`, a new node :math:`u` and edge :math:`(u, v)` are added to the open graph. Node :math:`u` is measured in plane :math:`XY` with angle :math:`\alpha = 0` and replaces :math:`v` in the open graph's sequence of input nodes.
```
```suggestion
    For every input node :math:`v`, a new node :math:`u` and edge :math:`(u, v)` are added to the open graph. Node :math:`u` is measured in Pauli axis `X` and replaces :math:`v` in the open graph's sequence of input nodes.
```

graphix/circ_ext/compilation.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+        if pexp.angle == 0:  # No rotation
+            return
+
+        # We assume that nodes in the Pauli strings have been mapped to qubits.
+        modified_qubits = [
+            qubit
+            for qubit in range(circuit.width)
+            if qubit in pexp.pauli_string.x_nodes | pexp.pauli_string.y_nodes | pexp.pauli_string.z_nodes
+        ]
```
thierry-martinez: All tests pass with the following change:
```suggestion
        modified_qubits = list(pexp.pauli_string.x_nodes | pexp.pauli_string.y_nodes | pexp.pauli_string.z_nodes)
```

I realize we discussed this earlier and the ladder transformation output does change with this modification. However, if the order matters, should we add a test to enforce it?

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,475 @@
...
+        Nodes on which a Pauli Y operator is applied.
+    z_nodes : AbstractSet[int]
+        Nodes on which a Pauli Z operator is applied.
+    sign : Sign
+        Phase of the Pauli string.
+    """
+
+    x_nodes: AbstractSet[int] = dataclasses.field(default_factory=frozenset)
+    y_nodes: AbstractSet[int] = dataclasses.field(default_factory=frozenset)
+    z_nodes: AbstractSet[int] = dataclasses.field(default_factory=frozenset)
```
thierry-martinez: Since `x_nodes`, `y_nodes`, and `z_nodes` are disjoint, I propose representing `PauliString` as `Mapping[int, Axis]`, which is more canonical. See commit https://github.com/thierry-martinez/graphix/commit/828f96dc5ebe1ee0bf1f9955a52bbe92ae10bde0 (and in https://github.com/thierry-martinez/graphix-stim-compiler/commit/a743335d6d60df50394b5992f51d3a2129810339 for `graphix-stim-compiler`).

thierry-martinez: (By "more canonical", I mean that disjointness is structurally enforced.)

matulni: Hi @thierry-martinez, thanks for this comment, indeed it bothered me that disjointness was not structurally enforced. I like your mapping idea, but I fear that using `Axis` may be a bit confusing: in Graphix this object is associated with measurements (or at least, with a notion of space), and I'm not sure if it makes much sense in this context.
What do you think about using `graphix.Pauli` objects ? Another advantage would be that we can explicitly map nodes to the identity (now, they are implicitly mapped by not being in the sets (or in the dictionary keys in your commit)). Including identity operators will certainly be less efficient, but all the Pauli-string processing routines will remain with the same complexity, so I think it's not a big problem (and it may improve readability). If you agree, I'll adapt your commit accordingly.

graphix/circ_ext/compilation.py
matulni: Hi @thierry-martinez, this is tricky.

From a code perspective, I think that in the current Python implementation, calling `list` on a `set` of (small?) integers always returns an ordered list:

```python
assert [0, 1, 2] == list[{1, 0, 2}]
```

Ultimately, I think that this is because (small?) integers are hashed with their own value:

```python
a, b = 0, 1
assert a.__hash__() == 0
assert b.__hash__() == 1
```

So in our case, since we are dealing with small integers (number of qubits), in both the PR's implementation  and your suggestion, we get the same list of instructions (and hence all tests pass trivially).

Below, I paste a `pexp_ladder_pass` function where you can specify the qubit order by hand:

```python
from itertools import pairwise, chain
from graphix.circ_ext.extraction import PauliExponential, PauliExponentialDAG
from graphix.fundamentals import ANGLE_PI


def pexp_ladder_pass(pexp_dag: PauliExponentialDAG, circuit: Circuit, modified_qubits: list[int]) -> None:
    r"""Add a Pauli exponential DAG to a circuit by using a ladder decomposition.

    The input circuit is modified in-place. This function assumes that the Pauli exponential DAG has been remap, i.e., its Pauli strings are defined on qubit indices instead of output nodes. See :meth:`PauliString.remap` for additional information.

    Parameters
    ----------
    pexp_dag: PauliExponentialDAG
        The Pauli exponential rotation to be added to the circuit. Its Pauli strings are assumed to be defined on qubit indices.
    circuit : Circuit
        The circuit to which the operation is added. The input circuit is assumed to be compatible with ``pexp_dag.output_nodes``.

    Notes
    -----
    Pauli exponentials in the DAG are compiled sequentially following an arbitrary total order compatible with the DAG. Each Pauli exponential is decomposed into a sequence of basis changes, CNOT gates, and a single :math:`R_Z` rotation:

    .. math::

        R_Z(\phi) = \exp \left(-i \frac{\phi}{2} Z \right),

    with effective angle :math:`\phi = -2\alpha`, where :math:`\alpha` is the angle encoded in `self.angle`. Basis changes map :math:`X` and :math:`Y` operators to the :math:`Z` basis before entangling the qubits in a CNOT ladder.

    Gate set: H, CNOT, RZ, RY

    See https://quantumcomputing.stackexchange.com/questions/5567/circuit-construction-for-hamiltonian-simulation/11373#11373 for additional information.
    """

    def add_pexp(pexp: PauliExponential, circuit: Circuit) -> None:
        r"""Add the Pauli exponential unitary to a quantum circuit.

        This function modifies the input circuit in-place.

        Parameters
        ----------
        circuit : Circuit
            The quantum circuit to which the Pauli exponential is added.

        Notes
        -----
        It is assumed that the ``x``, ``y``, and ``z`` node sets of the Pauli string in the exponential are well-formed, i.e., contain valid qubit indices and are pairwise disjoint.
        """
        if pexp.angle == 0:  # No rotation
            return

        angle = -2 * pexp.angle * pexp.pauli_string.sign

        if len(modified_qubits) == 0:  # Identity
            return

        q0 = modified_qubits[0]

        if len(modified_qubits) == 1:
            if q0 in pexp.pauli_string.x_nodes:
                circuit.rx(q0, angle)
            elif q0 in pexp.pauli_string.y_nodes:
                circuit.ry(q0, angle)
            else:
                circuit.rz(q0, angle)
            return

        add_basis_change(pexp, q0, circuit)

        for q1, q2 in pairwise(modified_qubits):
            add_basis_change(pexp, q2, circuit)
            circuit.cnot(control=q1, target=q2)

        circuit.rz(modified_qubits[-1], angle)

        for q2, q1 in pairwise(modified_qubits[::-1]):
            circuit.cnot(control=q1, target=q2)
            add_basis_change(pexp, q2, circuit)

        add_basis_change(pexp, modified_qubits[0], circuit)

    def add_basis_change(pexp: PauliExponential, qubit: int, circuit: Circuit) -> None:
        """Apply an X or a Y basis change to a given qubit if required by the Pauli string.

        This function modifies the input circuit in-place.

        Parameters
        ----------
        pexp : PauliExponential
            The Pauli exponential under consideration.
        qubit : int
            The qubit on which the basis-change operation is performed.
        circuit : Circuit
            The quantum circuit to which the basis change is added.
        """
        if qubit in pexp.pauli_string.x_nodes:
            circuit.h(qubit)
        elif qubit in pexp.pauli_string.y_nodes:
            add_hy(qubit, circuit)

    def add_hy(qubit: int, circuit: Circuit) -> None:
        """Add a pi rotation around the z + y axis.

        This function modifies the input circuit in-place.

        Parameters
        ----------
        qubit : int
            The qubit on which the basis-change operation is performed.
        circuit : Circuit
            The quantum circuit to which the basis change is added.
        """
        circuit.rz(qubit, ANGLE_PI / 2)
        circuit.ry(qubit, ANGLE_PI / 2)
        circuit.rz(qubit, ANGLE_PI / 2)

    for node in chain(*reversed(pexp_dag.partial_order_layers[1:])):
        pexp = pexp_dag.pauli_exponentials[node]
        add_pexp(pexp, circuit)
```

If you compile a Pauli exponential with different `modified_qubit` order, you get a different circuit (with the PR's implementation or your commit you always get the same set of instructions):

```python
from graphix.circ_ext.compilation import pexp_ladder_pass
from graphix.circ_ext.extraction import PauliString, PauliExponential, PauliExponentialDAG
from graphix.sim.base_backend import NodeIndex
from graphix.fundamentals import ANGLE_PI

alpha = 0.1

pexp = PauliExponentialDAG(
                    pauli_exponentials={
                        0: PauliExponential(alpha / 2, PauliString(x_nodes={4,3}, y_nodes={1, 2})),
                    },
                    partial_order_layers=[{1, 2, 3, 4}, {0}],
                    output_nodes=[1, 2, 3, 4],
                )


outputs_mapping = NodeIndex()
outputs_mapping.extend(pexp.output_nodes)
pexp_rmp = pexp.remap(outputs_mapping.index)

qc1 = Circuit(4)
pexp_ladder_pass_2(pexp_rmp, qc1, [0, 1, 2, 3])

qc2 = Circuit(4)
pexp_ladder_pass_2(pexp_rmp, qc2, [0, 2, 1, 3])
assert qc1.instruction != qc2.instruction
```

But to my surprise, they seem to represent the same unitary, up to a phase:

```python
s1 = qc1.simulate_statevector().statevec
s2 = qc2.simulate_statevector().statevec
assert s1.isclose(s2)
```

In summary:
- I think that Python's implementation makes it difficult to check the different between ordering the qubits because of how integers are hashed.
- I haven't managed to find an example where the order matters when we specify it by hand.
- I'm not convinced that the order does not matter in general.

matulni: *Edit*: not even up to a phase:

```python
import numpy as np
assert(np.all(np.isclose(s1.flatten(),s2.flatten())))
```

graphix/circ_ext/extraction.py
thierry-martinez: I think that `Axis` expresses precisely what we want to enforce structurally (at least for the value part; keys are not required to be actual nodes). `Axis` is defined in `fundamentals` and is a subtype of `IXYZ`, which is a component of `Pauli`. As you mentioned, using `IXYZ` or `Pauli` would allow nodes to be mapped to `I`; therefore, we lose canonicality because mapping to `I` is equivalent to not being mapped. This should not be a serious problem, but the invariant that the set of keys equals precisely the nodes not mapped to `I` is nice to have, and defining a rotation over `I` by a given angle is a bit strange (it would be the identity, of course). `Pauli` is a pair of `IXYZ` and `Sign`, and we will not use the `Sign` either, which is also a loss of canonicality.

thierry-martinez: (In other words, `Pauli` is a Pauli string of length 1.)

graphix/circ_ext/compilation.py
matulni: Ok, I convinced myself that **the order on which we iterate over the qubits** when synthesising the Pauli exponential **does not matter**.

Consider the ZZ-Pauli exponential:

$$U(\alpha) = \exp \left(i \frac{\alpha}{2} Z_0 Z_1 \right) =\text{diag}\left[e^{i \alpha/2}, e^{-i \alpha/2}, e^{-i \alpha/2}, e^{i \alpha/2}\right].$$

The star-pass synthesis in the order $[q_0, q_1]$ yields
```
q_0: ──■─────────────■──
     ┌─┴─┐┌───────┐┌─┴─┐
q_1: ┤ X ├┤ Rz(α) ├┤ X ├
     └───┘└───────┘└───┘
```
whereas the synthesis in the order $[q_1, q_0]$ yields

```
     ┌───┐┌───────┐┌───┐
q_0: ┤ X ├┤ Rz(α) ├┤ X ├
     └─┬─┘└───────┘└─┬─┘
q_1: ──■─────────────■──
```

and these circuits represent the same unitary. To see that, first, note that they are equivalent up to a relabelling $q_0 \leftrightarrow q_1$. Then, notice that according to Eq. (1) the computational-basis states that are sensitive to a relabelling, i.e., $|01\rangle$ and $|10\rangle$, pick up the same phase, $e^{-i \alpha/2}$. In other words,

$$ U(\alpha) = \text{SWAP} \cdot U(\alpha) \cdot \text{SWAP}^{\dagger}.$$

This is true for $N$-qubit case too $U_N(\alpha) = \exp \left(i \frac{\alpha}{2} Z_0 \dots Z_N \right)$, since all the computational basis states that are connected by a permutation have the same number of non-zero bits, and therefore pick up the same phase, $U(\alpha) |n\rangle = e^{(-1)^m i \alpha/2} |n\rangle$, with $m$ the number of 1s in `bin(n)`. If there are $M$-identities in the Pauli string, we go back to the case $U_{N-M}(\alpha)$, so we can prove it by induction.

The argument does not change if we have $X$ or $Y$ gates in the Pauli string, since we simply proceed by doing a change of basis (i.e., applying $H$ or $H_Y$ on the corresponding qubit), which is independent of the synthesis order.

graphix/circ_ext/extraction.py
matulni: Fixed in [2b4a008](https://github.com/TeamGraphix/graphix/pull/445/commits/2b4a008a76eec92ae77703ec4f06892955cbc4cc)

graphix/circ_ext/compilation.py
matulni: Unnecessary sorting removed in [2b4a008](https://github.com/TeamGraphix/graphix/pull/445/commits/2b4a008a76eec92ae77703ec4f06892955cbc4cc)

graphix/circ_ext/extraction.py
matulni: Fair enough, I'm convinced, thanks for the comment :)

graphix/circ_ext/compilation.py
thierry-martinez: Thank you very much for the explanation! Regarding the points you mentioned in the comment above, `hash` is indeed the identity over $[0, 2^{60}]$. However, the iteration order of a set is unspecified, and there is no guarantee that elements are enumerated in order of their hash values, nor that this order is stable from one Python version to another: for instance, `list(set([1,2])) == [1,2]` whereas `list(set([1,8])) == [8,1]`.

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,477 @@
...
+    """Dataclass representing a Pauli string over a set of MBQC nodes.
+
+    Attributes
+    ----------
+    x_nodes : AbstractSet[int]
+        Nodes on which a Pauli X operator is applied.
+    y_nodes : AbstractSet[int]
+        Nodes on which a Pauli Y operator is applied.
+    z_nodes : AbstractSet[int]
+        Nodes on which a Pauli Z operator is applied.
```
thierry-martinez: Oops, I forgot to update this!

```suggestion
    axes : Mapping[int, Axis]
        Nodes on which a Pauli operator X, Y, or Z is applied.
```

tests/test_circ_extraction.py
```diff
@@ -0,0 +1,221 @@
...
+from graphix.sim.base_backend import NodeIndex
+from graphix.states import BasicStates
+from graphix.transpiler import Circuit
+
+if TYPE_CHECKING:
+    from numpy.random import Generator
+
+
+class PauliExpTestCase(NamedTuple):
+    p_exp: PauliExponentialDAG
```
pranav97nair: Small nitpick - I would rename this variable `pexp_dag` to match the convention you use everywhere else in the codebase and to minimize the chances of it being confused with the `PauliExponential` object.

matulni: Done in [9137a7d](https://github.com/TeamGraphix/graphix/pull/445/commits/9137a7d830147eecfe7998addc6dd11beb84db43)

graphix/circ_ext/compilation.py
```diff
@@ -0,0 +1,181 @@
...
+
+    Notes
+    -----
+    Pauli exponentials in the DAG are compiled sequentially following an arbitrary total order compatible with the DAG. Each Pauli exponential is decomposed into a sequence of basis changes, CNOT gates, and a single :math:`R_Z` rotation:
+
+    .. math::
+
+        R_Z(\phi) = \exp \left(-i \frac{\phi}{2} Z \right),
+
+    with effective angle :math:`\phi = -2\alpha`, where :math:`\alpha` is the angle encoded in `self.angle`. Basis changes map :math:`X` and :math:`Y` operators to the :math:`Z` basis before entangling the qubits in a CNOT ladder.
```
thierry-martinez:
```suggestion
    with effective angle :math:`\phi = -2\alpha`, where :math:`\alpha` is the angle encoded in ``self.angle``. Basis changes map :math:`X` and :math:`Y` operators to the :math:`Z` basis before entangling the qubits in a CNOT ladder.
```

graphix/circ_ext/compilation.py
```diff
@@ -0,0 +1,181 @@
...
+
+    This method synthesizes a circuit by sequentially applying the Clifford map and the Pauli exponential DAG (Directed Acyclic Graph) in the extraction result. It performs a validation check to ensure that the output nodes of both components are identical and it maps the output node numbers to qubit indices.
+
+    Parameters
+    ----------
+    er : ExtractionResult
+        The result of the extraction process, containing both the ``clifford_map`` and the ``pexp_dag``.
+    pexp_cp: Callable[[PauliExponentialDAG, Circuit], None] | None
+        Compilation pass to synthetize a Pauli exponential DAG. If ``None`` (default), :func:`pexp_ladder_pass` is employed.
+    cm_cp: Callable[[PauliExponentialDAG, Circuit], None] | None
```
thierry-martinez:
```suggestion
    cm_cp: Callable[[CliffordMap, Circuit], None] | None
```

graphix/circ_ext/compilation.py
```diff
@@ -0,0 +1,181 @@
...
+    """Convert a circuit extraction result into a quantum circuit representation.
+
+    This method synthesizes a circuit by sequentially applying the Clifford map and the Pauli exponential DAG (Directed Acyclic Graph) in the extraction result. It performs a validation check to ensure that the output nodes of both components are identical and it maps the output node numbers to qubit indices.
+
+    Parameters
+    ----------
+    er : ExtractionResult
+        The result of the extraction process, containing both the ``clifford_map`` and the ``pexp_dag``.
+    pexp_cp: Callable[[PauliExponentialDAG, Circuit], None] | None
+        Compilation pass to synthetize a Pauli exponential DAG. If ``None`` (default), :func:`pexp_ladder_pass` is employed.
```
thierry-martinez:
```suggestion
        Compilation pass to synthesize a Pauli exponential DAG. If ``None`` (default), :func:`pexp_ladder_pass` is employed.
```

graphix/circ_ext/compilation.py
```diff
@@ -0,0 +1,181 @@
...
+    cm_cp: Callable[[PauliExponentialDAG, Circuit], None] | None
+        Compilation pass to synthetize a Clifford map. If ``None`` (default), a `ValueError` is raised since there is still no default pass for Clifford map integrated in Graphix.
```
thierry-martinez:
```suggestion
        Compilation pass to synthesize a Clifford map. If ``None`` (default), a `ValueError` is raised since there is still no default pass for Clifford map integrated in Graphix.
```

graphix/circ_ext/compilation.py
```diff
@@ -0,0 +1,181 @@
...
+
+    cm_cp(er.clifford_map.remap(inputs_mapping.index, outputs_mapping.index), circuit)
+    pexp_cp(er.pexp_dag.remap(outputs_mapping.index), circuit)
+    return circuit
+
+
+def pexp_ladder_pass(pexp_dag: PauliExponentialDAG, circuit: Circuit) -> None:
+    r"""Add a Pauli exponential DAG to a circuit by using a ladder decomposition.
+
+    The input circuit is modified in-place. This function assumes that the Pauli exponential DAG has been remap, i.e., its Pauli strings are defined on qubit indices instead of output nodes. See :meth:`PauliString.remap` for additional information.
```
thierry-martinez:
```suggestion
    The input circuit is modified in-place. This function assumes that the Pauli exponential DAG has been remapped, i.e., its Pauli strings are defined on qubit indices instead of output nodes. See :meth:`PauliString.remap` for additional information.
```

graphix/circ_ext/compilation.py
```diff
@@ -0,0 +1,181 @@
...
+    """
+
+    def add_pexp(pexp: PauliExponential, circuit: Circuit) -> None:
+        r"""Add the Pauli exponential unitary to a quantum circuit.
+
+        This function modifies the input circuit in-place.
+
+        Parameters
+        ----------
+        circuit : Circuit
```
thierry-martinez, line 107:
```suggestion
        pexp: PauliExponential
            The Pauli exponential to add.
        circuit : Circuit
```

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,478 @@
...
+    Attributes
+    ----------
+    axes : Mapping[int, Axis]
+        Mapping between nodes and the applied Pauli operator.
+    sign : Sign
+        Phase of the Pauli string.
+
+    Notes
+    -----
+    The identity operator is ommitted in this representation, which means that in general it is not possible to infer the size of the Hilbert space from an instance of `PauliString` alone.
```
```suggestion
    The identity operator is omitted in this representation, which means that in general it is not possible to infer the size of the Hilbert space from an instance of `PauliString` alone.
```

graphix/circ_ext/compilation.py
thierry-martinez:
```suggestion
        Compilation pass to synthesize a Pauli exponential DAG. If ``None`` (default), :func:`pexp_ladder_pass` is employed.
```

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,478 @@
...
+        cm_cp: Callable[[CliffordMap, Circuit], None] | None = None,
+    ) -> Circuit:
+        """Transpile the extraction result to circuit.
+
+        Transpilation is only supported when the pair Pauli-exponential DAG and Clifford map represents a unitary transformation.
+
+        Parameters
+        ----------
+        pexp_cp: Callable[[PauliExponentialDAG, Circuit], None] | None
+            Compilation pass to synthetize a Pauli exponential DAG. If ``None`` (default), :func:`pexp_ladder_pass` is employed.
```
```suggestion
            Compilation pass to synthesize a Pauli exponential DAG. If ``None`` (default), :func:`pexp_ladder_pass` is employed.
```

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,478 @@
...
+        cm_cp: Callable[[PauliExponentialDAG, Circuit], None] | None
+            Compilation pass to synthetize a Clifford map. If ``None`` (default), a `ValueError` is raised since there is still no default pass for Clifford map integrated in Graphix.
```
```suggestion
            Compilation pass to synthesize a Clifford map. If ``None`` (default), a `ValueError` is raised since there is still no default pass for Clifford map integrated in Graphix.
```

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,478 @@
...
+
+@dataclass(frozen=True)
+class PauliExponential:
+    r"""Dataclass representing a Pauli exponential over a set of MBQC nodes.
+
+    A Pauli exponential corresponds to the unitary operator
+
+    .. math::
+
+        U(\alpha) = \exp \left(i \frac{alpha}{2} P\right),
```
```suggestion
        U(\alpha) = \exp \left(i \frac{\alpha}{2} P\right),
```

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,478 @@
...
+        This routine associates a Pauli exponential to each measured node in ``flow``. The flow's partial order defines a partial order between the Pauli exponentials such that Pauli exponentials in the same layer commute.
+
+        Parameters
+        ----------
+        flow : PauliFlow[Measurement]
+            A focused Pauli flow.
+
+        Returns
+        -------
+        PauliExponentialRotation
```
```suggestion
        PauliExponentialDAG
```

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,478 @@
...
+    x_map: Mapping[int, PauliString]
+    z_map: Mapping[int, PauliString]
+    input_nodes: Sequence[int]
+    output_nodes: Sequence[int]
+
+    @staticmethod
+    def from_focused_flow(flow: PauliFlow[Measurement]) -> CliffordMap:
+        """Extract a Clifford map from a focused Pauli flow.
+
+        This routine associates a two Pauli strings (one per generator of the Pauli group, X and Z) to each input node in ``flow.og``.
```
```suggestion
        This routine associates two Pauli strings (one per generator of the Pauli group, X and Z) to each input node in ``flow.og``.
```

graphix/circ_ext/compilation.py
thierry-martinez:
```suggestion
        Compilation pass to synthetize a Clifford map. If ``None`` (default), a ``ValueError`` is raised since there is still no default pass for Clifford map integrated in Graphix.
```

graphix/circ_ext/extraction.py
```suggestion
    The identity operator is ommitted in this representation, which means that in general it is not possible to infer the size of the Hilbert space from an instance of ``PauliString`` alone.
```

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,478 @@
...
+        ----------
+        flow : PauliFlow[Measurement]
+            A focused Pauli flow. The resulting Pauli string is extracted from its correction function.
+        node : int
+            A measured node whose associated Pauli string is computed.
+
+        Returns
+        -------
+        PauliExponential
+            Primary extraction string associated to the input measured nodes. The sets in the returned `PauliString` instance are disjoint.
```
```suggestion
            Primary extraction string associated to the input measured nodes. The sets in the returned ``PauliString`` instance are disjoint.
```

graphix/circ_ext/extraction.py
```diff
@@ -0,0 +1,478 @@
...
+@dataclass(frozen=True)
+class PauliExponentialDAG:
+    """Dataclass to represent a multi-qubit rotation formed by a sequence of Pauli exponentials extracted from a pattern.
+
+    Attributes
+    ----------
+    pauli_exponentials: Mapping[int, PauliExponential]
+        Mapping between measured nodes (``keys``) and Pauli exponentials (``values``).
+    partial_order_layers: Sequence[AbstractSet[int]]
+        Partial order between the Pauli exponentials in a layer form. The set `layers[i]` comprises the nodes in layer `i`. Nodes in layer `i` are "larger" in the partial order than nodes in layer `i+1`. The pattern's output nodes are always in layer 0.
```
```suggestion
        Partial order between the Pauli exponentials in a layer form. The set ``layers[i]`` comprises the nodes in layer ``i``. Nodes in layer ``i`` are "larger" in the partial order than nodes in layer ``i+1``. The pattern's output nodes are always in layer 0.
```

PR #447, opened 2026-02-20
Add `perform_pauli_pushing` to `Pattern`

thierry-martinez:
This commit exposes the method `perform_pauli_pushing` in `Pattern`.

The method performs the Pauli pushing in place, unless called with `copy=True`, in which case a new pattern is returned.

By default, the method uses `to_space_optimal_pattern` to prepare the nodes on a need-by-need basis. If called with `standardize=True`, a standardized pattern is returned instead.

Warnings are issued if Pauli nodes have not been inferred or if `leave_nodes` contains non-Pauli nodes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Moved the conditional logic to `graphix.simulator` to remove code duplication in the backends.
   - Solves [#428](https://github.com/TeamGraphix/graphix/issues/428).
 
+- #438: `ComplexUnit.try_from` now uses `cmath.isclose` for float comparison and has optional parameters `rel_tol` and `abs_tol`.
+
 - #440, #441: `Statevec.nqubits` now returns the correct value.
 
 ### Changed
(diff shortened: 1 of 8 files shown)


graphix/optimization.py
```diff
@@ -244,8 +245,38 @@ def extract_graph(self) -> nx.Graph[int]:
             graph.add_edge(u, v)
         return graph
 
-    def perform_pauli_pushing(self, leave_nodes: set[Node] | None = None) -> Self:
-        """Move all Pauli measurements before the other measurements (except nodes in `leave_nodes`)."""
+    def perform_pauli_pushing(self, leave_nodes: set[Node] | None = None, *, stacklevel: int = 1) -> Self:
```
matulni: Should we type with `collections.abc.set` (`AbstractSet`) instead ?

graphix/optimization.py
```diff
@@ -244,8 +245,38 @@ def extract_graph(self) -> nx.Graph[int]:
...
+
+        if leave_nodes:
+            leave_non_pauli_nodes = [
+                cmd.node
+                for cmd in self.m_list
+                if not isinstance(cmd.measurement, PauliMeasurement) and cmd.node in leave_nodes
+            ]
+            if leave_non_pauli_nodes:
+                warn(
+                    f"`leave_nodes` contains nodes that are not Pauli: {leave_non_pauli_nodes}.",
```
matulni:
```suggestion
                    f"`leave_nodes` contains nodes that are not Pauli: {leave_non_pauli_nodes}. The constraint has no effect on these nodes.",
```
(or something along those lines...)

graphix/optimization.py
thierry-martinez: Indeed. Fixed in 6be7458.

graphix/optimization.py
```diff
@@ -244,8 +245,38 @@ def extract_graph(self) -> nx.Graph[int]:
...
+                for cmd in self.m_list
+                if not isinstance(cmd.measurement, PauliMeasurement) and cmd.node in leave_nodes
+            ]
+            if leave_non_pauli_nodes:
+                warn(
+                    f"`leave_nodes` contains nodes that are not Pauli: {leave_non_pauli_nodes}. The constraint has no effect on these nodes.",
+                    stacklevel=stacklevel + 1,
+                )
+
         if leave_nodes is None:
```
emlynsg: Update to `else`

thierry-martinez: I removed the `if leave_nodes is None:` block in 30b87e2.

PR #452, opened 2026-02-26
Adding uv to the Github CI and requirements to `pyproject`

### LINKED ISSUE #443: Replace `nox` with `uv` in CI
Considering upgrading the CI from `nox`-based, to `uv`-based.
There are 3 available approaches:
- Use `nox` as-is, with `pip` backend
- Use `nox` with `uv` backend
- Use pure `uv`

Performance was tested on two runs of the CI across Python 3.10 to 3.13 (see [example](https://github.com/qat-inria/graphix/actions/runs/22100090207)). 
Relative to the run-time of the current `nox`-based CI, **`nox` with `uv` backend was faster** (took ~77\% of the time), **and pure `uv` slightly faster again** (took ~73\% of the time). My suggestion is that we maintain `nox`, which is cleaner to have in the CI testing across many platforms, but move to a `uv` backend to get most of the speed-up.

At the same time, this change would allow us to pick up `uv` for our development workflow (and wouldn't prevent other approaches). It would:
- simplify the repo (we can remove `requirements.txt` files, absorbing dependencies into the `pyproject.toml`)
    - users install either with `pip install -e .[dev]` or more easily with `uv sync --all-extras`
- we could commit the `uv.lock` file, and periodically update it, if we wanted to have more control over dependencies
    - having this in the repo would mean exactly replicable development environments by default (no new venv/conda env or pip install steps)

emlynsg:
This PR changes Graphix to use `uv` instead of `pip` in the `nox` routine for CI tests.
To adhere to the requirements of `uv`, it also includes a `uv.lock` file locking dependencies for CI reproducibility, and has merged requirements into the `pyproject.toml` file in line with PEP best practice (https://packaging.python.org/en/latest/specifications/pyproject-toml/). A new Github action has been created to check the uv lockfile matches the pyproject.

The installation process remains the same with this change - either `pip install -e .` or `uv sync` (the latter preferred for reproducibility and speed). See changes to the `CONTRIBUTING.md` and `README.md` for more details.

This PR also turns off `ruff` preview mode, to keep linting more stable, alongside changing the dependabot to use `uv`. Changes to Graphix internal files are due to linting changes introduced by this reversion.


README.md
```diff
@@ -28,12 +28,13 @@ Install `graphix` with `pip`:
...
 ```
 
-Install together with [extra packages](https://github.com/TeamGraphix/graphix/blob/master/requirements-extra.txt):
+Install with [extra dependencies](https://github.com/TeamGraphix/graphix/blob/master/pyproject.toml) (e.g. pyzx):
 
 ```bash
 pip install graphix[extra]
 ```
 
+Graphix also supports using `uv`.
```
matulni: Maybe add a bash line to show how to install using `uv`

docs/source/conf.py
```diff
@@ -14,7 +14,7 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
 project = "graphix"
-copyright = "2022, Team Graphix"
+copyright = "2026, Team Graphix"
```
thierry-martinez: We could put a year range: `2022-2026`.

docs/source/contributing.rst
```diff
@@ -3,24 +3,33 @@ Contributing to Graphix
...
+    uv run pytest
 
-    pip install -c requirements-dev.txt nox
-    nox --python 3.12
+Format code before committing::
 
-    pip install .[dev]
-    pytest --cov=./graphix --cov-report=xml --cov-report=term
+    uv run ruff check --select I --fix .
+    uv run ruff format .
```
thierry-martinez, line 33: Could we add an equivalent to the code coverage command from before:
```
uv run pytest --cov=./graphix --cov-report=xml --cov-report=term --doctest-modules
```

.github/workflows/ruff.yml
```diff
@@ -16,36 +16,29 @@ jobs:
...
-        run: python -m pip install --upgrade pip && python -m pip install ruff -c requirements-dev.txt
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          enable-cache: true
+          cache-dependency-glob: "pyproject.toml"
 
       - name: Run Ruff Linter
-        run: ruff check --output-format=github
+        run: uv run --extra dev ruff check --output-format=github
```
thierry-martinez: Previously, we installed just `ruff` (constrained by `requirements-dev.txt`) instead of all dev dependencies: `pip install ruff -c requirements-dev.txt`

With `uv`, the equivalent is:
```suggestion
        run: uv run --with ruff ruff check --output-format=github
```

.github/workflows/ruff.yml
```diff
@@ -16,36 +16,29 @@ jobs:
...
-      - name: Install dependencies
-        run: python -m pip install --upgrade pip && python -m pip install ruff -c requirements-dev.txt
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          enable-cache: true
 
       - name: Run Ruff Formatter Check
-        run: ruff format --check
+        run: uv run --extra dev ruff format --check
```
thierry-martinez:
```suggestion
        run: uv run --with ruff ruff format --check
```

docs/source/contributing.rst
```diff
@@ -3,24 +3,33 @@ Contributing to Graphix
...
 This page summarises useful tips for developing the project locally.
 
+Setup
+-----
+
+Clone the repository and install all development dependencies::
+
+    git clone https://github.com/TeamGraphix/graphix.git
+    cd graphix
+    uv sync --extra dev --extra extra
```
thierry-martinez, line 14: I would add the step `pip install uv`, since `uv` is not part of the standard Python distribution.

CONTRIBUTING.md
```diff
@@ -9,48 +9,36 @@ The aim of `graphix` is to make the measurement-based quantum computing (MBQC) a
...
 We recommend to [fork the repository](https://docs.github.com/en/get-started/quickstart/fork-a-repo) before working on new features, whether that being an existing issue or your own idea.
-Once created, you'll need to clone the repository, and you can follow below to set up the environment. You may want to set up virtual environment, such as `conda env` or `pipenv` before setting up the environment.
+Once created, you'll need to clone the repository, and you can follow below to set up the environment. We recommend you use a virtual environment like `uv`, but you may prefer to use `conda env` or `pipenv`.
 
 ```bash
-git clone git@github.com:<username>/graphix.git
+git clone git@github.com:TeamGraphix/graphix.git
 cd graphix
-pip install -e .[dev]
+uv sync --extra dev --extra extra
```
thierry-martinez, line 18: I would recommend to add `pip install uv` here as well.

CONTRIBUTING.md
```diff
@@ -9,48 +9,36 @@ The aim of `graphix` is to make the measurement-based quantum computing (MBQC) a
...
+uv run nox
 ```
 
-With the development dependencies installed, run the test suite explicitly:
+Before committing, format the code:
 
 ```bash
-pip install .[dev]
-pytest --cov=./graphix --cov-report=xml --cov-report=term
+uv run ruff check --select I --fix .
```
thierry-martinez: If I understand correctly, the rule `I` is already selected in `pyproject.toml`.

```suggestion
uv run ruff check --fix .
```

I would add a point about test coverage: `uv run pytest --cov=./graphix --cov-report=xml --cov-report=term --doctest-modules`

CONTRIBUTING.md
```diff
@@ -60,38 +48,35 @@ enable the linter and basic type checking on save:
...
 
-Neither `mypy` nor `pyright` detect editable packages (installed with
-`pip install -e`), unless they are installed in `compat` or `strict`
-mode. A workaround is to install `graphix` in `strict` mode by running
-the following command in the `graphix` repository.
+Or with pip:
 
 ```bash
-pip install -e . --config-settings editable_mode=strict
+pip install . --config-settings editable_mode=strict
```
thierry-martinez: The `--config-settings editable_mode=strict` flag is only useful with `-e` (editable installs); it ensures type-checking works correctly with editable packages. I don't know if `uv` offers an equivalent setting for installing packages as editable while preserving type-checking support.

pyproject.toml
```diff
@@ -24,24 +24,66 @@ classifiers = [
...
+    "nox==2026.2.9",
+    "psutil",
+    "pytest",
+    "pytest-benchmark",
+    "pytest-cov",
+    "pytest-mock",
+    "pytest-mpl",
+    "qiskit>=1.0",
+    "qiskit_qasm3_import",
+    "qiskit-aer; python_version < '3.14'",
```
thierry-martinez: This Python version constraint can be removed (see #462).

```suggestion
    "qiskit-aer",
```

thierry-martinez (commented):
Very nice! Thanks. I have some minor remarks. I also question whether we should version `uv.lock`, since it appears to be a large generated file likely to cause merge conflicts. I'd prefer to stick with our policy of specifying version constraints in `pyproject.toml` on a case-by-case basis.

CONTRIBUTING.md
emlynsg: The command I've changed it to seems to work. We can also put this in the pyproject.toml to enforce strict editable mode always, but not sure if that's what we want?

CONTRIBUTING.md
```diff
@@ -9,48 +9,38 @@ The aim of `graphix` is to make the measurement-based quantum computing (MBQC) a
 ## Getting started working on your contribution for `graphix`
 
 We recommend to [fork the repository](https://docs.github.com/en/get-started/quickstart/fork-a-repo) before working on new features, whether that being an existing issue or your own idea.
-Once created, you'll need to clone the repository, and you can follow below to set up the environment. You may want to set up virtual environment, such as `conda env` or `pipenv` before setting up the environment.
+Once created, you'll need to clone the repository, and you can follow below to set up the environment. We recommend you use a virtual environment like `uv`, but you may prefer to use `conda env` or `pipenv`.
```
matulni:
```suggestion
Once created, you'll need to clone the repository, and you can follow below to set up the environment. We recommend you use a project manager like `uv`, but you may prefer to manage your own virtual environment with `conda env` or `pipenv`.
```

CONTRIBUTING.md
```diff
@@ -9,48 +9,38 @@ The aim of `graphix` is to make the measurement-based quantum computing (MBQC) a
...
+uv run ruff check .
+uv run ruff format --check .
+uv run mypy
+uv run pytest
 ```
 
-### Local checks
-
-To replicate the CI pipeline locally, install `nox` and run the tests:
+You can also run the full CI suite locally:
```
matulni:
```suggestion
You can also run the full CI test suite locally:
```
My understanding is that our `nox` file doesn't do anything about typecheckers or linters.

CONTRIBUTING.md
```diff
@@ -9,48 +9,38 @@ The aim of `graphix` is to make the measurement-based quantum computing (MBQC) a
...
 
 ```bash
-# Be sure to use the latest version of ruff
-pip install -U ruff
-# Sort imports and format
-ruff check --select I --fix .
-ruff format .
+uv run ruff check .
+uv run ruff format --check .
+uv run mypy
```
matulni, line 28:
```suggestion
uv run mypy
uv run pyright
```

CONTRIBUTING.md
```diff
@@ -9,48 +9,38 @@ The aim of `graphix` is to make the measurement-based quantum computing (MBQC) a
 ## Getting started working on your contribution for `graphix`
 
 We recommend to [fork the repository](https://docs.github.com/en/get-started/quickstart/fork-a-repo) before working on new features, whether that being an existing issue or your own idea.
-Once created, you'll need to clone the repository, and you can follow below to set up the environment. You may want to set up virtual environment, such as `conda env` or `pipenv` before setting up the environment.
+Once created, you'll need to clone the repository, and you can follow below to set up the environment. We recommend you use a virtual environment like `uv`, but you may prefer to use `conda env` or `pipenv`.
 
 ```bash
-git clone git@github.com:<username>/graphix.git
```
matulni, line 15: I think this was consistent with the recommendation of forking the repo above.

pyproject.toml
```diff
@@ -24,24 +24,66 @@ classifiers = [
...
+dependencies = [
+    "matplotlib",
+    "networkx",
+    "numpy>=2",
+    "opt_einsum",
+    "quimb",
+    "scipy",
+    "sympy",
+    "typing_extensions",
+]
```
matulni, line 36: Should we take the opportunity of this PR to pin all dependencies ?
I believe that this will yield safer and more stable code, not sure if dependabot takes care of updating them...
```suggestion
dependencies = [
    "matplotlib==3.10.8",
    "networkx==3.6.1",
    "numpy==2.4.4",
    "quimb==1.13.0",
    "scipy==1.17.1",
    "sympy==1.14.0",
    "typing_extensions==4.15.0",
]
```

Also, I believe that `opt_einsum` is not used in the codebase (tests pass without), but could someone else double check please?

README.md
```diff
@@ -28,12 +28,23 @@ Install `graphix` with `pip`:
...
 ```bash
 pip install graphix[extra]
 ```
 
+### Editable installation
+
+As well as support for editable installation using pip, Graphix also supports using `uv`:
+
+```bash
+git clone git@github.com:TeamGraphix/graphix.git
```
matulni: I think the HTTPS protocol is better known than SSH

README.md
```diff
@@ -28,12 +28,23 @@ Install `graphix` with `pip`:
...
 ```
 
-Install together with [extra packages](https://github.com/TeamGraphix/graphix/blob/master/requirements-extra.txt):
+Install with [extra dependencies](https://github.com/TeamGraphix/graphix/blob/master/pyproject.toml) (e.g. pyzx):
 
 ```bash
 pip install graphix[extra]
 ```
 
+### Editable installation
```
matulni: Is this heading correct? Shouln't be `uv` instead of editable ?

.github/workflows/ci.yml
```diff
@@ -25,16 +25,17 @@ jobs:
     steps:
       - uses: actions/checkout@v5
         with:
-          fetch-depth: 0  # Fetch all, necessary to find tags and branches
+          fetch-depth: 0 # Fetch all, necessary to find tags and branches
           fetch-tags: true
 
-      - uses: actions/setup-python@v5
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v5
```
matulni: I believe [latest version of the Gtihub action is v8.0](https://github.com/marketplace/actions/astral-sh-setup-uv)

```suggestion
        uses: astral-sh/setup-uv@v8
```

.github/workflows/ci.yml
```diff
@@ -25,16 +25,17 @@ jobs:
...
         with:
-          python-version: ${{ matrix.python }}
+          enable-cache: true
+          cache-dependency-glob: "pyproject.toml"
```
matulni: [It's suggested](https://docs.astral.sh/uv/guides/integration/github/#installation) to pin the uv version as well

```suggestion
          version: "0.11.3" # Install a specific version of uv.
          enable-cache: true
          cache-dependency-glob: "pyproject.toml"
```

.github/workflows/ci.yml
```diff
@@ -25,16 +25,17 @@ jobs:
...
 
-      - run: python -m pip install --upgrade pip
+      - name: "Set up Python ${{ matrix.python }}"
+        run: uv python install ${{ matrix.python }}
```
matulni: Is there a difference between this and setting 
```
with:
          python-version: ${{ matrix.python }}
```
in the previous action as shown in the [docs](https://docs.astral.sh/uv/guides/integration/github/#multiple-python-versions) ?

noxfile.py
```diff
@@ -15,11 +15,13 @@
 if TYPE_CHECKING:
     from collections.abc import Callable
 
+nox.options.default_venv_backend = "uv"
```
matulni: Let's keep the default as a fallback

```suggestion
nox.options.default_venv_backend = "uv|virtualenv"
```

.github/workflows/ci.yml
```diff
@@ -25,16 +25,17 @@ jobs:
...
 
-      - name: Setup nox
-        run: pip install -c requirements-dev.txt nox
-
-      - run: nox --python ${{ matrix.python }}
+      - name: Run nox
+        run: uvx nox -db uv --python ${{ matrix.python }}
```
matulni: Two questions:
1) If I understood correctly [this comment](https://github.com/astral-sh/uv/issues/6579#issuecomment-2465715970), it seems that `nox` installs the appropriate Python version. Does this mean that we can skip the `uv python install ...` above ?
2) Is there a reason why we are doing `uvx nox` instead of `uv run nox` ?

.github/workflows/cov.yml
```diff
@@ -16,22 +15,17 @@ jobs:
...
-          fetch-depth: 0  # Fetch all, necessary to find tags and branches
+          fetch-depth: 0 # Fetch all, necessary to find tags and branches
           fetch-tags: true
 
-      - name: Set up Python
-        uses: actions/setup-python@v5
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v5
         with:
+          enable-cache: true
```
matulni, line 24: See comments in `ci.yml` about versions

matulni (commented):
Thanks for the PR!
Here are some comments

pyproject.toml
emlynsg: This was kind of the idea of the `lockfile`. We each use the `lockfile` (which itself pins the deps) and dependabot can update that. I think the general recommendations that I've read lean towards defining compatible ranges in the `pyproject` (e.g. `matplotlib>3,<4`) and then pinning the exact version for reproducibility in a shared lockfile that we update periodically. Any thoughts @thierry-martinez ?

.github/workflows/ci.yml
emlynsg: Not really but you're right that it's more readable to use the with construction.

.github/workflows/ci.yml
emlynsg: Yes I think I had run into problems with using `uv` instead of `uvx`, but that's been fixed by the current dependency layout so I have returned it back to `uv run `as you say.

emlynsg:
Thanks for comments @matulni , @thierry-martinez , ready for review again.

.github/workflows/ci.yml
```diff
@@ -25,16 +25,15 @@ jobs:
...
-          fetch-depth: 0  # Fetch all, necessary to find tags and branches
+          fetch-depth: 0 # Fetch all, necessary to find tags and branches
           fetch-tags: true
 
-      - uses: actions/setup-python@v5
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v8.0.0
         with:
-          python-version: ${{ matrix.python }}
+          version: "0.11.3" # Install a specific version of uv.
```
thierry-martinez: The difficulty here is that this field is not tracked by Dependabot.

What I would do is to
- add a new file `.github/uv-requirements.txt`:
```
uv==0.11.3
```

- instruct dependabot to track this file in `.github/dependabot.yml`:
```yaml
  - package-ecosystem: "pip"
    directory: "/.github"
    schedule:
      interval: "weekly"
``` 

- define a local composite action that installs the required `uv` version in `.github/actions/setup-uv/action.yml` (we use `pip` to install the required version, but we reproduce the cache logic from `astral-sh/setup-uv`):
```yaml
name: Set up uv
description: Install the pinned version of uv and cache its package cache

runs:
  using: composite
  steps:
    - name: Install uv
      shell: bash
      run: pip install -r .github/uv-requirements.txt

    - name: Cache uv packages
      uses: actions/cache@v4
      with:
        path: ~/.cache/uv
        key: uv-${{ runner.os }}-${{ hashFiles('uv.lock') }}
        restore-keys: |
          uv-${{ runner.os }}-
```
- use it that way in `ci.yml` and other pipelines:
```yaml
  - uses: ./.github/actions/setup-uv
```

Alternatively, we could just leave the field blank to get the default behavior: _searches for version in config files, then latest_ (documented [here](https://github.com/astral-sh/setup-uv?tab=readme-ov-file#inputs)), and specify the version in `pyproject.toml` (documented [here](https://docs.astral.sh/uv/reference/settings/#required-version)):
```toml
[tool.uv]
required-version = "0.11.3"
```
But that would force users to use that version locally, and I think that is something we would prefer to avoid (nor is it tracked by Dependabot).

pyproject.toml
thierry-martinez: Ok, perhaps I was wrong to suggest not to version `uv.lock`. I was afraid that adding this file would force the users to have these exact versions of the dependencies installed in their environments, and that there would be a lot of merge conflicts, but:

- `uv.lock` will only be considered when we use `uv` to make development in the `graphix` repo; when we use `graphix` as a dependency in another project, then only the constraints in `pyproject.toml` will be considered.

- I believe merge conflicts can handled more cleanly with these additions in `.gitattributes` (stolen from `pixi.lock`):
```
# SCM syntax highlighting & preventing 3-way merges
uv.lock merge=binary linguist-language=TOML linguist-generated=true -diff
```

Additionally, we would like Dependabot to keep track of this lock file too by adding this entry to `.github/dependabot.yml`:
```
- package-ecosystem: "uv"
  directory: "/"
  schedule:
    interval: "weekly"
```

thierry-martinez (approved):
LGTM! Thank for this big change!

matulni (approved):
Great, thanks!

PR #455, opened 2026-03-02
Fix #449: rewrite causal-flow algorithm iteratively.

### LINKED ISSUE #449: [Bug]: `RecursionError` in causal-flow finding algorithm
Causal-flow finding algorithm is implemented in a recursive fashion following the presentation in the original publication (see references in the source code). Linear open graphs  with $\sim$ 1000 nodes raise a `RecursionError` exception on a MacBook Air M4, 24 GB running OS 15.7.3.

The following code snippet allows to reproduce the issue:

```python
import networkx as nx


def create_line_og(n_nodes: int) -> OpenGraph[Plane]:
    if n_nodes < 2:
        raise ValueError("n_nodes must be larger than 1.")

    graph: nx.Graph[int] = nx.Graph([(i, i + 1) for i in range(n_nodes - 1)])
    input_nodes = [0]
    output_nodes = [n_nodes - 1]
    measurements = dict.fromkeys(range(n_nodes - 1), Plane.XY)
    return OpenGraph(graph, input_nodes, output_nodes, measurements)


n_nodes = 990
while True:
    og = create_line_og(n_nodes)
    try:
        og.extract_causal_flow()
    except RecursionError:
        break
    n_nodes += 1

print(f"Recursion limit hit at {n_nodes}")
```

Note that a flow correction function can be found for these graphs with the cubic Pauli-flow finding algorithm.

Since the recursion in the causal-flow finding algorithm is terminal, it should be possible to rewrite the algorithm in an iterative fashion without needing to represent the stack in a custom data structure.

matulni:
This commit replaces the recursion call in the causal flow algorithm by a `while` loop. This rewriting prevents the `RecursionError` reported in #449 for linear graphs with more than 1000 nodes.

## Example

```python
import timeit

import networkx as nx


def create_line_og(n_nodes: int) -> OpenGraph[Plane]:
    if n_nodes < 2:
        raise ValueError("n_nodes must be larger than 1.")

    graph: nx.Graph[int] = nx.Graph([(i, i + 1) for i in range(n_nodes - 1)])
    input_nodes = [0]
    output_nodes = [n_nodes - 1]
    measurements = dict.fromkeys(range(n_nodes - 1), Plane.XY)
    return OpenGraph(graph, input_nodes, output_nodes, measurements)

def compute_flow(n_nodes: int):
    og = create_line_og(n_nodes)
    def _compute_flow():
        og.extract_causal_flow()
    return _compute_flow

n_nodes = 5000
t = timeit.Timer(compute_flow(n_nodes))
print(t.timeit(5))  # 6.928544000002148
```

## TODO
Update CHANGELOG


graphix/flow/_find_cflow.py
```diff
@@ -57,64 +56,32 @@ def find_cflow(og: OpenGraph[_PM_co]) -> CausalFlow[_PM_co] | None:
         frozenset(corrected_nodes)
```
thierry-martinez: Sorry for commenting outside the changes of the PR, but would it not be clearer with `frozenset(og.output_nodes)` then? The comment could then be removed.

graphix/flow/_find_cflow.py
```diff
@@ -57,64 +56,32 @@ def find_cflow(og: OpenGraph[_PM_co]) -> CausalFlow[_PM_co] | None:
...
-    -------
-    CausalFlow[_PM_co] | None
-        A causal flow object if the open graph has causal flow, `None` otherwise.
-    """
-    corrected_nodes_new: set[int] = set()
-    corrector_nodes_new: set[int] = set()
-    curr_layer: set[int] = set()
+    corrected_nodes_new: set[int]
+    corrector_nodes_new: set[int]
+    curr_layer: set[int]
```
thierry-martinez: These variables are local to the body of the loop, so I would inline their type annotation at their initialization inside the loop.

tests/test_opengraph.py
```diff
@@ -863,6 +863,31 @@ def test_pflow(self, test_case: OpenGraphFlowTestCase, fx_rng: Generator) -> Non
...
+
+        Open graph structure:
+
+        [0]-1-...-1498-(1499)
+
+        Notes
+        -----
+        This linear graph will trigger a `RecursionError` in the recursive implementation of the causal-flow finding algorithm.
+        """
+        n_nodes = 1200
```
thierry-martinez: We could use `sys.getrecursionlimit()` instead, which makes the test independent of the environment.

```suggestion
        n_nodes = sys.getrecursionlimit()
```

tests/test_opengraph.py
```diff
@@ -863,6 +863,31 @@ def test_pflow(self, test_case: OpenGraphFlowTestCase, fx_rng: Generator) -> Non
...
+        og = OpenGraph(
+            graph=nx.Graph([(i, i + 1) for i in range(n_nodes - 1)]),
+            input_nodes=[0],
+            output_nodes=[n_nodes - 1],
+            measurements=dict.fromkeys(range(n_nodes - 1), Measurement.XY(0.2)),
+        )
+        c_ref = {i: frozenset({i + 1}) for i in og.measurements}
+        pol_ref = tuple(frozenset({i}) for i in range(n_nodes - 1, -1, -1))
```
thierry-martinez:
```suggestion
        pol_ref = tuple(frozenset({i}) for i in reversed(range(n_nodes)))
```

tests/test_opengraph.py
```diff
@@ -863,6 +863,31 @@ def test_pflow(self, test_case: OpenGraphFlowTestCase, fx_rng: Generator) -> Non
...
+
+        [0]-1-...-1498-(1499)
+
+        Notes
+        -----
+        This linear graph will trigger a `RecursionError` in the recursive implementation of the causal-flow finding algorithm.
+        """
+        n_nodes = 1200
+        og = OpenGraph(
+            graph=nx.Graph([(i, i + 1) for i in range(n_nodes - 1)]),
```
thierry-martinez:
```suggestion
            graph=nx.path_graph(n_nodes),
```

emlynsg (approved):
Agree with Thierry's comments. Nice fix and clean tests!

tests/test_opengraph.py
matulni: Nice, I didn't know about that. Done in 858dd22577c85bd4943029df65f6e16901f4cb0a

tests/test_opengraph.py
matulni: Done in 858dd22577c85bd4943029df65f6e16901f4cb0a

tests/test_opengraph.py
matulni: Done in 858dd22577c85bd4943029df65f6e16901f4cb0a

graphix/flow/_find_cflow.py
matulni: Done in 61921e2ba7f18ed1b85c63f0e1172d6736d754dd

graphix/flow/_find_cflow.py
matulni: I changed the variable name in 61921e2ba7f18ed1b85c63f0e1172d6736d754dd.
I still see value in the comment, though: `XZCorrections` objects can exist for patterns with no outputs, so we have to run a bunch of sanity checks to avoid having an empty first layer. This is never necessary for flow objects, at the time of writing that code I thought it was a good reminder :)

graphix/flow/_find_cflow.py
```diff
@@ -47,74 +45,39 @@ def find_cflow(og: OpenGraph[_PM_co]) -> CausalFlow[_PM_co] | None:
...
     corrected_nodes = set(og.output_nodes)
     corrector_candidates = corrected_nodes - set(og.input_nodes)
     non_input_nodes = og.graph.nodes - set(og.input_nodes)
 
     cf: dict[int, frozenset[int]] = {}
     # Output nodes are always in layer 0. If the open graph has flow, it must have outputs, so we never end up with an empty set at `layers[0]`.
     layers: list[frozenset[int]] = [
-        frozenset(corrected_nodes)
+        frozenset(og.output_nodes)
     ]  # A copy is necessary because `corrected_nodes` is mutable and changes during the algorithm.
```
thierry-martinez: This comment is no longer relevant. The whole expression can now fit on the same line.

matulni: Ahh sorry, I misunderstood which comment you referred to. Done in 115d8008b81673c96875d005601ea372fddadcd2

Issue #100, opened 2023-11-15
Add methods to `Statevec`


king-p3nguin:
*Edit (2025/12/08)*: First part of this issue subsumed in #386 for UnitaryDesign2026
*Edit (2026/03/11)*: `.to_dict` method introduced in #457 

The following methods may improve the utility of `Statevec`:
- `.draw()`: show state vector in latex format (in jupyter notebook) or text (in command line)
  - add `format` to parameter (`"latex"` or `"text"`)

masa10-f:
@king-p3nguin  
Hi! I have a question about one of the methods you show.
Would you explain the more detail of `.to_dict()`? I don't grasp the overview of this method.

king-p3nguin:
@masa10-f 
I was thinking of what Qiskit does in its method:
https://qiskit.org/documentation/stubs/qiskit.quantum_info.Statevector.html#qiskit.quantum_info.Statevector.to_dict

The only difference is that graphix adopts bit-endian, whereas Qiskit uses little-endian.

thierry-martinez:
@matulni is working on `to_dict`.

emlynsg:
@thierry-martinez Added to the initial post extra details on what we expect from a PR that resolves this issue.

matulni:
Subsumed by #501

PR #446, opened 2026-02-18
Introduction of informational `pydoclint` CI job on push and PR

emlynsg:
**Draft PR for presentation and discussion**

This draft PR:

- introduces a new job in the CI that runs `pydoclint`, and reports on failing lines in each file of the repository
- adds to the pre-commit hook to include Ruff and a commented `pydoclint` step (to be reactivated when we have completed standardising the docstrings of the repository)
- fixes docstrings/type annotations that were causing `pydoclint` to fail when run.

*Note: The graphix-qasm-parser dependency was causing the CI to fail, and was commented out.*

Example output:

## pydoclint report

<details><summary><code>graphix/_linalg.py</code> — 11 violation(s)</summary>

```
    18: DOC102: Method `MatGF2.__new__`: Docstring contains more arguments than in function signature.
    18: DOC110: Method `MatGF2.__new__`: The option `--arg-type-hints-in-docstring` is `True` but not all args in the docstring arg list have type hints
    18: DOC103: Method `MatGF2.__new__`: Docstring arguments are different from function arguments. (Or could be other formatting issues: https://jsh9.github.io/pydoclint/violation_codes.html#notes-on-doc103 ). Arguments in the docstring but not in the function signature: [-------: , MatGF2: , Return: ].
    18: DOC201: Method `MatGF2.__new__` does not have a return section in docstring
    18: DOC203: Method `MatGF2.__new__` return type(s) in docstring not consistent with the return annotation. Return annotation has 1 type(s); docstring return section has 0 type(s).
    37: DOC501: Method `MatGF2.mat_mul` has raise statements, but the docstring does not have a "Raises" section
    37: DOC503: Method `MatGF2.mat_mul` exceptions in the "Raises" section in the docstring do not match those in the function body. Raised exceptions in the docstring: []. Raised exceptions in the body: ['ValueError'].
    66: DOC203: Method `MatGF2.compute_rank` return type(s) in docstring not consistent with the return annotation. Return annotation types: ['np.intp']; docstring return section types: ['int']
    131: DOC103: Method `MatGF2.gauss_elimination`: Docstring arguments are different from function arguments. (Or could be other formatting issues: https://jsh9.github.io/pydoclint/violation_codes.html#notes-on-doc103 ). Arguments in the function signature but not in the docstring: [ncols: int | None]. Arguments in the docstring but not in the function signature: [n_cols: int ].
[...]

PR #457, opened 2026-03-05
Add `Statevec.to_dict` and `Statevec.to_prob_dict` methods

matulni:
This PR adds `Statevec.to_dict` and `Statevec.to_prob_dict` methods to convert a statevector in dictionary form as suggested in #100. It allows to choose between big-endian encoding (Graphix default) or little-endian encoding (Qiskit default) in the ket representation for easier cross-platform comparison.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - #458: The type for the `input_state` parameter in the simulator now allows `None` to indicate that the input qubits have already been specified in the backend.
 
+- #465: Fix #464. Functions in `graphix._linalg.py` convert arrays to c-contiguous form before passing them to numba-jitted functions.
+
 ### Changed
 
 - #181, #423: Structural separation of Pauli measurements
--- a/graphix/_linalg.py
+++ b/graphix/_linalg.py
@@ -31,7 +31,9 @@ class MatGF2(npt.NDArray[np.uint8]):
             MatGF2
         """
         arr = np.array(data, dtype=np.uint8, copy=copy)
-        return super().__new__(cls, shape=arr.shape, dtype=arr.dtype, buffer=arr)
+        return super().__new__(
+            cls, shape=arr.shape, dtype=arr.dtype, buffer=arr, strides=arr.strides
+        )  # `strides=arr.strides` allows to preserve the memory layout of the original data.
 
     def mat_mul(self, other: MatGF2 | npt.NDArray[np.uint8]) -> MatGF2:
         r"""Multiply two matrices.
@@ -60,7 +62,7 @@ class MatGF2(npt.NDArray[np.uint8]):
                 f"Dimension mismatch. Attempted to multiply `self` with shape {self.shape} and  `other` with shape {other.shape}"
             )
 
-        return MatGF2(_mat_mul_jit(self, other), copy=False)
+        return MatGF2(_mat_mul_jit(np.ascontiguousarray(self), np.ascontiguousarray(other)), copy=False)
 
     def compute_rank(self) -> np.intp:
         """Get the rank of the matrix.
@@ -149,7 +151,7 @@ class MatGF2(npt.NDArray[np.uint8]):
         ncols_value = self.shape[1] if ncols is None else ncols
         mat_ref = MatGF2(self) if copy else self
 
-        return MatGF2(_elimination_jit(mat_ref, ncols=ncols_value, full_reduce=False), copy=False)
+        return MatGF2(_elimination_jit(np.ascontiguousarray(mat_ref), ncols=ncols_value, full_reduce=False), copy=False)
 
     def row_reduction(self, ncols: int | None = None, copy: bool = True) -> MatGF2:
         """Return row-reduced echelon form (RREF) by performing Gaussian elimination.
@@ -170,7 +172,7 @@ class MatGF2(npt.NDArray[np.uint8]):
         ncols_value = self.shape[1] if ncols is None else ncols
         mat_ref = self.copy() if copy else self
 
-        return MatGF2(_elimination_jit(mat_ref, ncols=ncols_value, full_reduce=True), copy=False)
+        return MatGF2(_elimination_jit(np.ascontiguousarray(mat_ref), ncols=ncols_value, full_reduce=True), copy=False)
 
 
 def solve_f2_linear_system(mat: MatGF2, b: MatGF2) -> MatGF2:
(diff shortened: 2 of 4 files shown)


graphix/sim/statevec.py
```diff
@@ -25,6 +25,7 @@
     from graphix.parameter import ExpressionOrFloat, ExpressionOrSupportsFloat, Parameter
     from graphix.sim.data import Data
 
+_ENCODING = Literal["LSB", "MSB"]
```
thierry-martinez: **nit**: Since `_ENCODING` appears to be used only for type checking, you could define it within the `TYPE_CHECKING` block. This would also allow moving the `typing.Literal` import there.

graphix/sim/statevec.py
```diff
@@ -427,6 +428,110 @@ def isclose(self, other: Statevec, *, rtol: float = 1e-09, atol: float = 0.0) ->
...
+        abs_tol: float = 1e-8,
+    ) -> dict[str, complex]:
+        r"""Convert the statevector to dictionary form.
+
+        This dictionary representation uses a ket-like notation where the dictionary ``keys`` are qubit strings
+        for the basis vectors and ``values`` are the corresponding complex amplitudes. Amplitudes below a certain threshold are filtered out.
+
+        Parameters
+        ----------
+        encoding : _ENCODING, default="MSB"
```
thierry-martinez: Private constants like `_ENCODING` should not be referenced in docstrings; inline the definition.

graphix/sim/statevec.py
```diff
@@ -427,6 +428,110 @@ def isclose(self, other: Statevec, *, rtol: float = 1e-09, atol: float = 0.0) ->
...
+        -----
+        The encoding determines the bit ordering convention used when mapping basis states to dictionary
+        keys. Consider a tensor product of three qubits:
+
+        .. math::
+
+        \lvert\psi\rangle = q_0 \otimes q_1 \otimes q_2.
+
+        If ``encoding == "MSB"`` the first qubit is represented in the Most Significant Bit -> ``q0q1q2``. This is the default representation in Graphix.
+        If ``encoding == "LSB"`` the first qubit is represented in the Least Significant Bit -> ``q2q1q1``. This is the default representation in other software packages such as Qiskit.
```
emlynsg: ` ``q2q1q0`` `

emlynsg (approved):
LGTM!

graphix/sim/statevec.py
```diff
@@ -427,6 +428,110 @@ def isclose(self, other: Statevec, *, rtol: float = 1e-09, atol: float = 0.0) ->
...
+        -------
+        dict[str, float]
+            The probability distribution associated to the statevector in dictionary form.
+
+        See Also
+        --------
+        .. :meth:`to_dict`
+        """
+        return {
+            key: float(abs(amp) ** 2) for key, amp in self.to_dict(encoding, rel_tol=rel_tol, abs_tol=abs_tol).items()
```
thierry-martinez: This cast to `float` can be removed, as `abs(amp)` is already a `float` for `complex` inputs.

graphix/sim/statevec.py
matulni: I agree, I'll document as:
``encoding : Literal["LSB", "MSB"], default="MSB"``

Note however that running `pydoctlint` yields
``433: DOC105: Method `Statevec.to_dict`: Argument names match, but type hints in these args do not match: encoding, rel_tol, abs_tol``

I guess we can maybe opt-out that rule when we implement #446 @emlynsg

graphix/sim/statevec.py
```diff
@@ -427,6 +430,101 @@ def isclose(self, other: Statevec, *, rtol: float = 1e-09, atol: float = 0.0) ->
...
+        >>> sv.to_dict()
+        {'01': np.complex128(1+0j)}
+        >>> sv.to_dict(encoding="LSB")
+        {'10': np.complex128(1+0j)}
+        """
+        mask = np.logical_not(np.isclose(np.abs(self.flatten()), 0, rtol=rtol, atol=atol))
+        i_vals = np.arange(1 << self.nqubit)[mask]
+        amp_vals = self.flatten()[mask]
+
+        return {_format_encoding(self.nqubit, i, encoding): amp for i, amp in zip(i_vals, amp_vals, strict=True)}
```
thierry-martinez: We could reduce the code duplication by defining a higher-order function, as you tried at first, but keeping it internal and calling it with the identity function for `to_dict` and with the squared norm for `to_prob_dict`, to avoid the difficulty with the typing of the default parameter that you encountered.


```python
if TYPE_CHECKING:
    _ScalarT = TypeVar("_ScalarT", bound=np.generic[Any])
[...]
    def _to_dict_map(
        self,
        f: Callable[[npt.NDArray[np.object_ | np.complex128]], npt.NDArray[_ScalarT]],
        encoding: _ENCODING = "MSB",
        *,
        rtol: float = 0.0,
        atol: float = 1e-8,
    ) -> dict[str, _ScalarT]:
        mask = np.logical_not(np.isclose(np.abs(self.flatten()), 0, rtol=rtol, atol=atol))
        i_vals = np.arange(1 << self.nqubit)[mask]
        amp_vals = f(self.flatten()[mask])

        return {_format_encoding(self.nqubit, i, encoding): amp for i, amp in zip(i_vals, amp_vals, strict=True)}
```

and then the body of `to_dict` becomes:

```suggestion
        return self._to_dict_map(lambda x: x, encoding, rtol=rtol, atol=atol)
```

graphix/sim/statevec.py
```diff
@@ -427,6 +430,101 @@ def isclose(self, other: Statevec, *, rtol: float = 1e-09, atol: float = 0.0) ->
...
+
+        See Also
+        --------
+        .. :meth:`to_dict`
+        """
+        mask = np.logical_not(np.isclose(np.abs(self.flatten()), 0, rtol=rtol, atol=atol))
+        i_vals = np.arange(1 << self.nqubit)[mask]
+        amp2_vals = np.abs(self.flatten()[mask]) ** 2
+
+        return {_format_encoding(self.nqubit, i, encoding): amp2 for i, amp2 in zip(i_vals, amp2_vals, strict=True)}
```
thierry-martinez: With `_to_dict_map`, the body becomes:
```suggestion
        return self._to_dict_map(lambda x: np.abs(x) ** 2, encoding, rtol=rtol, atol=atol)
```

graphix/sim/statevec.py
```diff
@@ -427,6 +430,101 @@ def isclose(self, other: Statevec, *, rtol: float = 1e-09, atol: float = 0.0) ->
         """
         return math.isclose(self.fidelity(other), 1, rel_tol=rtol, abs_tol=atol)
 
+    def to_dict(
+        self,
+        encoding: _ENCODING = "MSB",
+        *,
+        rtol: float = 0.0,
+        atol: float = 1e-8,
+    ) -> dict[str, complex]:
```
thierry-martinez: Note that this return type is incorrect in the presence of symbolic parameters. Static type checkers don't catch this because they type the dict comprehension as `dict[str, Any]` (as we can see with `reveal_type`). However, with the proposed refactor, we obtain a more precise type, and the return type should be `dict[str, np.object_ | np.complex128]`.

tests/test_statevec.py
```diff
@@ -207,6 +207,28 @@ def test_isclose_tolerance(self) -> None:
         assert not zero.isclose(almost)
         assert zero.isclose(almost, atol=1e-6)
 
+    def test_to_dict(self) -> None:
```
thierry-martinez: Wouldn't it be more readable with `@pytest.mark.parametrize` using tuples?

matulni: Indeed, done in [43a538f](https://github.com/TeamGraphix/graphix/pull/457/commits/43a538f95539508406360d5265089bcf0fda8d7e)

thierry-martinez: Can't we parametrize over `sv` too?

PR #458, opened 2026-03-05
Allow `None` in simulator's `input_state` type

thierry-martinez:
The `run` method handles `None` and doesn't add input nodes to the backend in this case, assuming that the backend has already been prepared by the caller. However, the type annotation prevented using `None` as an `input_state` value.

```python
        if input_state is not None:
            self.backend.add_nodes(self.pattern.input_nodes, input_state)
```

**Update [2026-03-06T13:16:30]:** Allow `None` in `Pattern.simulate_pattern` and `Matrix` in `Statevec.isclose`.

### DIFF
--- a/tests/test_statevec.py
+++ b/tests/test_statevec.py
@@ -193,15 +193,17 @@ class TestFidelityIsclose:
 
     def test_isclose_orthogonal(self) -> None:
         zero = Statevec(data=BasicStates.ZERO)
-        assert not zero.isclose(BasicStates.ONE.to_statevector())
+        one = Statevec(data=BasicStates.ONE)
+        assert not zero.isclose(one)
 
     def test_isclose_global_phase(self) -> None:
         plus = Statevec(data=BasicStates.PLUS)
-        assert plus.isclose(np.array([1, 1]) / np.sqrt(2) * np.exp(1j * 0.7))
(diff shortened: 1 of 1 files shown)


graphix/sim/statevec.py
```diff
@@ -388,14 +388,14 @@ def expectation_value(self, op: Matrix, qargs: Sequence[int]) -> complex:
...
-    def fidelity(self, other: Statevec) -> float:
+    def fidelity(self, other: Matrix | Statevec) -> float:
         r"""Calculate the fidelity against another statevector.
 
         The fidelity is defined as :math:`|\langle\psi_1|\psi_2\rangle|^2`.
 
         Parameters
         ----------
-        other : :class:`Statevec`
+        other : :class:`Matrix` | :class:`Statevec`
```
matulni: Could it be a problem that with the enlarged type of `other` it becomes easier to mistakenly compare two vectors in different basis ?
Before this was to some extent guaranteed by the type-checker (in the sense that we could only compare `Statevec` objects), and the responsibility of having an `np.array` in the correct basis was delegated to the initialisation step.

But maybe I'm being overly cautious...

matulni (commented):
Thanks for the PR, just one small comment

graphix/sim/statevec.py
emlynsg: I don't have a strong opinion either way, but worth noting Matrix input is similarly not supported in DensityMatrix fidelity either.

PR #460, opened 2026-03-07
Allow printing `limit` to be `None`, remove deprecated `print_pattern`

thierry-martinez:
This commit:
- updates `Pattern` methods (`to_ascii`, `to_unicode`, `to_latex`) and `pattern_to_str` to allow `limit` to be `None`.
- removes the deprecated `print_pattern` function.

### DIFF
--- a/graphix/flow/_find_gpflow.py
+++ b/graphix/flow/_find_gpflow.py
@@ -39,6 +39,13 @@ class AlgebraicOpenGraph(Generic[_AM_co]):
 
     It reuses the class `:class: graphix.sim.base_backend.NodeIndex` introduced for managing the mapping between node numbers and qubit indices in the internal state of the backend.
 
+    Attributes
+    ----------
+        og (OpenGraph)
+        non_inputs (NodeIndex) : Mapping between matrix indices and non-input nodes (labelled with integers).
+        non_outputs (NodeIndex) : Mapping between matrix indices and non-output nodes (labelled with integers).
+        non_outputs_optim (NodeIndex) : Mapping between matrix indices and a subset of non-output nodes (labelled with integers).
+
     Notes
     -----
     At initialization, `non_outputs_optim` is a copy of `non_outputs`. The nodes corresponding to zero-rows of the order-demand matrix are removed for calculating the :math:`P` matrix more efficiently in the `:func: _compute_correction_matrix_general` routine.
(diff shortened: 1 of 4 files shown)


graphix/pretty_print.py
```diff
@@ -253,23 +253,29 @@ def pattern_to_str(
...
     separator = r"\," if output == OutputFormat.LaTeX else " "
     command_list = list(pattern)
     if target is not None:
         command_list = [command for command in command_list if command.kind in target]
     if not left_to_right:
         command_list.reverse()
-    truncated = len(command_list) > limit
-    short_command_list = command_list[: limit - 1] if truncated else command_list
+    truncated = limit is not None and len(command_list) > limit
+    short_command_list = command_list[: limit - 1] if limit is not None and truncated else command_list
```
matulni, line 276:
```suggestion
    short_command_list = command_list[: limit - 1] if truncated else command_list
```
Here `truncated` is enough, 

$$A \land (A \land B) = A \land B,$$

right?

graphix/pretty_print.py
```diff
@@ -253,23 +253,29 @@ def pattern_to_str(
...
     result = separator.join(command_to_str(command, output) for command in short_command_list)
     if output == OutputFormat.LaTeX:
         result = f"\\({result}\\)"
-    if truncated:
+    if limit is not None and truncated:
```
matulni, line 280:
```suggestion
    if truncated:
```

graphix/pattern.py
```diff
@@ -374,42 +374,64 @@ def __eq__(self, other: object) -> bool:
...
+        left_to_right: bool, optional
+            If ``True``, the first command will appear at the beginning of
+            the resulting string. If ``False`` (the default), the first command will
+            appear at the end of the string.
+        limit: int | None, optional
+            If set to an int (default: 40), only first ``limit`` commands are printed,
+            and an ellipsis is added at the end to indicate that some commands have been elided.
+            If ``limit=None``, there is no limit on the number of printed commands.
+        target: Container[command.CommandKind], optional
+            If set, only commands of kinds specified in ``target`` are printed.
```
emlynsg, line 392: To make it searchable when the docs are updated, I think it would be good to explicitly mention here something like `For example, you can print only measurements with ``command.M`` or X corrections with ``command.X`` `.

graphix/pretty_print.py
thierry-martinez: You're right: logically the test is redundant because `truncated` already implies `limit is not None`. However, mypy does not track implications across boolean variables, so it won't narrow the type of `limit` without the explicit check. I added the following comment in commit 9249f0e:
```python
    # Note: The redundant test `limit is not None` is required for mypy
    # to narrow the type of `limit` in the then-branch.
```

matulni: I see thanks, for the clarification and adding the comment

graphix/pattern.py
thierry-martinez: I added runnable (and pytest-checked) examples.

Issue #235, opened 2024-11-19
[Bug]: `to_pyzx_graph`/`from_pyzx_graph` don't preserve the semantics when calling `pyzx.simplify.full_reduce`.


thierry-martinez:
The following code should output fidelity 1, but outputs fidelity 0.5. The code takes a circuit with a Toffoli gate, transpiles it, converts the pattern to a PyZX graph, performs `full_reduce` and then converts the resulting graph back to a pattern: simulating the two patterns don't give equivalent state vectors.

```
from graphix import Circuit
from graphix.opengraph import OpenGraph
import pyzx
from graphix.pyzx import to_pyzx_graph, from_pyzx_graph
import numpy as np
from copy import deepcopy

c = Circuit(3)
c.ccx(0, 1, 2)
p = c.transpile(opt=False).pattern
og = OpenGraph.from_pattern(p)
pyg = to_pyzx_graph(og)
pyg.normalize()
pyg_copy = deepcopy(pyg)
pyzx.simplify.full_reduce(pyg)
pyg.normalize()
t = pyzx.tensorfy(pyg)
t2 = pyzx.tensorfy(pyg_copy)
assert pyzx.compare_tensors(t, t2)
og2 = from_pyzx_graph(pyg)
p2 = og2.to_pattern()
s = p.simulate_pattern()
s2 = p2.simulate_pattern()
print(np.abs(np.dot(s.flatten().conj(), s2.flatten())))
```

PR #461, opened 2026-03-08
Add warning when default random number generator is used

thierry-martinez:
This commit adds a warning when `ensure_rng` is called with `None`.

It includes the following fixes:

- In `DepolarisingNoiseModel`, the field `rng` has been removed: the current API favors passing `rng` as a method parameter instead of storing it inside the structure.

- `rand_dm` and `rand_channel_kraus` now use the `rng` passed as a parameter for the entire computation, instead of relying partially on the default random number generator.

- Since #235 appears not to have been properly fixed, `test_full_reduce_toffoli` is marked as `xfail`.

This PR subsumes #459.

### DIFF
--- a/graphix/rng.py
+++ b/graphix/rng.py
@@ -15,13 +15,13 @@ _rng_local = threading.local()
 
 
 def ensure_rng(rng: Generator | None = None, *, stacklevel: int = 1) -> Generator:
-    """Return a default random-number generator if `None` is given.
+    """Return a default random-number generator if ``None`` is given.
 
     Parameters
     ----------
     rng: Generator | None, optional
-        If set and not `None`, this value is returned.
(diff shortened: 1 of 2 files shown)


graphix/rng.py
```diff
@@ -13,10 +14,25 @@
...
+        Stack level to use for warnings. Defaults to 1, meaning that warnings
+        are reported at this function's call site.
+
+    Notes
+    -----
+    A warning is issued if the default random-number generator is used.
+    """
     if rng is not None:
         return rng
+    warnings.warn("Default random-number generator is used. Results are not reproducible.", stacklevel=stacklevel + 1)
```
emlynsg: Would `may not be reproducible` be a better characterisation?

PR #476, opened 2026-04-02
Add native Clifford map compilation pass 

### LINKED ISSUE #104: Unitary from pattern

Hey all, I'm interested if there is any (efficient) way to get the unitary implemented from a given pattern. I did not find any direct function but I think this could be done easily if one can set the input state arbitrarily (issue #53) and check for flow of a pattern.  

**Additional context**

I would propose a (pseudo) code in that direction:

```python
def get_unitary_from_pattern(pattern):
		
	# check if given pattern has flow, i.e. it implements indeed a unitary
	assert has_flow(pattern)

	# num_qubits .. defined via input/ouput nodes
	assert pattern.num_input_nodes == pattern.num_output_nodes
	
	num_qubits = pattern.num_input_nodes
	N          = 2**num_qubits

	#----- get unitary -----
	# do <i|U|j>=U_ij, where |n>=(0, .., 1, .., 0) std-basis
	# maybe the lib (tensornetwork) allows a more efficient way?

	U = np.zeros((N, N), dtype=np.complex128)
	
	for j in range(N): #can be async

		# set |j> input
		state_j    = np.zeros((N), dtype=np.complex128) 
		state_j[j] = 1
		pattern.set_input(state_j)

		# get matrix elements
		tn        = pattern.simulate_pattern(backend='tensornetwork')
		state_U_j = tn.to_statevector()

		U[:, j] = state_U_j

		#same as
		#for i in range(N): U[i, j] = state_U_j[i]

	return U
```

Maybe there is a better way to do this? Maybe even with the current code base?

matulni:
This PR adds a new compilation pass to natively synthesise (unitary) Clifford maps. It subsumes the `stim` compilation pass recently introduced  in [`graphix-stim-compiler`](https://github.com/qat-inria/graphix-stim-compiler) so that it is possible to do the `Pattern` -> `Circuit` conversion without external dependencies:

```python

qc = Circuit(2)
qc.cnot(1, 0)
qc.rz(0, 0.2)
qc.h(0)
print(qc)  # Circuit(width=2, instr=[CNOT(0, 1), RZ(0, pi/5), H(0)])

pattern = qc.transpile().pattern
qc_extracted = pattern.extract_opengraph().extract_circuit()
print(qc_extracted)  # Circuit(width=2, instr=[H(1), H(0), CNOT(1, 0), H(1), H(0), H(0), RX(0, pi/5)])

s_ref = qc.simulate_statevector().statevec
s_extracted = qc_extracted.simulate_statevector().statevec
assert s_ref.isclose(s_extracted)
```

### Summary of changes

- Introduced new methods: `OpenGraph.extract_circuit`, `CliffordMap.to_tableau` and new function `graphix.circ_ext.compilation.cm_berg_pass`.

Further, some improvements in the existing circuit extraction API are made:
-  Added new field `dim` to `PauliString` to represent the dimension of the Hilbert space.
- Define `PauliString` on qubit indices and remove all `remap` methods.
- Represent `x_map` and `z_map` attributes of `CliffordMap` as sequences of `PauliString` instead of mappings.

### Discussion

This PR implements a custom adaptation of van den Berg's sweeping algorithm introduced in Ref. [1]. We note that in this paper their objective is to sample random Clifford operators, so here have to be a bit more careful to obtain the correct Clifford operator. Namely, we have to:

- compute the adjoint of the compiled Clifford,
- properly account for the sign phases
- perform column operation on the _full_ tableau instead of two-rows subtableaus.

The last point is crucial since the (naive) implementation in this PR may increase the reported $O(n^2)$ complexity. I list below some open questions beyond the scope of this PR:

- What is the complexity scaling of this implementation and how does it compare to Stim's version ? (This should be easy to do, since the stim pass in [`graphix-stim-compiler`](https://github.com/qat-inria/graphix-stim-compiler) allows to make sensible comparisons between both methods without much glue code).

- Do we find an $O(n^2)$ complexity (which to my knowledge is the best-known bound)? If not, it may be worth exploring the algorithms in Ref. [2].

- Explore Ref. [3] to synthesise Clifford isometries (corresponding to open graphs with larger number of outputs than inputs). This is particularly interesting since stim does not support these objects but it requires some prior work to represent circuit with ancillas in Graphix.

- Explore Ref. [4] to synthesise Clifford unitaries on a given qubit topology. This should not be too complicated, since their algorithm is based on van den Berg's algorithm implemented here.

- Finally, a vague idea: one of the virtues of the Pauli exponential DAG + Clifford map form in Simmon's paper is that it splits the Clifford and the non-Clifford part. Now, our compilation passes stitch both parts together in a single circuit. It would be more interesting to engineer an interface with stim (or some custom simulator), so that we can classically simulate the Clifford part, and somehow represent the output state as a stabilizer state that we pass to the PdDAG part.

### References
[1] Van Den Berg, 2021. A simple method for sampling random Clifford operators (arXiv:2008.06011)
[2] Bravyi and Maslov, 2020, Hadamard-free circuits expose the structure of the Clifford group. (arXiv:2003.09412)
[3] Brugière et al., 2025, A graph-state based synthesis framework for Clifford isometries (arXiv:2212.06928)
[4] Winderl et al., 2024, Architecture-Aware Synthesis of Stabilizer Circuits from Clifford Tableaus (arXiv:2309.08972)


graphix/circ_ext/compilation.py
```diff
@@ -181,3 +188,167 @@ def add_hy(qubit: int, circuit: Circuit) -> None:
...
+    For the mapping between tableau updates and Clifford gates (H, S, CNOT) see [2].
+
+    References
+    ----------
+    [1] Van Den Berg, 2021. A simple method for sampling random Clifford operators (arxiv:2008.06011).
+    [2] Aaronson, Gottesman, (2004). Improved Simulation of Stabilizer Circuits (arXiv:quant-ph/0406196).
+    """
+    tab = clifford_map.to_tableau()
+    n = len(clifford_map.output_nodes)
+    if len(clifford_map.input_nodes) != n:
```
emlynsg, line 261: As discussed in the meeting, does it make more sense to check for list equality rather than matching `len`?

graphix/circ_ext/compilation.py
```diff
@@ -5,14 +5,23 @@
...
 if TYPE_CHECKING:
     from collections.abc import Callable
+    from typing import TypeAlias
 
+    from graphix._linalg import MatGF2
     from graphix.circ_ext.extraction import CliffordMap, ExtractionResult, PauliExponential, PauliExponentialDAG
+    from graphix.instruction import Instruction
+
+    # NOTE: This alias could be defined at the level of graphix.instruction, and treat all qubit indices as `Qubit`. This change would affect many files in the codebase, so as a temporary solution `Qubit` is casted to `int` in this module.
+    Qubit: TypeAlias = int | np.int_
```
thierry-martinez, line 24: There is a similar definition in the `random_objects` module: [`IntLike`](https://github.com/TeamGraphix/graphix/blob/69b41c4fa9e16f5ec7db8f4525b9c9837596728d/graphix/random_objects.py#L28).
However, it seems you don't rely on `Qubit` including `np.int_` in this PR: if I replace the type alias with `Qubit = int`, the code type-checks. Indeed, you cast qubits to `int` in a few places before calling methods that accept `Qubit`, for instance on line [296](https://github.com/matulni/graphix/blob/a02b5cf367a563fcea5da3ed706c31b2603b52c0/graphix/circ_ext/compilation.py#L296).

graphix/circ_ext/compilation.py
matulni: I just remembered that `CliffordMap.remap` only modifies the `x_map` and `z_map` attributes, meaning that `input_nodes` and `output_nodes` still refer to the original open graph's input-node and output-node lists. This means that they are only equal in length not in values.

It may still be worth rethinking  the `remap` functionality. Currently (for the `CliffordMap`) it's only used by the method `to_tableau`. Since this is an instance method, `to_tableau` could also be responsible of managing the mapping between matrix indices and nodes, without requiring the user to do this conversion beforehand. In this case, `to_tableau` could return a new `Tableau` object, which would be a dataclass containing the `MatGF2` matrix, and two `NodeIndex` mapping input and outputs to qubit indices. The issue with this is that it won't be as explicit as it is now that we use the same qubit-node mapping in the compilation of the Pauli exponential and the Clifford map (see `:func: graphix.circ_ext.compilation.er_to_circuit`).

This node-matrix indices bookkeeping is a recurring problem in Graphix, see `:class: graphix.flow._find_gpflow.AlgebraicOpenGraph`.

matulni: With the current design, we have to remap the CliffordMap to get the tableau, which is a bit painful since we have to provide the qubit-node mapping. Maybe some middle-ground could to introduce a default for the `remap` methods of both `PauliExponentialDAG` and `CliffordMap`: since these objects contain the input and output nodes of the original open graph, they can manage the remapping to qubits on their own.

graphix/circ_ext/compilation.py
```diff
@@ -31,7 +40,7 @@ def er_to_circuit(
     pexp_cp: Callable[[PauliExponentialDAG, Circuit], None] | None
         Compilation pass to synthesize a Pauli exponential DAG. If ``None`` (default), :func:`pexp_ladder_pass` is employed.
     cm_cp: Callable[[CliffordMap, Circuit], None] | None
-        Compilation pass to synthesize a Clifford map. If ``None`` (default), a ``ValueError`` is raised since there is still no default pass for Clifford map integrated in Graphix.
+        Compilation pass to synthesize a Clifford map. If ``None`` (default), :func:`cm_berg_pass` is employed. This pass only handles unitaries so far (Clifford maps with the same number of input and ouptut nodes).
```
thierry-martinez:
```suggestion
        Compilation pass to synthesize a Clifford map. If ``None`` (default), :func:`cm_berg_pass` is employed. This pass only handles unitaries so far (Clifford maps with the same number of input and output nodes).
```

graphix/circ_ext/compilation.py
```diff
@@ -181,3 +188,167 @@ def add_hy(qubit: int, circuit: Circuit) -> None:
...
+    Raises
+    ------
+    NotImplementedError
+        If ``len(clifford_map.input_nodes) != len(clifford_map.output_nodes)``.
+    AssertionError
+        If an unexpected pivot position is encountered during Step 4.
+
+    Notes
+    -----
+    This pass only handles unitaries so far (Clifford maps with the same number of input and ouptut nodes).
```
thierry-martinez:
```suggestion
    This pass only handles unitaries so far (Clifford maps with the same number of input and output nodes).
```

graphix/circ_ext/compilation.py
```diff
@@ -181,3 +188,167 @@ def add_hy(qubit: int, circuit: Circuit) -> None:
...
+    T_q = \begin{pmatrix}
+        XX & XZ \\
+        ZX & ZZ
+    \end{pmatrix}
+
+    1. Clear elements in the XZ-block by applying single-qubit gates (H or S).
+
+    2. Use CNOT gates to reduce the XX block to a single pivot column.
+
+    3. Apply a SWAP gate to bring the pivot to the diagonal if neccesary.
```
thierry-martinez:
```suggestion
    3. Apply a SWAP gate to bring the pivot to the diagonal if necessary.
```

graphix/circ_ext/compilation.py
```diff
@@ -181,3 +188,167 @@ def add_hy(qubit: int, circuit: Circuit) -> None:
...
+        instructions.append(H(q))
+
+    def add_s(tab: MatGF2, instructions: list[Instruction], q: Qubit) -> None:
+        tab[:, -1] ^= tab[:, q] & tab[:, q + n]
+        tab[:, q + n] = tab[:, q] ^ tab[:, q + n]
+        q = int(q)
+        instructions.extend((S(q), Z(q)))  # We append Sdagger to get C instead of C^dagger
+
+    def add_cnot(tab: MatGF2, instructions: list[Instruction], qc: Qubit, qt: Qubit) -> None:
+        tab[:, -1] ^= tab[:, qc] * tab[:, qt + n] * (tab[:, qt] ^ tab[:, qc + n] ^ 1)
```
thierry-martinez: Perhaps `&` would be more readable than `*` for bits.

```suggestion
        tab[:, -1] ^= tab[:, qc] & tab[:, qt + n] & (tab[:, qt] ^ tab[:, qc + n] ^ 1)
```

graphix/circ_ext/compilation.py
```diff
@@ -181,3 +188,167 @@ def add_hy(qubit: int, circuit: Circuit) -> None:
...
+        tab[:, qt] ^= tab[:, qc]
+        tab[:, qc + n] ^= tab[:, qt + n]
+        instructions.append(CNOT(control=int(qc), target=int(qt)))
+
+    def add_swap(tab: MatGF2, instructions: list[Instruction], q0: Qubit, q1: Qubit) -> None:
+        q0, q1 = int(q0), int(q1)  # Cast to `int` to avoid typing issues
+        for shift in [0, n]:
+            tab[:, [q0 + shift, q1 + shift]] = tab[:, [q1 + shift, q0 + shift]]
```
thierry-martinez: Isn't it simpler?
```suggestion
        tab[:, [q0, q1, q0 + n, q1 + n]] = tab[:, [q1, q0, q1 + n, q0 + n]]
```

graphix/circ_ext/compilation.py
```diff
@@ -181,3 +188,167 @@ def add_hy(qubit: int, circuit: Circuit) -> None:
...
+
+    def add_h(tab: MatGF2, instructions: list[Instruction], q: Qubit) -> None:
+        q = int(q)  # Cast to `int` to avoid typing issues
+        tab[:, -1] ^= tab[:, q] & tab[:, q + n]
+        tab[:, [q, q + n]] = tab[:, [q + n, q]]  # The usual tuple assignment `a, b = b, a` does not work here.
+        instructions.append(H(q))
+
+    def add_s(tab: MatGF2, instructions: list[Instruction], q: Qubit) -> None:
+        tab[:, -1] ^= tab[:, q] & tab[:, q + n]
+        tab[:, q + n] = tab[:, q] ^ tab[:, q + n]
```
thierry-martinez:
```suggestion
        tab[:, q + n] ^= tab[:, q]
```

graphix/circ_ext/extraction.py
```diff
@@ -56,9 +58,9 @@ def to_circuit(
         Parameters
         ----------
         pexp_cp: Callable[[PauliExponentialDAG, Circuit], None] | None
-            Compilation pass to synthesize a Pauli exponential DAG. If ``None`` (default), :func:`pexp_ladder_pass` is employed.
-        cm_cp: Callable[[PauliExponentialDAG, Circuit], None] | None
-            Compilation pass to synthesize a Clifford map. If ``None`` (default), a `ValueError` is raised since there is still no default pass for Clifford map integrated in Graphix.
+            Compilation pass to synthesize a Pauli exponential DAG. If ``None`` (default), :func:`graphix.circ_ext.compilation.pexp_ladder_pass` is employed.
+        cm_cp: Callable[[CliffordMap, Circuit], None] | None
+            Compilation pass to synthesize a Clifford map. If ``None`` (default), :func:`graphix.circ_ext.compilation.cm_berg_pass` is employed. This pass only handles unitaries so far (Clifford maps with the same number of input and ouptut nodes).
```
```suggestion
            Compilation pass to synthesize a Clifford map. If ``None`` (default), :func:`graphix.circ_ext.compilation.cm_berg_pass` is employed. This pass only handles unitaries so far (Clifford maps with the same number of input and output nodes).
```

graphix/circ_ext/extraction.py
```diff
@@ -369,6 +371,63 @@ def remap(self, inputs_mapping: Callable[[int], int], outputs_mapping: Callable[
...
+        """
+        n = len(self.input_nodes)
+        if n != len(self.output_nodes):
+            raise NotImplementedError(
+                f"Isometries are not supported yet: # of inputs ({len(self.input_nodes)}) must be equal to the # of outputs ({len(self.output_nodes)})."
+            )
+
+        tab = MatGF2(np.zeros((2 * n, 2 * n + 1)))
+
+        for mapping, shift in zip((self.x_map, self.z_map), (0, n), strict=True):
```
thierry-martinez: Isn't it simpler without `zip`?
```suggestion
        for mapping, shift in (self.x_map, 0), (self.z_map, n):
```

graphix/circ_ext/compilation.py
```diff
@@ -181,3 +188,167 @@ def add_hy(qubit: int, circuit: Circuit) -> None:
...
+    def do_step_1(tab: MatGF2, instructions: list[Instruction], row_idx: int) -> None:
+        col_idx_zx = np.flatnonzero(tab[row_idx, n : 2 * n])  # Don't take the sign column
+        for j in col_idx_zx:
+            # Each iteration sets the element `tab[row_idx, n+j]` to 0.
+            add_s(tab, instructions, int(j)) if tab[row_idx, j] else add_h(tab, instructions, int(j))
+
+    def do_step_2(tab: MatGF2, instructions: list[Instruction], row_idx: int) -> int:
+        col_idx_xx = np.flatnonzero(tab[row_idx, :n])
+        while len(col_idx_xx) > 1:
+            for i in range(0, len(col_idx_xx) - 1, 2):  # itertools.batched only avaialble in Python 3.12+
```
thierry-martinez:
```suggestion
            for i in range(0, len(col_idx_xx) - 1, 2):  # itertools.batched only available in Python 3.12+
```

graphix/circ_ext/compilation.py
matulni: Yes indeed!

matulni:
Hi @thierry-martinez  and @emlynsg.

I concluded that it's not a good idea to remove the `input_nodes` and `output_nodes` fields from `CliffordMap`. These attributes specify the embedding of the Clifford map in a Hilbert space (in the same way as in open graphs or patterns). If we remove them, we could still recover the nodes from `x_map` and `z_map`, but we lose the ordering between nodes, so the Clifford map is ill defined. The same logic applies to `PauliExponentialDAG`.

However, I took from the discussion that it's weird that `.remap` methods don't act on the input node list and output node list. There's indeed no need to keep this trace of the past, and `CliffordMap.remap` should return a fully valid `CliffordMap`. In commit b3ac6be0 I modify `remap` to act on all attributes. I added tests to show that `CliffordMap.remap().remap()` is equivalent to `CliffordMap.remap()` if we use the same mapping, and that we have recover the original object if we use the inverse mapping. Now, `.to_tableau` checks that the Clifford map has been remapped by comparing the list element and not just their lengths as suggested by [Emlyn's comment](https://github.com/TeamGraphix/graphix/pull/476#discussion_r3078531182).  To simplify the use of `to_tableau`, I added a default value to `.remap` so that nodes are mapped to qubit indices automatically (another advantage of having `input_nodes` and `output_nodes` as attributes of `CliffordMap`).

Finally, in commit 5ef2fb2 I added a new field `dim` to `PauliString` so that these objects are fully determined on their own.

Ready for another round of review!

Edit: This [PR](https://github.com/qat-inria/graphix-stim-compiler/pull/3) updates the stim compiler plugin to comply with the new field in `PauliString`

graphix/circ_ext/extraction.py
```diff
@@ -77,16 +80,19 @@ class PauliString:
...
+        Dimension of the Hilbert space on which the Pauli string acts.
     axes : Mapping[int, Axis]
         Mapping between nodes and the applied Pauli operator.
     sign : Sign
         Phase of the Pauli string.
 
     Notes
     -----
-    The identity operator is omitted in this representation, which means that in general it is not possible to infer the size of the Hilbert space from an instance of ``PauliString`` alone.
+    The identity operators in the Pauli string are omitted in ``axes``, but they can be inferred from the dimension of the Hilbert space ``dim``.
```
thierry-martinez, line 91: Identity operators can be inferred only if node indices are known, for instance if they always follow from `range(dim)`. However, this does not hold when the node numbering of the open graph does not form an initial segment of integers.

matulni: You are right. Do you think we should add an attribute `nodes: list[int]` to `PauliString` (and also to `PauliExponential`) to fully determine the space on which these objects live?

emlynsg: Would it be easier and more robust just to catch all the nodes in the keys of the axes object (mapping to nothing or I) and keep the current attribute structure?

matulni: We discussed the mapping to identity in the previous PR, see [this thread](https://github.com/TeamGraphix/graphix/pull/445#discussion_r2982795193)

matulni:
Commit 57790dc9ccffcaba272df4e6c34d45dc4998883d updates the API to incorporate the changes discussed this morning: namely `PauliString` is defined as a mapping from qubit indices to `Axes` so that it's unambiguously determined, and mappings in `CliffordMap` are defined over qubit indices as well. The now unnecessary `remap` functions are removed.

emlynsg (approved):
All makes sense to me!

thierry-martinez (approved):
LGTM! I am still wondering if the type alias `Qubit` is needed (cf. my comment in `compilation.py`) but this looks very good overall.

matulni:
> LGTM! I am still wondering if the type alias `Qubit` is needed (cf. my comment in `compilation.py`) but this looks very good overall.

You are right, I had forgotten about that. I removed the unnecessary type alias in [1b048f5](https://github.com/TeamGraphix/graphix/pull/476/commits/1b048f5972a166b09810df738414f13cb1565cb1)

In [df2aced](https://github.com/TeamGraphix/graphix/pull/476/commits/df2aced23e8135b88fedb082684092dc759baa7f), I renamed the class property `PauliFlow.pauli_strings` as `PauliFlow.extraction_pauli_strings` because I think it's more clear.

I'll merge if CI passes!

PR #479, opened 2026-04-10
Add new visualization tools

matulni:
This PR adds new visualization methods: `OpenGraph.draw`, `PauliFlow.draw` and `XZCorrections.draw`. Further, the method `Pattern.draw_graph` is replaced by `Pattern.draw` with an `annotations` parameter which allows to visualise the underlying open graph, flow or xz corrections.

This PR does not modify the core logic of the existing visualization tools. However, two stylistic modifications are made: input nodes are represented with squares instead of red-edge circles and it incorporates the layer visualisation introduced in #444.

## TODO
Update README with new figure.


graphix/pattern.py
```diff
@@ -1526,65 +1527,144 @@ def _warn_non_inferred_pauli_measurements(self, stacklevel: int) -> None:
                 warnings.warn("Pattern with non-inferred Pauli measurements.", stacklevel=stacklevel + 1)
                 return
 
-    def draw_graph(
+    def draw_flow(
```
emlynsg: I think the naming here is a little confusing since it is called `draw_flow`, also has a parameter `flow_from_pattern`, and doesn't have to draw the flow at all. I think `draw_graph` was clearer than `draw_flow`, but perhaps `draw_resource` or just `draw` would be even better?

graphix/visualization.py
```diff
@@ -1,909 +1,1110 @@
...
+            ha="center",
+            va="top",
+            fontsize=LAYER_FS,
+            color=LAYER_C,
+            transform=base + offset,
+        )
 
-        g = flow.correction_function
+        # Draw horizontal arrow indicating measurement order with "Layer" label below
+        offset = mtransforms.ScaledTranslation(0, -27 / 72, fig.dpi_scale_trans)
```
emlynsg, line 529: Maybe better to save for further work later, but I think shifting the layers arrow down a little would be nicer - further from the nodes, and leaving less white space before the border of the plot.

docs/source/visualization.rst
```diff
@@ -2,7 +2,7 @@ Visualization tool
 ==================
 
 :mod:`graphix.visualization` module
-+++++++++++++++++++++++++++++++++++
++++++++++++++++++++++++++++++++++++++++++
```
thierry-martinez: Is this change intentional?

examples/visualization.py
```diff
@@ -90,7 +88,7 @@
     3: Measurement.YZ(0),
 }
 og = OpenGraph(graph, inputs, outputs, measurements)
-vis = GraphVisualizer(og)
-vis.visualize(show_measurement_planes=True)
+cf = og.extract_gflow()
+cf.draw()
```
thierry-martinez: Should we preserve the flag?
```suggestion
cf.draw(measurement_labels=True)
```

graphix/flow/core.py
```diff
@@ -359,6 +361,50 @@ def to_unicode(self, multiline: bool = False) -> str:
...
+    def draw(
+        self,
+        pauli_measurements: bool = True,
+        measurement_labels: bool = False,
+        node_labels: bool | Mapping[int, str] = True,
+        node_distance: tuple[float, float] = (1, 1),
+        legend: bool = True,
+        figsize: tuple[int, int] | None = None,
+        filename: Path | None = None,
+    ) -> None:
```
thierry-martinez: I propose to reduce some code duplication in this commit: https://github.com/thierry-martinez/graphix/commit/5a6279af23ae13b1a960a3b8165efbe94867681e

graphix/pattern.py
thierry-martinez: I suggest consolidating `draw_flow` and `draw_xzcorrections` into a single method `draw` with an optional `annotations` parameter that accepts the enum values `Flow` (the default), `XZCorrections`, or `None`.

graphix/visualization.py
```diff
@@ -1,909 +1,1110 @@
...
 
-    def visualize(
-        self,
-        show_pauli_measurement: bool = True,
-        show_local_clifford: bool = False,
-        show_measurement_planes: bool = False,
-        show_loop: bool = True,
+    @staticmethod
+    def from_opengraph(
+        og: OpenGraph[AbstractMeasurement],
```
thierry-martinez, line 204: The parameter `local_clifford` is missing: is it intentional?

graphix/pattern.py
```diff
@@ -1526,65 +1527,144 @@ def _warn_non_inferred_pauli_measurements(self, stacklevel: int) -> None:
...
+        pauli_measurements : bool, default=True
+            If ``True``, Pauli-measured nodes are highlighted with distinct coloring.
+        measurement_labels : bool, default=False
+            If ``True``, measurement labels (planes and axis) are displayed in the visualization.
+        node_labels : bool | Mapping[int, str], default=True
+            If ``True``, display numeric node labels. If a mapping, use custom labels
+            for nodes specified in the mapping.
+        local_clifford : Mapping[int, Clifford] | None, default=None
+            Mapping of node identifiers to local Clifford operators. If provided,
+            operators are displayed on their corresponding nodes.
```
thierry-martinez:
```suggestion
        local_clifford : bool, default=False
            If ``True``, the local Clifford operators are printed.
```

graphix/pattern.py
```diff
@@ -1526,65 +1527,144 @@ def _warn_non_inferred_pauli_measurements(self, stacklevel: int) -> None:
...
```
thierry-martinez:
```suggestion
        local_clifford : bool, default=False
            If ``True``, the local Clifford operators are printed.
```

graphix/visualization.py
```diff
@@ -1,909 +1,1110 @@
...
+        # If we don't specify the generic type (we don't need to), mypy will complain.
+
+        # Circumvent import loop
+        from graphix.flow.core import CausalFlow  # noqa: PLC0415
+
+        pos = (
+            _compute_positions_causal_flow(flow)
+            if isinstance(flow, CausalFlow)
+            else _compute_positions_partial_order(flow)
+        )
```
thierry-martinez, line 263: I am not particularly fond of `singledispatch`, but if we want to use it here, we can do as I did in this commit (by explicitly passing a type as an argument to `register`): https://github.com/thierry-martinez/graphix/commit/179d683e310a947e11a4c979bf9b8a0eae35be31

graphix/visualization.py
```diff
@@ -1,909 +1,1110 @@
...
+
+    lmax = len(pol) - 1
+    # Change the x coordinates of the nodes based on their layer, sort in descending order
+    for layer_idx, layer in enumerate(pol):
+        for node in layer:
+            pos[node][0] = lmax - layer_idx
+    return {k: (x, y) for k, (x, y) in pos.items()}
+
+
+def _scale_positions(pos: Mapping[int, _Point], node_distance: tuple[float, float]) -> dict[int, _Point]:
```
thierry-martinez: `mypy` complains because we cannot return a `Mapping[int, _Point]` as a `dict[int, _Point]`. However, since this is an internal function that is only called with `dict` instances, I think it is cleaner to assume `pos` is a `dict`, avoiding the unnecessary copy.

```suggestion
def _scale_positions(pos: dict[int, _Point], node_distance: tuple[float, float]) -> dict[int, _Point]:
```

docs/source/visualization.rst
matulni: Ah, this is because at first it was :mod:`graphix.flow.visualization`. When I reverted to the old file structure I forgot to update this

graphix/visualization.py
matulni: Yes. `local_clifford` only makes sense when you call the constructor from a pattern, since this is never the case for `opengraph` I figured it was not necessary. I can add it if you think it's worth.

graphix/flow/core.py
matulni: Very elegant, thanks!

graphix/visualization.py
matulni: With your proposed refactor of the options, it'll fit very nicely in the dataclass `VisualizationOptions` (and it's there if we want to plot just the open graph of the pattern).

matulni: Done in 13e3d54b500a1c3ce61c36959a6a666405b1f893

graphix/pattern.py
matulni: Done in 13e3d54b500a1c3ce61c36959a6a666405b1f893

graphix/pattern.py
matulni: Fixed in 13e3d54b500a1c3ce61c36959a6a666405b1f893

graphix/pattern.py
matulni: Fixed in 13e3d54b500a1c3ce61c36959a6a666405b1f893

graphix/visualization.py
matulni: Agreed, done in d432ab01

graphix/visualization.py
matulni: Mmm, yes, I'm not happy with this solution. It's not super robust: if you tweak the node distance, the arrow can appear outside the plot. I couldn't figure out how to make it work in all cases, but I think that the layer arrow is a nice addition regardless. Maybe we could fill an issue and leave this detail (together with your suggestion) for the backlog?

docs/source/visualization.rst
matulni: Fixed in https://github.com/TeamGraphix/graphix/commit/13e3d54b500a1c3ce61c36959a6a666405b1f893

examples/visualization.py
matulni: I added your suggestion in 720eb7cd for the sake of consistency with the existing code, but either way we should rework all the examples in the future.

graphix/visualization.py
matulni: Oh very nice, I did't know `register` accepted arguments. Don't you think it's nicer with `singledispatch`? It avoids the import inside the function...

thierry-martinez: I have no strong opinion on this, but note that we still have import loops with `singledispatch`. `from graphix.flow.core import CausalFlow` should be moved to the top level indeed, but that forces `GraphVisualizer` to be imported inside the `draw` methods in `flow/core.py` to break the import loop (though, admittedly, this seems more logical this way). Additionally, I'm not sure what a good name would be for merging `_compute_positions_causal_flow` and `_compute_positions_partial_order` (and should we keep `_compute_positions_opengraph` separate?).

matulni: If we went this way, we could merge all three functions into a single one: `_compute_positions`. I think it makes sense since these three functions do the same thing (in different ways, suitable for different scenarios), but it may make the code less readable...

matulni: Done in ab396cb0

graphix/visualization.py
```diff
@@ -1,909 +1,1093 @@
-"""Functions to visualize the resource state of MBQC pattern."""
+"""Functions to visualize PauliFlow and XZCorrections objects."""
```
thierry-martinez:
```suggestion
"""Functions to visualize graph-based objects: open graphs, flows and XZ-corrections."""
```

thierry-martinez (approved):
LGTM! Thanks, the visualization API is much better that way!

emlynsg (approved):
LGTM!

PR #481, opened 2026-04-13
Fix #454: New space minimization API

### LINKED ISSUE #454: [Bug]: `minimize_space` can increase `max_space`
This is a longstanding bug but I just realized it had not been properly reported yet.

The following example provides a family of patterns where `max_space` is larger after calling `minimize_space` than before. `sz = 4` and `depth = 3` are the minimum values for which the problem occurs, but the gap between `max_space` before and after increases as these values grow (and the initial `max_space` depends only on `sz`, not on `depth`).

```python
from graphix import Pattern, command
from graphix.measurements import Measurement

sz = 4
depth = 3
block = Pattern(input_nodes=range(sz))
block.extend(command.N(sz + i) for i in range(sz))
for i in range(1, sz):
    block.extend(command.E((j, sz + i)) for j in range(sz))
    block.extend(command.E((sz + j, sz + i)) for j in range(1, i))
block.extend(command.M(i) for i in range(sz))
p, _ = block.compose(block, {i: o for i, o in zip(block.input_nodes, block.output_nodes, strict=True)})
for _ in range(depth):
    p, _ = p.compose(block, {i: o for i, o in zip(block.input_nodes, p.output_nodes, strict=True)})
print(f"Before minimize_space: {p.max_space()=}")
p.minimize_space()
print(f"After minimize_space: {p.max_space()=}")
```

These patterns have X measurements; therefore they do not have causal flows, so that `minimize_space` uses the heuristics implemented in `_measurement_order_space` (selecting nodes with minimum degree first).

thierry-martinez:
This commit introduces a new module `space_minimization` containing algorithm implementations related to pattern space minimization.

The new `minimize_space` method and function are now defined on `StandardizedPattern`, because they require the pattern to be standardized and can take advantage of the specialized representation. These take a new parameter `heuristics` to customize the heuristics used for space minimization, enabling future introduction of new heuristics.

The method `max_space` is now defined for both `Pattern` (as before) and `StandardizedPattern`. In `StandardizedPattern`, `max_space` returns the maximum space of the pattern that would be returned by `to_space_optimal_pattern`, without actually constructing it. This method is used as an evaluation function for `minimize_space`.

The default heuristics implemented in this commit ensure that `minimize_space` does not return a pattern with greater `max_space` than the original:
```python
[minimization_using_causal_flow, greedy_minimization_by_degree, do_nothing_for_space_minimization]
```
By contrast, the default heuristics implemented previously in `graphix` correspond to the following list of heuristics:
```python
[minimization_using_causal_flow, greedy_minimization_by_degree]
```


graphix/space_minimization.py
```diff
@@ -0,0 +1,302 @@
...
+    try:
+        cf = pattern.extract_causal_flow()
+    except FlowError:
+        return None
+    else:
+        meas_order = tuple(chain(*reversed(cf.partial_order_layers[1:])))
+        return meas_order, True
+
+
+def do_nothing_for_space_minimization(pattern: StandardizedPattern) -> tuple[tuple[Node, ...], bool] | None:
```
emlynsg: I would suggest a consistent naming convention for these heuristic functions, e.g. `space_minimization_do_nothing`, `space_minimization_using_causal_flow`, etc.

graphix/space_minimization.py
```diff
@@ -0,0 +1,302 @@
...
+    needed for classical simulation.
+
+    Parameters
+    ----------
+    pattern : Pattern
+        The pattern.
+
+    Returns
+    -------
+    n_nodes : int
```
matulni:
```suggestion
    max_active : int
```

graphix/space_minimization.py
```diff
@@ -0,0 +1,302 @@
...
+
+
+def standardized_pattern_max_space(pattern: StandardizedPattern) -> int:
+    """Compute the maximum number of nodes that must be present in the graph (graph space) during the execution of the space-optimal pattern for the given measurement order.
+
+    This is equivalent to ``pattern.to_space_optimal_pattern().max_space()``.
+
+    Returns
+    -------
+    n_nodes : int
```
matulni:
```suggestion
    max_active : int
```

graphix/space_minimization.py
```diff
@@ -0,0 +1,302 @@
...
+    # keys() should be converted into `set` because it is transient.
+    _update_dependency(set(pattern.results.keys()), dependency)
+    meas_order = []
+    removable_edges: set[tuple[Node, Node]] = set()
+    while not_measured:
+        min_edges = len(nodes) + 1
+        next_node: Node | None = None
+        for i in not_measured:
+            if not dependency[i]:
+                connected_edges = set(graph.edges(i)) & edges
```
matulni: I noticed that `graph.edges(node)` "rewrites" the graph edges so that `node` is the first element of the tuple. This can lead to unexpected results:

```python
import networkx as nx

g = nx.Graph([(0, 1), (1, 2), (0, 2), (2, 3)])

assert not set(g.edges(2)) == set(g.edges(2)) & set(g.edges)
assert set(g.edges(0)) == set(g.edges(0)) & set(g.edges)
```

Could this be a problem here ?

graphix/space_minimization.py
```diff
@@ -0,0 +1,302 @@
...
+                if min_edges > len(connected_edges):
+                    min_edges = len(connected_edges)
+                    next_node = i
+                    removable_edges = connected_edges
+        assert next_node is not None
+        meas_order.append(next_node)
+        _update_dependency({next_node}, dependency)
+        not_measured -= {next_node}
+        edges -= removable_edges
```
matulni: I believe this is equivalent to the existing function and avoids potential problems with `graph.edge`. It requires importing `operator` (suggested by ruff to avoid a lambda).

```suggestion
    graph = pattern.extract_graph()
    nodes = set(graph.nodes)
    not_measured = nodes - set(pattern.output_nodes)
    dependency = _extract_dependency(pattern)
    # keys() should be converted into `set` because it is transient.
    _update_dependency(set(pattern.results.keys()), dependency)
    meas_order = []
    while not_measured:
        next_node, min_degree = max(graph.degree, key=operator.itemgetter(1))
        for i in not_measured:
            if not dependency[i] and graph.degree(i) < min_degree:
                next_node = i
                min_degree = graph.degree(i)
        meas_order.append(next_node)
        _update_dependency({next_node}, dependency)
        not_measured -= {next_node}
        graph.remove_nodes_from({next_node})
```

graphix/space_minimization.py
```diff
@@ -0,0 +1,302 @@
...
+        By default:
+        ``[minimization_using_causal_flow, greedy_minimization_by_degree, do_nothing_for_space_minimization]``
+
+    Returns
+    -------
+    StandardizedPattern
+        The optimized pattern.
+    """
+    if heuristics is None:
+        heuristics = [minimization_using_causal_flow, greedy_minimization_by_degree, do_nothing_for_space_minimization]
```
matulni: I'm keen on the idea of having an `Enum` to enumerate all possible heuristics implemented in Graphix, but I can't see a strong reason other than (very subjective) "elegance". Do you think it would add too much boilerplate without bringing much gain?

Here's an example on how to have functions as  `enum` members inspired from https://stackoverflow.com/a/74302109

```python
from enum import Enum, member

def f1(x):
    print("F1 says:", x)

def f2(x):
    print("F2 says:", x)

class A(Enum):
    F1 = member(f1)
    F2 = member(f2)

    def __call__(self, *args, **kwargs):
        return self.value(*args, **kwargs)

for f in A:
    f(3)
```

graphix/space_minimization.py
```diff
@@ -0,0 +1,302 @@
...
+                _best_pattern, best_max_space = best_pattern_with_max_space
+                if best_max_space <= tentative_max_space:
+                    continue
+            best_pattern_with_max_space = tentative_pattern, tentative_max_space
+    if best_pattern_with_max_space is None:
+        return pattern
+    return best_pattern_with_max_space[0]
+
+
+def minimization_using_causal_flow(pattern: StandardizedPattern) -> tuple[tuple[Node, ...], bool] | None:
```
matulni: I would suggest introducing  a `NamedTuple` or `dataclass` for the heuristic result, but I don't have strong opinion if you think it's too much boilerplate.

matulni (commented):
Thanks, new API is much nicer! I made some comments.

graphix/space_minimization.py
thierry-martinez: Nice catch, it is a problem indeed, thanks! Fix and test added in bf11693a.

graphix/space_minimization.py
thierry-martinez: You're right: mutating the graph makes the code much simpler. I made it even more simpler in 08e21d33, following your idea.

graphix/space_minimization.py
thierry-martinez: A `dataclass` is better than a tuple indeed! Fixed in 8993ecc.

graphix/space_minimization.py
thierry-martinez: I would prefer to let users use the module prefix `space_minimization.` if they prefer. However, I renamed `do_nothing_for_space_minimization` to the more descriptive `keep_measurement_order_unchanged` in 8993ecc.

graphix/space_minimization.py
thierry-martinez: Could you elaborate on why you think this would be more elegant? Currently, a heuristic is a function from patterns to measurement orders, and I don't see what benefits we could expect from wrapping this in an `Enum`. A drawback is that wrapping heuristics in an enum turns them into a _closed-world_ representation, where the set of applicable heuristics is limited to those listed in the enum. By contrast, defining them as simple functions makes the representation _open-world_, where new heuristics can be freely defined by users.

graphix/space_minimization.py
matulni: Very nice, thanks!

graphix/space_minimization.py
matulni: I feel it's a bit disappointing that there's not an abstraction encompassing all the heuristics defined in Graphix. Grouping them (in an `Enum` or something else) would allow to see all existing heuristics at a glance, and it would enable autocompletion in most IDEs.

In addition, with the `Enum` abstraction, the default pipeline (`[minimization_using_causal_flow, greedy_minimization_by_degree, keep_measurement_order_unchanged]`) would be encoded in the abstraction itself (by the order of the `Enum` members) instead of being hard-coded in the body of `minimize_space` (and copy-pasted in various docstrings).

I take your point that wrapping heuristics in an enum turns them into a closed-world representation, so perhaps an `Enum` is not the best abstraction for what I'd like to achieve. Would grouping them in a module and defining the default pipeline as a constant  (a tuple) be better ? Again, if you think that's too much for just 3 heuristics, I'm happy with your current design too.

thierry-martinez: I introduced a namespace `SpaceMinimizationHeuristics` to define them there as `causal_flow`, `greedy_degree`, and `order_unchanged`. I also defined the constant `DEFAULT_HEURISTICS` and restructured the documentation.

matulni (approved):
LGTM, thanks!

emlynsg (approved):
I think we need to be very clear in the docstrings and documentation the difference between the minimise space, to_space_minimial_pattern and the heuristics.
The StandardizedPattern methods permute the measurement order (returning a StandardizedPattern) and the methods on the Pattern class have the freedom to move all command types.

thierry-martinez:
@emlynsg , I tried to improve the docstrings in e4f47481. What do you think of this?

emlynsg:
@thierry-martinez makes much more sense now, thanks!

PR #484, opened 2026-04-17
Replacing Graphix transpiler with `graphix-jcz-transpiler`

emlynsg:
This PR takes the causal flow-based transpilation method detailed in the [graphix-jcz-transpiler](https://github.com/qat-inria/graphix-jcz-transpiler) and adds it to Graphix, replacing the previous transpilation approach into one that is more principled and consistent with the literature [1].

Major changes include:

- Full rework of the `transpiler.py` module to implement $J,\, CZ$ gate decomposition and use the causal-flow-based approach to construct patterns
- Addition of a $J(\alpha)$ gate to circuits, with equivalent additions to random circuits, `ops` and management to codebase components related to OpenQASM
- Updates to many tests to account for the new transpiler generating `BlochMeasurement` commands which need to be inferred as Pauli (usually for preprocessing)

[1] The Measurement Calculus
[Vincent Danos](https://arxiv.org/search/quant-ph?searchtype=author&query=Danos,+V), [Elham Kashefi](https://arxiv.org/search/quant-ph?searchtype=author&query=Kashefi,+E), [Prakash Panangaden](https://arxiv.org/search/quant-ph?searchtype=author&query=Panangaden,+P)


thierry-martinez:
Veriphix tests are likely being killed by the OOM killer. When running the Veriphix test suite locally, I observed that it hangs on the very first test of `test_client.py`. This test runs a delegated random circuit that only standardizes the pattern, without using Pauli presimulation or space minimization. There is a difficulty that I don't know how to address here: the new transpiler cannot match the performance of the original one without Pauli presimulation, but we do not yet know how to implement Pauli presimulation without setting the input.

emlynsg:
> Veriphix tests are likely being killed by the OOM killer. When running the Veriphix test suite locally, I observed that it hangs on the very first test of `test_client.py`. This test runs a delegated random circuit that only standardizes the pattern, without using Pauli presimulation or space minimization. There is a difficulty that I don't know how to address here: the new transpiler cannot match the performance of the original one without Pauli presimulation, but we do not yet know how to implement Pauli presimulation without setting the input.

I agree that seems to be the case. I can imagine it should work if either `minimize_space` or `to_space_minimal_pattern` are applied -- let's discuss tomorrow morning.

thierry-martinez (approved):
LGTM! Thank you very much for this big change.

graphix/transpiler.py
```diff
@@ -7,45 +7,77 @@
...
 
 @dataclass
-class TranspileResult:
+class TranspileResult(Generic[_R, _CO]):
     """
     The result of a transpilation.
 
-    pattern : :class:`graphix.pattern.Pattern` object
-    classical_outputs : tuple[int,...], index of nodes measured with *M* gates
+    pattern : :class:`graphix.pattern.Pattern` or :class:`graphix.flow.core.CausalFlow` object
```
matulni:
```suggestion
    result : :class:`graphix.pattern.Pattern` or :class:`graphix.flow.core.CausalFlow` object
```

graphix/transpiler.py
```diff
@@ -7,45 +7,77 @@
...
+    def pattern(self) -> Pattern:
+        """Return pattern from TranspileResult if any.
+
+        Raises
+        ------
+            TypeError: if the TranspileResult stores a flow instead.
+        """
+        if not isinstance(self.result, graphix.pattern.Pattern):
+            raise TypeError("result is not a Pattern; use `.flow` or `.result`")
+        return self.result
```
matulni: I think it's preferable to enforce the typing with the type-checker instead of doing dynamic checks:

```suggestion
    def pattern(self: TranspileResult[Pattern, _CO]) -> Pattern:
        """Return pattern from `TranspileResult` if any."""
        return self.result
```

You could also set `_CO` to be `tuple[int, ...]` to be more precise.

Just out of curiosity, is this an LLM suggestion? I observe that they are super defensive with runtime checks and never default to type-checkers

graphix/transpiler.py
```diff
@@ -7,45 +7,77 @@
...
+    @property
+    def flow(self) -> CausalFlow[BlochMeasurement]:
+        """Return causal flow from TranspileResult if any.
+
+        Raises
+        ------
+            TypeError: if the TranspileResult stores a pattern instead.
+        """
+        if not isinstance(self.result, CausalFlow):
+            raise TypeError("result is not a CausalFlow; use `.pattern` or `.result`")
```
matulni: See previous comment

graphix/transpiler.py
```diff
@@ -387,559 +434,90 @@ def m(self, qubit: int, axis: Axis) -> None:
...
+        Raises
+        ------
+            IllformedCircuitError: if the pattern is ill-formed (operation on already measured node)
+        """
+        indices: list[int | None] = list(range(self.width))
+        n_nodes = self.width
+        measurements: dict[int, BlochMeasurement] = {}
+        classical_outputs: dict[int, command.M] = {}
+        inputs = list(range(n_nodes))
+        graph: nx.Graph[int] = nx.Graph()  # In future change to OpenGraph not nx
```
matulni: Comment is no longer relevant, right ?

```suggestion
        graph: nx.Graph[int] = nx.Graph()
```

graphix/transpiler.py
```diff
@@ -387,559 +434,90 @@ def m(self, qubit: int, axis: Axis) -> None:
...
+                        raise IllformedCircuitError
+                    # If edge exists, remove it; else, add it
+                    if graph.has_edge(i0, i1):
+                        graph.remove_edge(i0, i1)
+                    else:
+                        graph.add_edge(i0, i1)
+                    continue
+                assert_never(instr_jcz.kind)
+        outputs = [i for i in indices if i is not None]
+        outputs.extend(classical_outputs.keys())
```
matulni: Why are the classical outputs added to the `OpenGraph` outputs? If they are measured, they shouldn't be outputs, right?

graphix/transpiler.py
```diff
@@ -387,559 +434,90 @@ def m(self, qubit: int, axis: Axis) -> None:
...
-        control_node1: int,
-        control_node2: int,
-        target_node: int,
-        ancilla: Sequence[int],
-    ) -> tuple[int, int, int, list[command.Command]]:
-        """MBQC commands for CCX gate.
+        z_corrections: dict[int, set[int]] = {}
+        for node, correctors in x_corrections.items():
+            (corrector,) = correctors
+            z_targets = set(graph.neighbors(corrector)) - {node}
```
matulni:
```suggestion
            z_targets = og.neighbors(correctors) - {node}
```

graphix/transpiler.py
```diff
@@ -387,559 +434,90 @@ def m(self, qubit: int, axis: Axis) -> None:
...
-        output_nodes = [node for node in out if node is not None]
-        pattern.reorder_output_nodes(output_nodes)
-        return TranspileResult(pattern, tuple(classical_outputs))
-
-    @classmethod
-    def _cnot_command(
-        cls, control_node: int, target_node: int, ancilla: Sequence[int]
-    ) -> tuple[int, int, list[command.Command]]:
-        """MBQC commands for CNOT gate.
+    def transpile_to_cflow(self) -> TranspileResult[CausalFlow[BlochMeasurement], dict[int, command.M]]:
```
matulni: For discussion: this is inconsistent with naming in the `OpenGraph` suite where we have `extract_causal_flow`.

I think we need to homogenise the naming of flows (one possibility could be to call everything `cflow`, `gflow` and `pflow`, including the classes), although this is outside the scope of this review. We can leave it as is for now.

graphix/transpiler.py
matulni: Ok I understand now that they are dealt with when converting the flow to a pattern. It may be worth adding a comment...

graphix/transpiler.py
```diff
@@ -1152,3 +1005,19 @@ def transpile_swaps(circuit: Circuit) -> TranspileSwapsResult:
...
+    pattern.extend(tr.classical_outputs.values())
+    return TranspileResult(pattern, tuple(tr.classical_outputs.keys()))
+
+
+class IllformedCircuitError(Exception):
+    """Raised if the circuit is ill-formed."""
+
+    def __init__(self) -> None:
+        """Build the exception."""
+        super().__init__("Ill-formed pattern")
```
matulni:
```suggestion
        super().__init__("Ill-formed circuit")
```

graphix/qasm3_exporter.py
```diff
@@ -251,3 +274,22 @@ def domain_to_qasm3_lines(domain: Iterable[int], cmd: str) -> Iterator[str]:
     yield f"if ({condition}) {{\n"
     yield f"  {cmd};\n"
     yield "}\n"
+
+
+def _decompose_j_gates(circuit: Circuit) -> Circuit:
```
matulni: I think the API would be more user-friendly if this was a method of `Circuit` (like `Circuit.transpile_measurements_to_z_axis`). Also, `stacklevel` should be a key-word only argument of the function to be able to propagate warnings

graphix/transpiler.py
```diff
@@ -1085,14 +665,287 @@ def transpile_measurements_to_z_axis(self) -> Circuit:
...
+    Parameters
+    ----------
+        instr: the Y instruction to decompose.
+
+    Returns
+    -------
+        the decomposition.
+
+    """
+    return list(reversed([instruction.X(instr.target), instruction.Z(instr.target)]))
```
matulni:
```suggestion
    return [instruction.Z(instr.target), instruction.X(instr.target)]
```

graphix/transpiler.py
```diff
@@ -1085,14 +665,287 @@ def transpile_measurements_to_z_axis(self) -> Circuit:
...
+    if instr.kind == InstructionKind.J:
+        return [instr]
+    if instr.kind == InstructionKind.CZ:
+        return [instr]
+    if instr.kind == InstructionKind.I:
+        return []
+    if instr.kind == InstructionKind.H:
+        return [instruction.J(instr.target, 0)]
+    if instr.kind == InstructionKind.S:
+        return instruction_to_jcz(instruction.RZ(instr.target, ANGLE_PI / 2))
```
matulni: This could simply be 

```suggestion
        return decompose_rz(instruction.RZ(instr.target, ANGLE_PI / 2))
```
right ?

No need to for recursive calls

graphix/transpiler.py
```diff
@@ -1085,14 +665,287 @@ def transpile_measurements_to_z_axis(self) -> Circuit:
...
+    if instr.kind == InstructionKind.X:
+        return instruction_to_jcz(instruction.RX(instr.target, ANGLE_PI))
```
matulni:
```suggestion
        return decompose_rx(instruction.RX(instr.target, ANGLE_PI))
```

graphix/transpiler.py
```diff
@@ -1085,14 +665,287 @@ def transpile_measurements_to_z_axis(self) -> Circuit:
...
+    if instr.kind == InstructionKind.Y:
+        return instruction_list_to_jcz(decompose_y(instr))
+    if instr.kind == InstructionKind.Z:
+        return instruction_to_jcz(instruction.RZ(instr.target, ANGLE_PI))
```
matulni:
```suggestion
        return decompose_rz(instruction.RZ(instr.target, ANGLE_PI))
```

graphix/qasm3_exporter.py
```diff
@@ -40,6 +44,7 @@ def circuit_to_qasm3_lines(circuit: Circuit) -> Iterator[str]:
     Iterator[str]
         The OpenQASM 3.0 lines that represent the circuit.
     """
+    circuit = _decompose_j_gates(circuit)
```
matulni: This is inconsistent with the API measurement commands: if there are measurements along an axis other than Z, the parser will fail (users must convert it themselves). I think we should either convert all commands to get a graceful conversion or non.

tests/test_transpiler.py
```diff
@@ -39,111 +40,37 @@
...
-        circuit = Circuit(1)
-        circuit.rz(0, theta)
+    @pytest.mark.parametrize("instruction", INSTRUCTION_TEST_CASES)
+    def test_instruction_flow(self, fx_rng: Generator, instruction: InstructionTestCase) -> None:
+        circuit = Circuit(3, instr=[instruction(fx_rng)])
         pattern = circuit.transpile().pattern
-        state = circuit.simulate_statevector(rng=fx_rng).statevec
-        state_mbqc = pattern.simulate_pattern(rng=fx_rng)
-        assert state_mbqc.isclose(state)
+        flow = pattern.to_bloch().extract_causal_flow()
```
matulni, line 57: Here wouldn't it make sense to check on `circuit.transpile_to_cflow`?

tests/test_transpiler.py
```diff
@@ -39,111 +40,37 @@
...
+        rng = np.random.default_rng(420)
+        circuit = Circuit(3, instr=[instruction.CCX(0, (1, 2))])
         pattern = circuit.transpile().pattern
         pattern.minimize_space()
-        state = circuit.simulate_statevector(rng=rng).statevec
-        state_mbqc = pattern.simulate_pattern(rng=rng)
+        input_state = rand_state_vector(3, rng=rng)
+        state = circuit.simulate_statevector(input_state=input_state).statevec
+        state_mbqc = pattern.simulate_pattern(input_state=input_state, rng=rng)
         assert state_mbqc.isclose(state)
```
matulni: I think this test is already covered by `test_instructions`

matulni (commented):
Thanks @emlynsg for this nice work! Unfortunately I don't know the details about the JCZ transpiler so I only made comments about the code.

On a general note, your `decompose_...` functions lend themselves to return iterators (check the old `_transpile_rzz`), which are evaluated lazily and avoid allocating lists in the intermediate steps (you may also want to check [`map`](https://docs.python.org/3/library/functions.html#map)). I didn't check, so I won't make any claims on its performance (my guess is that it's not significant for the circuit sizes that we deal with). I will not bother you with it, we can leave it as is 😬

graphix/transpiler.py
emlynsg: Yeah you're right, I wanted to keep the API to return pattern with `.pattern` and this is what it suggested. Good point though, I always forget we can enforce typing on `self`.

graphix/qasm3_exporter.py
emlynsg: I don't understand what you are saying - this line doesn't change whether or not there are non-Z measurements, since it doesn't introduce additional measurements nor does it change existing ones - it just replaces J with RZ and H gates.

graphix/transpiler.py
emlynsg: Nice catches, saves a step for each of these.

graphix/qasm3_exporter.py
matulni: Sorry, I was not clear.

You can see in `instruction_to_qasm3` that if the circuit has `M` instructions with `instruction.axis` different from `Axis.Z`, the parser raises a `ValueError` and points the user to use `Circuit.transpile_measurements_to_z_axis`. In other words, the parser does not do any implicit conversion and it's the user's full responsability to prepare the circuit before parsing.

In a similar way, if the circuit has a `J` instruction, `instruction_to_qasm3` will fail. However, with current design, the parser implicitly removes all `J` instructions beforehand. All I'm saying is that this behaviour is not consistent with the design choice for `M` commands: for `J` we automatically modify the circuit and for `M` we trust the user to do so.

thierry-martinez: I understand the argument for consistency; however, I like the automatic transpilation to provide a better user experience.

What I propose:
- Move `_decompose_j_gates` into `Circuit.transpile_j_gates` to expose it as part of the public API.
- Add a `transpile: bool = True` parameter to the `circuit_to_qasm3*` functions. If `True`, these functions will call `transpile_measurements_to_z_axis` and `transpile_j_gates`. If `False`, the export should fail if non-Z measurements or J gates are detected.

emlynsg: Just added the bool flag.

graphix/qasm3_exporter.py
```diff
@@ -22,25 +22,31 @@
...
     """Export circuit instructions to line-by-line OpenQASM 3.0 representation.
 
     Returns
     -------
     Iterator[str]
         The OpenQASM 3.0 lines that represent the circuit.
     """
+    if transpile:
+        circuit = circuit.transpile_j_to_rzh().transpile_measurements_to_z_axis()
+    if any(instr.kind == InstructionKind.J for instr in circuit.instruction):
```
matulni: I'm not sure if we need this check, we are already checking in `instruction_to_qasm3`.
By the same logic, we should also check that there are not non-Z measurements (which we do at `instruction_to_qasm3`). 

If we decide to keep it:

```suggestion
    elif any(instr.kind == InstructionKind.J for instr in circuit.instruction):
```

graphix/qasm3_exporter.py
```diff
@@ -22,25 +22,31 @@
...
     """Export circuit instructions to OpenQASM 3.0 representation.
 
     Returns
     -------
     str
         The OpenQASM 3.0 string representation of the circuit.
     """
-    return "\n".join(circuit_to_qasm3_lines(circuit))
+    if transpile:
+        circuit = circuit.transpile_j_to_rzh().transpile_measurements_to_z_axis()
```
matulni: We are transpiling `j` and measurements in `circuit_to_qasm_lines`, so I think there's no need to transpile twice. 
```suggestion
```

graphix/qasm3_exporter.py
```diff
@@ -22,25 +22,31 @@
...
 
     Returns
     -------
     Iterator[str]
         The OpenQASM 3.0 lines that represent the circuit.
     """
+    if transpile:
+        circuit = circuit.transpile_j_to_rzh().transpile_measurements_to_z_axis()
+    if any(instr.kind == InstructionKind.J for instr in circuit.instruction):
+        raise ValueError("J gates must be decomposed before QASM3 export using `Circuit.transpile_j_to_rzh`.")
```
matulni:
```suggestion
        raise ValueError("J gates must be decomposed before QASM3 export using `Circuit.transpile_j_to_rzh`, or setting `transpile=True`.")
```

tests/test_pattern.py
```diff
@@ -1031,8 +1048,12 @@ def test_extract_xzc_rnd_circuit(self, fx_bg: PCG64, jumps: int) -> None:
         xzc.check_well_formed()
         p_test = xzc.to_pattern()
 
-        for p in [p_ref, p_test]:
-            p.remove_pauli_measurements()
+        p_test = p_test.infer_pauli_measurements()
```
matulni: Done below
```suggestion
```

graphix/transpiler.py
```diff
@@ -6,58 +6,78 @@
...
-    classical_outputs : tuple[int,...], index of nodes measured with *M* gates
+    result : :class:`graphix.pattern.Pattern` or :class:`graphix.flow.core.CausalFlow` object
+    classical_outputs : tuple[int, ...] | dict[int, command.M], index of nodes measured with *M* gates, with associated M commands as dictionary.
+
     """
 
-    pattern: Pattern
-    classical_outputs: tuple[int, ...]
+    result: _R
+    classical_outputs: _CO
```
matulni: Maybe it's an overkill, but do you think it would be worth to introduce a property `classical_outputs` to narrow the type `_CO` depending on the type of `_R`?

graphix/transpiler.py
```diff
@@ -1151,24 +966,77 @@ class TranspileSwapsResult:
...
+        for index, output in enumerate(self.outputs):
+            if output.kind == OutputKind.Qubit:
+                reduced_index[index] = reduced_counter
+                reduced_counter += 1
+        return tuple(reduced_index[index] for index in self.extract_outputs(OutputKind.Qubit))
+
+    def swap_statevec(self, statevec: Statevec) -> Statevec:
+        """Reorder the elements of a statevector obtained from a swapped circuit to restore the qubit ordering of the original circuit."""
+        psi = np.transpose(statevec.psi, self.extract_output_node_indices())
+        return Statevec(psi.flatten(), statevec.nqubit)
```
matulni: I think the logic in this method deserves to be a method of `Statevec` (or even `DenseState`).
The reason is twofold:
- It relies on the internal representation of `statevec.psi` (in fact, it will fail with the new jit backend).
- We could use it in `DenseStateBackend.sort_qubits` instead of repeated calls to `swap` (outside of this PR).

matulni (commented):
Thank you, very nice work. We should be able recycle some of your ideas in `TranspileSwapResult` in a more general API for `Circuit` !

I made some minor comments (and there remain a few comments from a previous review I made to resolve).

graphix/transpiler.py
thierry-martinez: Renamed into `transpile_to_causal_flow` in 094c159a.

graphix/qasm3_exporter.py
thierry-martinez: Fixed in 5f13bad5a.

graphix/qasm3_exporter.py
thierry-martinez: This check is now removed.

graphix/qasm3_exporter.py
thierry-martinez: Removed.

graphix/qasm3_exporter.py
thierry-martinez: Modified, and applied to measurements.

tests/test_pattern.py
thierry-martinez: Removed.

graphix/transpiler.py
thierry-martinez: I split `TranspileResult` into `TranspiledPattern` and `TranspiledFlow`.

graphix/transpiler.py
thierry-martinez: Added `DenseBackend.permute`.

graphix/sim/base_backend.py
```diff
@@ -826,12 +853,18 @@ def sort_qubits(self, output_nodes: Iterable[int]) -> None:
...
-        self.sort_qubits(output_nodes)
+    # @override
+    # def finalize(self, output_nodes: Iterable[int]) -> None:
+    #    """To be run at the end of pattern simulation."""
+    #    from copy import copy
+    #    duplicate = copy(self)
+    #    duplicate.sort_qubits(output_nodes)
+    #    self.state.permute([self.node_index.index(node) for node in output_nodes])
+    #    if not self.state.isclose(duplicate.state):
+    #        assert False, output_nodes
```
matulni: Can we remove this comment ?

CHANGELOG.md
matulni, line 1: Could you please update the changelog with the modifications in the backends?

matulni (approved):
Thanks, LGTM!

PR #489, opened 2026-04-21
Fix #235: Correct sign for YZ measurements in `from_pyzx_graph`

### LINKED ISSUE #235: [Bug]: `to_pyzx_graph`/`from_pyzx_graph` don't preserve the semantics when calling `pyzx.simplify.full_reduce`.

thierry-martinez:
This commit fixes a bug in `from_pyzx_graph` that failed to preserve the semantics of reduced ZX-diagrams. For YZ measurements (phase gadgets), a phase gadget with phase α adds a Z(α) rotation.

The negative sign introduced in #237 for XY measurements was incorrectly applied to YZ measurements. In #237, that sign change was necessary for XY measurements because a Z-spider with phase α corresponds to a measurement angle -α in MBQC.

The non-reduced ZX-diagrams were handled correctly because they did not contain any phase gadgets.

Tests have been refactored to verify both the non-reduced and reduced graphs using random circuits and the CCZ circuit. The `use_ccx` flag is omitted during random circuit generation to prevent tests from becoming too slow.

This commit also utilizes the `seed` keyword parameter for `cliffordT` (introduced in `pyzx` 0.10.0) to avoid modifying the global seed.


matulni (approved):
LGTM thanks for the fix

emlynsg (approved):

PR #490, opened 2026-04-21
Add common imports to top level namespace

emlynsg:
PR adds commonly used objects to the top level namespace `__init__.py`.
Minor changes were made to avoid circular imports.


graphix/__init__.py
```diff
@@ -2,9 +2,44 @@
 
 from __future__ import annotations
 
+from graphix._version import __version__  # or wherever your version lives
+from graphix.command import E, M, N, X, Z
```
thierry-martinez: Why is `C` excluded?

To avoid namespace pollution caused by single-letter symbols, I prefer using `from graphix import command`, allowing for the `command.` prefix.

Currently, my usage is inconsistent with other `graphix` objects that are designed for direct import. I propose resolving this by adding commands as static fields to the `Command` class and exporting that class in the top-level `graphix` namespace.

This change will improve IDE discoverability and allow `repr()` to consistently use the `Command.` prefix. At present, `repr()` for commands and patterns relies on the assumption that single-letter symbols are already present in the local namespace.

I would do the same for instructions as well.

graphix/__init__.py
```diff
@@ -2,9 +2,44 @@
...
+from graphix.flow.core import CausalFlow, GFlow, PauliFlow, XZCorrections
+from graphix.fundamentals import Axis, Plane, Sign
```
thierry-martinez: I would add `ANGLE_PI`, `angle_to_rad`, `rad_to_angle`.

graphix/__init__.py
```diff
@@ -2,9 +2,44 @@
...
+    "PauliMeasurement",
+    "Plane",
+    "Sign",
+    "Statevec",
+    "StatevectorBackend",
+    "X",
+    "XZCorrections",
+    "Z",
+    "__version__",
+]
```
thierry-martinez, line 71: I would add the branch selectors, `Clifford`, `KrausChannel`, `NoiseModel`, `DepolarisingNoiseModel`, `Pauli`, `Placeholder`, and `OutputFormat`.

graphix/optimization.py
```diff
@@ -707,7 +710,9 @@ def _update_corrections(node: Node, domain: AbstractSet[Node], correction: dict[
 
 def incorporate_pauli_results(pattern: Pattern) -> Pattern:
     """Return an equivalent pattern where results from Pauli presimulation are integrated in corrections."""
-    result = graphix.pattern.Pattern(input_nodes=pattern.input_nodes)
+    from graphix.pattern import Pattern  # noqa: PLC0415
```
thierry-martinez, line 731:
```suggestion
    from graphix import Pattern  # noqa: PLC0415
```

graphix/optimization.py
```diff
@@ -752,7 +757,9 @@ def incorporate_pauli_results(pattern: Pattern) -> Pattern:
 
 def remove_useless_domains(pattern: Pattern) -> Pattern:
     """Return an equivalent pattern where measurement domains that are not used given the specific measurement angles and planes are removed."""
-    new_pattern = graphix.pattern.Pattern(input_nodes=pattern.input_nodes)
+    from graphix.pattern import Pattern  # noqa: PLC0415
```
thierry-martinez, line 778:
```suggestion
    from graphix import Pattern  # noqa: PLC0415
```

graphix/optimization.py
```diff
@@ -772,7 +779,9 @@ def remove_useless_domains(pattern: Pattern) -> Pattern:
 
 def single_qubit_domains(pattern: Pattern) -> Pattern:
     """Return an equivalent pattern where domains contains at most one qubit."""
-    new_pattern = graphix.pattern.Pattern(input_nodes=pattern.input_nodes)
+    from graphix.pattern import Pattern  # noqa: PLC0415
```
thierry-martinez, line 800:
```suggestion
    from graphix import Pattern  # noqa: PLC0415
```

graphix/optimization.py
```diff
@@ -349,7 +348,9 @@ def expand_domain(domain: AbstractSet[int]) -> set[int]:
 
     def to_pattern(self) -> Pattern:
         """Return the standardized pattern."""
-        pattern = graphix.pattern.Pattern(input_nodes=self.input_nodes)
+        from graphix.pattern import Pattern  # noqa: PLC0415
```
thierry-martinez, line 396:
```suggestion
        from graphix import Pattern  # noqa: PLC0415
```

graphix/opengraph.py
```diff
@@ -532,6 +533,8 @@ def find_pauli_flow(self: OpenGraph[_AM_co], *, stacklevel: int = 1) -> PauliFlo
         >>> str(og.infer_pauli_measurements().extract_pauli_flow())
         'p(0) = {1}, p(1) = {2}; {0, 1} < {2}'
         """
+        from graphix.flow.core import PauliFlow  # noqa: PLC0415
```
```suggestion
        from graphix import PauliFlow  # noqa: PLC0415
```

graphix/opengraph.py
```diff
@@ -480,6 +479,8 @@ def find_gflow(self: OpenGraph[_PM_co]) -> GFlow[_PM_co] | None:
         ----------
         [1] Mitosek and Backens, 2024 (arXiv:2410.23439).
         """
+        from graphix.flow.core import GFlow  # noqa: PLC0415
```
```suggestion
        from graphix import GFlow  # noqa: PLC0415
```

graphix/opengraph.py
```diff
@@ -10,14 +10,13 @@
...
 from graphix.fundamentals import AbstractMeasurement, AbstractPlanarMeasurement
 from graphix.measurements import BlochMeasurement, Measurement
 
 if TYPE_CHECKING:
     from collections.abc import Callable, Collection, Iterable, Mapping, Sequence
 
-    from graphix.flow.core import CausalFlow
+    from graphix.flow.core import CausalFlow, GFlow, PauliFlow
     from graphix.parameter import ExpressionOrSupportsFloat, Parameter
     from graphix.pattern import Pattern
```
```suggestion
    from graphix import CausalFlow, GFlow, Pattern, PauliFlow
    from graphix.parameter import ExpressionOrSupportsFloat, Parameter
```

graphix/__init__.py
emlynsg: I like that approach, it's nice and avoids changing the syntax elsewhere. Committed changes.

graphix/__init__.py
```diff
@@ -2,9 +2,63 @@
 
 from __future__ import annotations
 
+from graphix._version import __version__  # or wherever your version lives
+from graphix.branch_selector import BranchSelector, ConstBranchSelector, FixedBranchSelector, RandomBranchSelector
```
matulni: Do we need to import to the namespace abstract base classes such as `BranchSelector` ? I can see the interest for `NoiseModel` where a user may want to define their own, but not so much for branch selectors.

graphix/__init__.py
```diff
@@ -2,9 +2,63 @@
...
+from graphix.channels import KrausChannel
```
matulni: Do we need `KrausChannel` in the namespace ?

graphix/opengraph.py
```diff
@@ -498,6 +496,8 @@ def find_gflow(self: OpenGraph[_PM_co]) -> GFlow[_PM_co] | None:
         ----------
         [1] Mitosek and Backens, 2024 (arXiv:2410.23439).
         """
+        from graphix import GFlow  # noqa: PLC0415
```
matulni: I guess this is to avoid there's a circular import, but why is there a circular import now ? Before we could import with `from graphix.flow.core import ...`

graphix/instruction.py
```diff
@@ -291,4 +291,35 @@ def visit(self, visitor: InstructionVisitor) -> RZ:
 
 
 InstructionWithoutRZZ = CCX | CNOT | SWAP | CZ | H | S | X | Y | Z | I | M | RX | RY | RZ
```
matulni: Could we put this line inside the `TYPE_CHECKING` block too?

graphix/__init__.py
```diff
@@ -2,9 +2,63 @@
...
+from graphix.clifford import Clifford
+from graphix.command import Command
+from graphix.flow.core import CausalFlow, GFlow, PauliFlow, XZCorrections
+from graphix.fundamentals import ANGLE_PI, Axis, Plane, Sign, angle_to_rad, rad_to_angle
 from graphix.graphsim import GraphState
+from graphix.instruction import Instruction
+from graphix.measurements import BlochMeasurement, Measurement, Outcome, PauliMeasurement
```
matulni: Are we importing type aliases to the namespace such as `Outcome` ? I don't know what is the usual practice, but in any case I think we should be consistent (for instance, `graphix.fundamentals.ParametrizedAngle` is not imported)

graphix/__init__.py
```diff
@@ -2,9 +2,63 @@
...
+from graphix.noise_models import DepolarisingNoiseModel, NoiseModel
```
matulni, line 15: Should we import `NoiselessNoiseModel` and `ComposeNoiseModel` ?

graphix/__init__.py
```diff
@@ -2,9 +2,63 @@
...
+from graphix.opengraph import OpenGraph
+from graphix.parameter import Placeholder
```
matulni, line 18: Should we import `AffineExpression` ?

graphix/__init__.py
```diff
@@ -2,9 +2,63 @@
...
 from graphix.pattern import Pattern
```
matulni: I would also import `DrawAnnotations` but maybe we should give it a more descriptive name.

graphix/__init__.py
```diff
@@ -2,9 +2,63 @@
...
-from graphix.sim.statevec import Statevec
+from graphix.pauli import Pauli
+from graphix.pretty_print import OutputFormat
+from graphix.sim.density_matrix import DensityMatrix, DensityMatrixBackend
+from graphix.sim.statevec import Statevec, StatevectorBackend
```
matulni: They can all four be imported from `graphix.sim`. Should we also import `MBQCTensorNet`, `TensorNetworkBackend` or wait until reworked?

graphix/__init__.py
```diff
@@ -2,9 +2,63 @@
...
+from graphix.states import BasicStates
```
matulni: I would import `PlanarState` too

graphix/__init__.py
```diff
@@ -2,9 +2,63 @@
...
 from graphix.transpiler import Circuit
```
matulni, line 26: I would also import:
- from `graphix.space_minimization`: (`SpaceMinimizationHeuristics`)
- (possibly, but to be discussed) from `graphix.optimization`: (`StandardizedPattern`, `single_qubit_domains`, `remove_useless_domains`, `incorporate_pauli_results`)
- from circ_ext `ExtractionResult`, `PauliString`, `PauliExponential`, `PauliExponentialDAG`, `CliffordMap`, (you may want to add them to `circ_ext.__init__` first).

graphix/command.py
```diff
@@ -237,5 +237,30 @@ class T(_KindChecker, BaseCommand):
...
+        """
+
+        N = N
+        M = M
+        E = E
+        C = C
+        X = X
+        Z = Z
+        S = S
+        T = T
```
matulni: Do we want to expose `T` too?

matulni (commented):
Thanks @emlynsg !
I made some comments, most of them are up to discussion, I'm not very experienced on what the standard practice is for object exposure in python

graphix/__init__.py
matulni: Too niche

graphix/__init__.py
matulni: for internal use only

emlynsg:
@thierry-martinez  @matulni  as per our discussion, the problem with the introduced loops seems like luck. `flow.core` imports `Pattern` imports `optimization` which imports `CausalFlow` from `flow.core`. I think the problem wasn't there before because the top-level imported `Pattern` first, whereas now in alphabetical order it imports `flow.core`.

graphix/flow/core.py
```diff
@@ -15,7 +15,7 @@
 # `override` introduced in Python 3.12, `assert_never` introduced in Python 3.11
 from typing_extensions import assert_never, override
 
-import graphix.pattern
+# `override` introduced in Python 3.12, `assert_never` introduced in Python 3.11
```
thierry-martinez: This comment is a duplicate.

graphix/command.py
```diff
@@ -237,5 +237,30 @@ class T(_KindChecker, BaseCommand):
     kind: ClassVar[Literal[CommandKind.T]] = dataclasses.field(default=CommandKind.T, init=False)
 
 
-Command = N | M | E | C | X | Z | S | T
+if TYPE_CHECKING:
```
thierry-martinez: I didn't find an ideal solution.

What I would propose:
- Define the classes `N`, `M`, ... in a namespace-class `Command`.
- Rename the current type-alias `Command` as `CommandType`. If possible, move its definition in a `TYPE_CHECKING` block.

graphix/command.py
```diff
@@ -237,5 +237,37 @@ class T(_KindChecker, BaseCommand):
...
+
+    N: type[N] = N
+    M: type[M] = M
+    E: type[E] = E
+    C: type[C] = C
+    X: type[X] = X
+    Z: type[Z] = Z
+    S: type[S] = S
+    T: type[T] = T
+    _members: ClassVar[tuple[type, ...]] = (N, M, E, C, X, Z, S, T)
```
thierry-martinez: Why do we need this field?

graphix/command.py
```diff
@@ -237,5 +237,37 @@ class T(_KindChecker, BaseCommand):
...
+    """
+
+    N: type[N] = N
+    M: type[M] = M
+    E: type[E] = E
+    C: type[C] = C
+    X: type[X] = X
+    Z: type[Z] = Z
+    S: type[S] = S
+    T: type[T] = T
```
thierry-martinez: I think it might be better to declare these fields as `TypeAlias` instead of `type[_]`. I believe anything that type-checks with `type[_]` would also work with `TypeAlias`, which, in addition, would allow us to use these aliases in type annotations like this:
```python
def f() -> Command.N:
    return Command.N(1)
```

graphix/command.py
```diff
@@ -237,5 +237,37 @@ class T(_KindChecker, BaseCommand):
     kind: ClassVar[Literal[CommandKind.T]] = dataclasses.field(default=CommandKind.T, init=False)
 
 
-Command = N | M | E | C | X | Z | S | T
+CommandType = N | M | E | C | X | Z | S | T
```
thierry-martinez: I propose to move `CommandType` to `TYPE_CHECKING` block in this commit: https://github.com/TeamGraphix/graphix/commit/91b2dee938cde9ddeba41077db02391c2cde840b

graphix/command.py
emlynsg: I thought it was necessary for the isinstance checks to work, but after checking again it is fine without, so I will remove in another push.

graphix/command.py
emlynsg: Great idea!

graphix/command.py
emlynsg: Thanks Thierry. I guess we will have to make quick PRs to merge the changes into Veriphix for this and the JCZ PR (and the others as we can later on).

thierry-martinez: I opened the following PRs to update the revdeps for namespaces:
- https://github.com/qat-inria/veriphix/pull/20
- https://github.com/thierry-martinez/graphix-stim-backend/pull/4
but I think we can merge this PR independently of those (we can leave the branch pinning in `noxfile.py` for now). I'll leave it to you to create the PR for https://github.com/emlynsg/veriphix/tree/jcz-test-fix .

thierry-martinez (approved):
LGTM! Could you add entries to the changelog regarding the new namespaces, the renaming of `CommandType` and `InstructionType`, and the fact that they are now only defined in the `TYPE_CHECKING` block?

graphix/instruction.py
```diff
@@ -290,36 +290,49 @@ def visit(self, visitor: InstructionVisitor) -> RZ:
...
+class Instruction(InstructionWithoutRZZ):
+    """Grouping of all instructions for namespace exposure.
+
+    Notes
+    -----
+    This class is not meant to be instantiated, but rather serves as a namespace for all instructions except RZZ.
+    The type alias for "any command" is :data:`InstructionKind`.
+    """
+
+    RZZ: type[RZZ] = RZZ
```
matulni: Why not `TypeAlias` here?

matulni (approved):
LGTM! Thanks!!

graphix/instruction.py
emlynsg: Oops!

PR #498, opened 2026-05-05
Fix #168: Remove Pauli measurements

### LINKED ISSUE #168: Support for arbitrary inputs in Pauli preprocessing

thierry-martinez:
This commit implements the algorithm described in [BMBdF+21], Theorem 4.12 in Section 4.3: Removing Clifford vertices.

The new method `Pattern.remove_pauli_measurements` removes almost all Pauli measurements on non-input nodes, and all of them if the pattern has flow, fixing #168.

[BMBdF+21] Miriam Backens, Hector Miller-Bakewell, Giovanni de Felice, Leo Lobski, and John van de Wetering, There and back again: A circuit extraction tale, Quantum, 2021,
https://doi.org/10.22331/q-2021-03-25-421


graphix/optimization.py
```diff
@@ -267,90 +271,13 @@ def perform_pauli_pushing(self, leave_nodes: AbstractSet[Node] | None = None, *,
         Pattern
```
matulni:
```suggestion
        StandardizedPattern
```

graphix/pattern.py
```diff
@@ -1776,6 +1776,48 @@ def perform_pauli_pushing(
...
+        self, *, copy: bool = False, standardize: bool = False, stacklevel: int = 1
+    ) -> Pattern:
+        """Remove non-input Pauli measurements from the given pattern.
+
+        See :func:`~remove_pauli_measurements.remove_pauli_measurements` for more information.
+
+        Parameters
+        ----------
+        pattern: StandardizedPattern
+            Standardized pattern to optimize.
```
```suggestion
```

Also, maybe we should add note in the docstring saying that if this method is applied on non-deterministic pattern, it will return a pattern does is not comparable in any way with the original pattern ? For instance, `Pattern(input_nodes=[], cmds=[N(0), N(1), E((0, 1)), M(0), M(1)]).remove_pauli_measurements(copy=True)` returns an empty pattern.

graphix/pattern.py
```diff
@@ -1776,6 +1776,48 @@ def perform_pauli_pushing(
...
+            If ``True``, the pattern is returned in standardized form.
+            The default is ``False``: the nodes are prepared on a
+            need-by-need basis, minimizing space usage.
+        stacklevel : int, optional
+            Stack level to use for warnings. Defaults to 1, meaning that warnings
+            are reported at this function's call site.
+
+        Returns
+        -------
+        StandardizedPattern
```
```suggestion
        Pattern
```

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,577 @@
...
+    """
+
+    @classmethod
+    def from_standardized_pattern(
+        cls, pattern: StandardizedPattern, leave_nodes: AbstractSet[Node] | None = None, *, stacklevel: int = 1
+    ) -> PauliPushingCut:
+        """Move Pauli measurements before the other measurements and return the cut between Pauli measurements and non-Pauli measurements.
+
+        If you only need the resulting pattern, you can use
+        :meth:`perform_pauli_pushing` instead.
```
matulni:
```suggestion
        :meth:`Pattern.perform_pauli_pushing` instead.
```

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,577 @@
...
+
+        Parameters
+        ----------
+        leave_nodes : AbstractSet[Node], optional
```
matulni, line 89:
```suggestion
        pattern: StandardizedPattern
        leave_nodes : AbstractSet[Node], optional
```

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,577 @@
...
+                        #                = S^s M^{-Z}
+                        shifted_domains[cmd.node] = s_domain
+                    case _:  # pragma: no cover
+                        assert_never(cmd.measurement.axis)
+                pauli_measurements.append(Command.M(node=cmd.node, measurement=cmd.measurement))
+        return cls(pattern, pauli_measurements, non_pauli_measurements, shifted_domains)
+
+    def measurements(self) -> list[Command.M]:
+        """Return the list of measurements, where Pauli measurements appear first and without signal."""
+        return self.pauli_measurements + self.non_pauli_measurements
```
matulni: Somewhat nitpicking, but could this be a property instead of an instance method?
I think it would be more intuitive, but I'm happy if you leave it as is

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,577 @@
...
+            c_dict,
+            z_dict,
+            x_dict,
+        )
+
+
+def _complement_subgraph(graph: nx.Graph[Node], s: set[Node]) -> None:
+    """Complement edges in a given subgraph."""
+    all_pairs = {(u, v) for u, v in itertools.combinations(s, 2)}
+    existing = {edge for edge in all_pairs if edge in graph.edges()}
```
matulni:
```suggestion
    all_pairs = set(itertools.combinations(s, 2))
    existing = all_pairs & graph.edges()
```

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,577 @@
...
+
+    def pivot_vertices(self, u: Node, v: Node) -> None:
+        """
+        Pivot two vertices.
+
+        Prerequisite (not checked):
+        - (u, v) is a graph edge;
+        - u and v are not input nodes.
+
+        Implements Lemmate 2.32 and 4.5 [BMBdF+21].
```
matulni:
```suggestion
        Implements Lemmas 2.32 and 4.5 [BMBdF+21].
```

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,577 @@
...
+        Implements Theorem 4.12, Steps 1 and 2.
+        """
+        for axis, remove in (
+            (Axis.Y, self.remove_y),  # Step 1: remove any non-input Y measured node
+            (Axis.Z, self.remove_z),  # Step 2: remove any non-input Z measured node
+        ):
+            while True:
+                node = next(iter(self.pauli_measurements[axis]), None)
+                if node is None:
+                    break
```
matulni: I think this is equivalent as slightly more readable:
```suggestion
            while self.pauli_measurements[axis]:
                node = next(iter(self.pauli_measurements[axis]))
```

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,577 @@
...
+        -------
+        bool
+            ``True`` if a node has been found and removed, ``False`` otherwise
+        """
+        for node in self.pauli_measurements[Axis.X]:
+            new_node = self.node_map[node]
+            internal_neighbors = set(self.graph.neighbors(new_node)) - self.input_node_set - self.output_node_set
+            if not internal_neighbors:
+                continue
+            v, *_ = internal_neighbors
```
matulni: I suggest this to avoid allocating the list `_`:

```suggestion
            v = next(iter(internal_neighbors))
```

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,577 @@
...
+        -------
+        bool
+            ``True`` if a node has been found and pivoted, ``False`` otherwise
+        """
+        for node in self.pauli_measurements[Axis.X]:
+            new_node = self.node_map[node]
+            non_input_output_nodes = set(self.graph.neighbors(new_node)) & self.output_node_set - self.input_node_set
+            if not non_input_output_nodes:
+                continue
+            v, *_ = non_input_output_nodes
```
matulni:
```suggestion
            v = next(iter(non_input_output_nodes))
```

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,577 @@
...
+            self.pivot_vertices(node, v)
+            return True
+        return False
+
+    def to_standardized_pattern(self) -> StandardizedPattern:
+        output_nodes: list[Node | None] = [None] * len(self.cut.original_pattern.output_nodes)
```
matulni: I think this is more readable, and we avoid one `_filter_none` (see comments below):

```suggestion
        output_nodes = [self.node_map[node] for node in self.cut.original_pattern.output_nodes]
```

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,577 @@
...
+        measurements: list[Command.M | None] = [None] * len(self.measurements)
+        z_dict: dict[Node, set[int]] = {}
+        x_dict: dict[Node, set[int]] = {}
+        c_dict: dict[Node, Clifford] = {}
+        for node, spec in self.node_specs.items():
+            if spec.is_output:
+                output_nodes[spec.index] = node
```
matulni: Can be removed if previous comment is accepted
```suggestion
```

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,577 @@
...
+                    c_dict[node] = spec.clifford
+            else:
+                cmd_m = self.measurements[spec.index].clifford(spec.clifford)
+                cmd_m.node = node
+                cmd_m.s_domain = _map_domain(self.node_map, cmd_m.s_domain)
+                cmd_m.t_domain = _map_domain(self.node_map, cmd_m.t_domain)
+                measurements[spec.index] = cmd_m
+        return StandardizedPattern(
+            self.cut.original_pattern.input_nodes,
+            _filter_none(output_nodes),
```
matulni: If previous comment is accepted:
```suggestion
            output_nodes,
```

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,577 @@
...
+            self.cut.original_pattern.results,
+            (cmd_n for cmd_n in self.cut.original_pattern.n_list if cmd_n.node in self.node_specs),
+            (frozenset(edge) for edge in self.graph.edges()),
```
matulni: This conversion is done by the `StandardizedPattern` constructor, right? No need to iterate twice, then?

```suggestion
            self.graph.edges(),
```

graphix/pattern.py
```diff
@@ -1776,6 +1776,48 @@ def perform_pauli_pushing(
...
+        """
+        from graphix.remove_pauli_measurements import PauliPushingCut, remove_pauli_measurements  # noqa: PLC0415
+
+        standardized_pattern = optimization.StandardizedPattern.from_pattern(self)
+        cut = PauliPushingCut.from_standardized_pattern(standardized_pattern, stacklevel=stacklevel + 1)
+        standardized_pattern = remove_pauli_measurements(cut)
+        pattern = standardized_pattern.to_pattern() if standardize else standardized_pattern.to_space_optimal_pattern()
+        if copy:
+            return pattern
+        self.__seq = pattern.__seq
```
matulni, line 1770: We need to update output nodes in case they have been pivoted (and therefore relabelled). Example:
```python
og = OpenGraph(
        graph=nx.Graph([(1, 2)]),
        input_nodes=[],
        output_nodes=[2],
        measurements={
            1: Measurement.X,
        },
    )
p = og.to_pattern()
p.remove_pauli_measurements()
p.simulate_pattern() # Fails
```

```suggestion
        self.__seq = pattern.__seq
        self.__output_nodes = pattern.output_nodes
```

graphix/pattern.py
matulni, line 1: Should we remove `Pattern.perform_pauli_measurements` and eventually `Pattern.remove_input_nodes`?

tests/test_remove_pauli_measurements.py
```diff
@@ -0,0 +1,270 @@
...
+    standardized_pattern2 = remove_pauli_measurements(cut)
+    assert len(standardized_pattern2.m_list) == 1
+
+
+def test_step_4_no_flow() -> None:
+    pattern = Pattern(input_nodes=(0,), output_nodes=(0,), cmds=[Command.N(1), Command.E((0, 1)), Command.M(1)])
+    standardized_pattern = StandardizedPattern.from_pattern(pattern)
+    cut = PauliPushingCut.from_standardized_pattern(standardized_pattern)
+    standardized_pattern2 = remove_pauli_measurements(cut)
+    assert len(standardized_pattern2.m_list) == 1
```
matulni, line 243: I'm not sure if I understand this test. Why are we supposed to have `len(standardized_pattern2.m_list) == 1` ?
Why `Command.M(1)` is not eliminated ?
If I add `Command.N(0)` to the list, the same test won't pass because `N(1)` is eliminated:

```python
pattern = Pattern(output_nodes=(0,), cmds=[Command.N(0), Command.N(1), Command.E((0, 1)), Command.M(1)])
standardized_pattern = StandardizedPattern.from_pattern(pattern)
cut = PauliPushingCut.from_standardized_pattern(standardized_pattern)
standardized_pattern2 = remove_pauli_measurements(cut)
assert len(standardized_pattern2.m_list) == 1 # Fails
```

matulni (commented):
Thanks a lot for this nice piece of work @thierry-martinez ! Code is super clear and new tool very useful! 

Now that `remove_pauli_nodes` will become a fundamental part of the workflow, I'm even more unhappy that the transformation Pattern -> OpenGraph or Pattern -> XZCorrections does not preserve information about the LC fragment. Maybe it may be worth thinking about this...

graphix/remove_pauli_measurements.py
thierry-martinez: Fixed: I've mentioned `Pattern.perform_pauli_pushing` and `StandardizedPattern.perform_pauli_pushing`.

graphix/remove_pauli_measurements.py
thierry-martinez: Fixed, and incorporated with the `if` above.

graphix/remove_pauli_measurements.py
thierry-martinez: Nice! Indeed, using `node_map` simplifies the reconstruction. I now use the same idea for `measurements`, and fields `index` and `is_output` are now removed.

graphix/pattern.py
thierry-martinez: Nice catch! Fixed and test added.

graphix/remove_pauli_measurements.py
thierry-martinez: Fixed: `measurements` is now a property returning `tuple[Command.M, ...]`, and `pauli_measurements` and `non_pauli_measurements` are now tuples as well. Previously the user could observe that each call returned a new mutable instance.

graphix/remove_pauli_measurements.py
thierry-martinez: Fixed: indeed, expressing the break condition directly in the `while` is preferable. I used the walrus operator to extract the `node` in the condition, if any.

graphix/pattern.py
thierry-martinez: I've replaced every call to `Pattern.perform_pauli_measurements` with a call to `Pattern.remove_pauli_measurements`. I addition, I removed the `Pattern.perform_pauli_measurements` method and the `measure_pauli` function.

However, I believe that the method `Pattern.remove_input_nodes` is still useful, because `remove_pauli_measurements` only acts on non-input Pauli nodes.

Before deleting `perform_pauli_measurements`, I ran a few benchmarks:

```python
def compare_perfs(rng: Generator) -> None:
    nqubits = 10
    depth = 10
    patterns = [rand_circuit(nqubits, depth, rng, use_ccx=True, use_rzz=True).transpile().pattern for _ in range(10)]
    for pattern in patterns:
        pattern.remove_input_nodes()
    print(f"{timeit(lambda: [pattern.copy().remove_pauli_measurements() for pattern in patterns], number=10)=}")
    print(f"{timeit(lambda: [pattern.copy().perform_pauli_measurements() for pattern in patterns], number=10)=}")
```

On my laptop, the results were:
- `remove_pauli_measurements`: **11.7s**
- `perform_pauli_measurements`: **27.9s**

For fairness I copy each pattern before calling the method; using `remove_pauli_measurements(copy=True)` would be even faster.

I also benchmarked the example script `qft_with_tn.py`. The execution times were: 
- `remove_pauli_measurements`: **13.84s**
- `perform_pauli_measurements`: **51.73s**

tests/test_remove_pauli_measurements.py
thierry-martinez: I've updated the docstring of `try_pivot_x_with_output_node` to specify that the `X`-measured node must be connected to an output node **that is _not_ also an input node**.

The test now includes this comment in `test_step_4_no_flow`:
```python
    # This example tests the case of a pattern that contains a
    # non-input X-measured node 1 which is connected to an output node
    # 0, where the node 0 is also an input.  In this situation Lemma
    # 4.11 cannot be applied; this exercices the filtering implemented
    # in the `try_pivot_x_with_output_node` method.
```

graphix/pattern.py
thierry-martinez: I believe that an empty pattern is expected in this case: the original pattern has no output nodes, so the resulting state is the empty state for both the original and the reduced patterns.

More generally, experimentally we observe that reduced patterns return equivalent states when simulated on the 0 branch. This can be explained for Pauli measurements because the XZ-diagram interprets the non-corrected branch. I am not sure whether this is sufficient to justify why the same behavior holds for non-Pauli measurements.

Below is an example (I'm not certain it is worth including it in the test suite yet):
```python
def test_non_deterministic_pattern() -> None:
    pattern = Pattern(
        cmds=[
            Command.N(0),
            Command.N(1),
            Command.N(2),
            Command.N(3),
            Command.E((0, 1)),
            Command.E((0, 2)),
            Command.E((1, 3)),
            Command.M(0),
            Command.M(1, Measurement.XY(ANGLE_PI / 4)),
        ]
    )
    reduced_pattern = pattern.remove_pauli_measurements(copy=True)
    bs = ConstBranchSelector(0)
    state_original = pattern.simulate_pattern(branch_selector=bs)
    state_reduced = reduced_pattern.simulate_pattern(branch_selector=bs)
    assert state_original.isclose(state_reduced)
```

thierry-martinez: I added a note about nondeterministic patterns in 9f59cde8.

tests/test_pattern.py
```diff
@@ -498,22 +427,22 @@ def test_standardize_two_cliffords(self, fx_bg: PCG64, jumps: int) -> None:
...
-        pattern.add(X(node=0, domain={1}))
-        pattern.results[2] = z
-        pattern.add(Z(node=0, domain={2}))
-        pattern.add(C(node=0, clifford=Clifford(c)))
-        pattern_ref = pattern.copy()
-        pattern.standardize()
-        state_ref = pattern_ref.simulate_pattern()
-        state_p = pattern.simulate_pattern()
-        assert state_p.isclose(state_ref)
+    # @pytest.mark.parametrize("jumps", range(1, 48))
```
matulni: This test can be just removed, right ? If is commented out for some reason, I suggest marking it with `pytest.mark.skip`.

CHANGELOG.md
```diff
@@ -11,12 +11,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - #490: Introduced new `Instruction` and `Command` namespace classes for instruction and command instantiation.
 
-- #476 Introduced new methods `OpenGraph.extract_circuit`, `CliffordMap.to_tableau` and new function `graphix.circ_ext.compilation.cm_berg_pass`. Circuit extraction can be done natively in Graphix.
+- #476: Introduced new methods `OpenGraph.extract_circuit`, `CliffordMap.to_tableau` and new function `graphix.circ_ext.compilation.cm_berg_pass`. Circuit extraction can be done natively in Graphix.
 
 ### Fixed
 
 ### Changed
 
+- #168, #498: `Pattern.remove_pauli_measurements` replaces `Pattern.perform_pauli_measurements`.  The new algorithm removes all non-input Pauli nodes from patterns that have a flow and returns a pattern that is equivalent for every input state.
```
matulni: `results` attribute of `Pattern` class is removed.

tests/test_graphsim.py
matulni, line 1: Genuine question: why is this test file removed if we keep `graphsim.py`?

graphix/remove_pauli_measurements.py
```diff
@@ -471,74 +462,78 @@ def try_pivot_x_with_output_node(self) -> bool:
...
             return True
         return False
 
+    def remove_isolated_internal_nodes(self) -> None:
+        """Remove isolated internal nodes."""
+        # Construct the list first since the graph should not be
+        # modified while enumerating isolated nodes.
+        for node in list(nx.isolates(self.graph)):
+            if node not in self.input_node_set and node not in self.output_node_set:
+                self._remove_node(node)
```
matulni: This method unconditionally removes isolated nodes, even if they are not Pauli-measured!

graphix/remove_pauli_measurements.py
```diff
@@ -301,6 +285,8 @@ def _apply_clifford(self, node: Node, clifford: Clifford) -> None:
         if spec.pauli_measurement is not None:
             axis = spec.pauli_measurement.axis
             spec.pauli_measurement = spec.pauli_measurement.clifford(clifford)
+            if node in self.input_node_set:
+                return
```
matulni, line 296: May be worth adding a comment reminding that input nodes are not registered in `pauli_measurements` dictionary.

thierry-martinez: Comment added in 408a379.

tests/test_graphsim.py
thierry-martinez: Oops, I forgot reverting it. Fixed in daf1d45.

CHANGELOG.md
thierry-martinez: Added in ef57707e (and mentioned the removal of `incorporate_pauli_results` as well).

graphix/remove_pauli_measurements.py
thierry-martinez: This reproduces the current behavior of `perform_pauli_measurements`.

Unfortunately, the existing test suite does not contain a test that checks this behavior, but the following test would pass:
```python
def test_isolated_nodes_non_pauli() -> None:
    pattern = Pattern(cmds=[Command.N(0), Command.N(1), Command.M(0), Command.M(1, ANGLE_PI / 4)])
    pattern.perform_pauli_measurements()
    assert list(pattern) == []
```
Note that I had to introduce an additional Pauli node 0 since `perform_pauli_measurements` exits early when there are no Pauli measurements.

I am not sure what the best course of action is. Eliminating isolated nodes seems like a useful optimization, regardless of whether nodes are Pauli or non-Pauli. Moreover, removing Pauli nodes can introduce isolated nodes, and since the procedure already has access to the graph and the logic needed to rebuild the pattern from that graph, it appears to be the appropriate place to perform this cleanup.

That said, I agree that it is somewhat counter-intuitive for the Pauli measurement removal routine to also delete non-Pauli nodes.

tests/test_pattern.py
thierry-martinez: Removed in 5b9c1302.

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,579 @@
...
+            if not isinstance(cmd_m.measurement, PauliMeasurement):  # pragma: no cover
+                msg = "Pauli measurement expected."
+                raise TypeError(msg)
+            self.node_specs[cmd_m.node].pauli_measurement = cmd_m.measurement
+            if cmd_m.node not in self.input_node_set:
+                self.pauli_measurements[cmd_m.measurement.axis].add(cmd_m.node)
+        self.node_map = {node: node for node in self.graph.nodes()}
+
+    def _apply_clifford(self, node: Node, clifford: Clifford) -> None:
+        """Apply a single-qubit Clifford gate to a node.
```
mgarnier59, line 281: This only applies to Pauli measured nodes. If that's how it goes in the algorithm then maybe we need to update the docstring.

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,579 @@
...
+        """
+        n_u = list(self.graph.neighbors(u))
+        _complement_subgraph(self.graph, n_u)
+        # |+⟩⟨+| + exp(-iπ/2) |-⟩⟨-| = H S† H
+        self._apply_clifford(u, Clifford.H @ Clifford.SDG @ Clifford.H)
+        for node in n_u:
+            # |0⟩⟨0| + exp(iπ/2) |1⟩⟨1| = S
+            self._apply_clifford(node, Clifford.S)
+
+    def pivot_vertices(self, u: Node, v: Node) -> None:
```
mgarnier59: the pivot operation is along an edge but do we care?

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,579 @@
...
+        del self.node_map[spec.src]
+        del self.node_specs[u]
+        self.graph.remove_node(u)
+
+    def remove_z(self, u: Node, sign: Sign) -> None:
+        """
+        Remove Z/-Z measurement.
+
+        Prerequisite (not checked):
+        - u measured in Z (sign==PLUS) or -Z (sign=MINUS);
```
mgarnier59: == in minus case

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,579 @@
...
+
+        Prerequisite (not checked):
+        - u measured in X (sign==PLUS) or -X (sign=MINUS);
+        - (u, v) is a graph edge;
+        - u and v are internal nodes.
+
+        Implements Lemma 4.9 [BMBdF+21].
+        """
+        self.pivot_vertices(u, v)
+        self.remove_z(v, sign)
```
mgarnier59, line 415: this is because the vertices have been swapped and that the old u is now the new v ?

graphix/remove_pauli_measurements.py
matulni: Thanks for the clarification, I agree with your comment. I think that having a separate method to remove non-Pauli measurements is too much boiler-plater and unnecessarily complicates the pipeline. Also, if I understand correctly, this is a very niche scenario which appears for non-deterministic patterns. 

How about issuing a warning when this occurs : "Non-Pauli measurement on an isolated node was removed" ?

matulni (approved):
LGTM! 
Thanks for the nice work

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,579 @@
...
+                if spec.pauli_measurement is None:  # pragma: no cover
+                    msg = "Pauli measurement expected."
+                    raise RuntimeError(msg)
+                remove(new_node, spec.pauli_measurement.sign)
+
+    def try_remove_x_with_internal_neighbor(self) -> bool:
+        """
+        Find an X measurement connected to internal neighbor and remove it if any.
+
+        Implements Theorem 4.12, Step 3.
```
mgarnier59: For steps 3 and 4 mention that the `goto` step 1 in handled further below?

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,579 @@
...
+            ``True`` if a node has been found and pivoted, ``False`` otherwise
+        """
+        for node in self.pauli_measurements[Axis.X]:
+            new_node = self.node_map[node]
+            non_input_output_nodes = set(self.graph.neighbors(new_node)) & self.output_node_set - self.input_node_set
+            v = next(iter(non_input_output_nodes), None)
+            if v is None:
+                continue
+            self.pivot_vertices(new_node, v)
+            return True
```
mgarnier59, line 484: Shouldn't there be a removal according to lemma 4.7 (`remoze_z`) ? Or is that taken care of in the removal of isolated nodes?

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,579 @@
...
+            return None
+        spec = self.node_specs[node]
+        new_m = original_m.clifford(spec.clifford)
+        new_m.node = node
+        new_m.s_domain = _map_domain(self.node_map, new_m.s_domain)
+        new_m.t_domain = _map_domain(self.node_map, new_m.t_domain)
+        return new_m
+
+    def to_standardized_pattern(self) -> StandardizedPattern:
+        n_list = tuple(cmd_n for cmd_n in self.cut.original_pattern.n_list if cmd_n.node in self.node_specs)
```
mgarnier59, line 517: do we expect that some nodes are not in `node_specs`?

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,579 @@
...
+        output_nodes = tuple(self.node_map[node] for node in self.cut.original_pattern.output_nodes)
+        measurements = tuple(new_m for original_m in self.measurements if (new_m := self._create_new_m(original_m)))
```
mgarnier59, line 519: are there extra possibilities that new_m can be interpreted as None beyond the explicit return?

graphix/remove_pauli_measurements.py
```diff
@@ -0,0 +1,579 @@
...
+    ----------
+    cut: PauliPushingCut
+        The Pauli-pushed pattern to optimize.
+
+    Returns
+    -------
+    StandardizedPattern
+        The pattern in which Pauli measurements have been removed.
+    """
+    process = _RemovePauliMeasurements(cut)
```
mgarnier59, line 615: I really like these structure becaus eit allows for other optimization processes

mgarnier59 (approved):
Overall it looks very good. The code is extremely clean. I have a couple of questions concerning the exact procedure but it looks ready to merge!

graphix/remove_pauli_measurements.py
mgarnier59: My bad. Cliffords are applied on neighbouring nodes that may not be Pauli.

graphix/remove_pauli_measurements.py
mgarnier59: Yes nodes are removed so don't need to reprepare them

graphix/remove_pauli_measurements.py
mgarnier59: x meas converted to z meas so removed by step 1 in next iteration

graphix/remove_pauli_measurements.py
thierry-martinez: Warning added in a0709c12.

graphix/remove_pauli_measurements.py
thierry-martinez: Renamed into `pivot_edge` in 935a411b.

graphix/remove_pauli_measurements.py
thierry-martinez: Fixed in a79c12a1. Thanks!

graphix/remove_pauli_measurements.py
thierry-martinez: Comment added in a79c12a1.

graphix/remove_pauli_measurements.py
thierry-martinez: Comment fixed in a79c12a1.

graphix/remove_pauli_measurements.py
thierry-martinez: Comment added in a79c12a1.

graphix/remove_pauli_measurements.py
thierry-martinez: Comment added in a79c12a1.

graphix/remove_pauli_measurements.py
thierry-martinez: `_create_new_m` has only two execution paths: one returns `None`, and the other returns an instance of `Command.M`, which always evaluates to `True` when used as a condition.

PR #507, opened 2026-05-13
Conversion methods for `PauliString`

matulni:
This commit introduces new methods for the `PauliString` class:
- `__str__` and `from_str` provide an interface with strings to simplify debugging,
- `to_tableau` and `from_tableau` provide an interface with the `_linalg` module to facilitate implementing algorithms based on the stabilizer formalism. 

Additionally, the static method `PauliString.from_measured_node` is subsumed by the function `extraction_ps_from_corrected_node`. This follows from the observation that the returned Pauli string is truly specific to the circuit extraction algorithm, and it's not the canonical stabilizer given by a flow's correction function. The latter reads

$$ S_i = Z_i K_i = Z_i X_{g(i)\setminus \\{i\\}} Z_{\text{Odd}[g(i)]\setminus \\{i\\}},$$

whereas the `extraction_ps_from_corrected_node` returns a Pauli string acting on the output nodes only. Work on MBQC fidelity shows that `PauliString` objects are useful beyond the circuit extraction routine, so this refactor aims at giving the `PauliString` class more independence.


graphix/circ_ext/extraction.py
```diff
@@ -95,69 +95,176 @@ class PauliString:
...
+
+        Examples
+        --------
+        >>> PauliString.from_str("+XYZ")
+        PauliString(dim=3, axes={0: Axis.X, 1: Axis.Y, 2: Axis.Z}, sign=Sign.PLUS)
+        >>> PauliString.from_str("-IXI")
+        PauliString(dim=3, axes={1: Axis.X}, sign=Sign.MINUS)
+        """
+        if len(ps) < 2:
+            raise ValueError("Input string must have at least 2 characters (a sign followed by operators).")
```
thierry-martinez: Should we disallow empty Pauli strings?
Should the default sign be `+`?

For comparison:
```python
>>> stim.PauliString("X")
stim.PauliString("+X")
>>> stim.PauliString("")
stim.PauliString("+")
```

In addition, `PauliString.from_str(str(PauliString(0, {})))` now raises an exception. I'm not sure whether this behavior is intentional.

graphix/circ_ext/extraction.py
```diff
@@ -95,69 +95,176 @@ class PauliString:
...
 
-        References
+        sign_char, ops = ps[0], ps[1:]  # Mypy disallows string unpacking
```
thierry-martinez: In Python, string unpacking is just a particular case of iterable unpacking. For example,
```python
sign_char, ops = ps[0], ps[1:]
```
uses slicing, so `ops` becomes a string. In contrast,
```python
sign_char, *ops = ps
```
produces a list of characters for `ops`. For the rest of the code, the difference is irrelevant.

If you prefer the more declarative string-unpacking style, you can be explicit about using iterable unpacking; mypy accepts this:
```suggestion
        sign_char, *ops = iter(ps)
```

graphix/circ_ext/extraction.py
```diff
@@ -95,69 +95,176 @@ class PauliString:
...
+
+        _sign_map = {"+": Sign.PLUS, "-": Sign.MINUS}
+        _axis_map = {"X": Axis.X, "Y": Axis.Y, "Z": Axis.Z}
```
thierry-martinez: The class `Axis` itself behaves much like a dictionary: `Axis["X"]` returns `Axis.X`. However, the expression `"X" in Axis` returns `False`. The proper mapping is `Axis.__members__`: `"X" in Axis.__members__` returns `True`.

graphix/circ_ext/extraction.py
```diff
@@ -95,69 +95,176 @@ class PauliString:
...
+            sign=_sign_map[sign_char],
+            dim=len(ops),
+            axes={i: _axis_map[op] for i, op in enumerate(ops) if op != "I"},
+        )
+
+    def __str__(self) -> str:
+        """Return a string representation of the Pauli string."""
+        pauli_str = (
+            str(self.sign),
+            *(getattr(self.axes.get(node), "name", "I") for node in range(self.dim)),
```
thierry-martinez: I would prefer to avoid `getattr` because it isn't type-safe. Moreover, it's clearer to test explicitly whether an axis exists or not rather than relying on the fact that there is no field `name` in `None`.

```suggestion
            *(axis.name if (axis := self.axes.get(node)) else "I" for node in range(self.dim)),
```

graphix/circ_ext/extraction.py
```diff
@@ -386,20 +493,72 @@ def to_tableau(self) -> MatGF2:
                 f"Isometries are not supported yet: # of inputs ({len(self.input_nodes)}) must be equal to the # of outputs ({len(self.output_nodes)})."
             )
 
-        tab = MatGF2(np.zeros((2 * n, 2 * n + 1)))
+        return MatGF2(np.vstack((*(ps.to_tableau() for ps in self.x_map), *(ps.to_tableau() for ps in self.z_map))))
```
```suggestion
        return MatGF2(np.vstack([ps.to_tableau() for psmap in (self.x_map, self.z_map) for ps in psmap]))
```

thierry-martinez (approved):
LGTM! I've added a few comments, but they're mainly nit-picking.

graphix/circ_ext/extraction.py
matulni: Done in [256a303](https://github.com/TeamGraphix/graphix/pull/507/commits/256a30308b49736dbe84e7bfd103deb4fbc2697a)

graphix/circ_ext/extraction.py
matulni: Yes, much cleaner this way. Done in [256a303](https://github.com/TeamGraphix/graphix/pull/507/commits/256a30308b49736dbe84e7bfd103deb4fbc2697a)

graphix/circ_ext/extraction.py
matulni: Thanks, done in [256a303](https://github.com/TeamGraphix/graphix/pull/507/commits/256a30308b49736dbe84e7bfd103deb4fbc2697a)

graphix/circ_ext/extraction.py
matulni: Done in [256a303](https://github.com/TeamGraphix/graphix/pull/507/commits/256a30308b49736dbe84e7bfd103deb4fbc2697a)

graphix/circ_ext/extraction.py
matulni: Nice catch, I missed the zero-dimensional case. I agree is best to be compatible with `stim`. Done in [256a303](https://github.com/TeamGraphix/graphix/pull/507/commits/256a30308b49736dbe84e7bfd103deb4fbc2697a)

pranav97nair (approved):
Nice work, thanks!

PR #511, opened 2026-05-19
Fix #468: Remove pyzx interface

thierry-martinez:
Module pyzx is removed from Graphix and moved in a separate plugin as we do for other optional dependencies.

https://github.com/thierry-martinez/graphix-pyzx/

See related PR:
https://github.com/thierry-martinez/graphix-pyzx/pull/1


matulni (approved):
Thanks, LGTM!

Some related comment (but outside the scope of this PR): plugins are scattered around three different pages (TeamGraphix, qat-inria, and thierry-martinez). It may be worth eventually to centralise all official plugins (i.e., included in the reverse-dependency check and released) in the same page, but of course not a priority.

pranav97nair (approved):
LGTM, thanks Thierry!

PR #515, opened 2026-05-26
Transpiler from MBQC+LC into pure MBQC

thierry-martinez:
This commit implements:

- `Clifford.to_opengraph`: returns an equivalent MBQC pattern (i.e., without local Clifford corrections) for a given single-qubit Clifford gate.

- `remove_local_clifford_commands`: transpiles an MBQC+LC pattern into an equivalent pure MBQC pattern by eliminating all local-Clifford commands.

- `Pattern.reindex` and `Command.reindex`: implements the renumbering step of pattern composition.


graphix/command.py
```diff
@@ -70,6 +79,10 @@ class BaseN(BaseCommand):
     node: int
     kind: ClassVar[Literal[CommandKind.N]] = dataclasses.field(default=CommandKind.N, init=False)
 
+    @override
+    def reindex(self, f: Callable[[Node], Node]) -> BaseN:
+        return BaseN(f(self.node))
```
matulni: I suggest this change to avoid reimplementing `N.reindex`:

```suggestion
    def reindex(self, f: Callable[[Node], Node]) -> Self:
        return dataclasses.replace(self, node=f(self.node))
```

graphix/optimization.py
```diff
@@ -614,3 +614,42 @@ def decompose_domain(
...
+            case CommandKind.C:
+                cmd_node = cmd.node
+                clifford_pattern = cmd.clifford.to_opengraph().to_pattern()
+                (output_node,) = clifford_pattern.output_nodes
+                # We avoid using `new_pattern.compose` here because
+                # pattern composition is linear in the size of each
+                # pattern, which would make transpilation run in
+                # quadratic time.
+                new_pattern.extend(clifford_pattern.reindex(lambda node: cmd_node if node == 0 else node + max_node))  # noqa: B023
+                max_node += output_node
```
matulni, line 646: I think this relies on the fact that `output_node == max(clifford_pattern.extract_nodes())` by construction. It may be worth a comment.

graphix/optimization.py
```diff
@@ -614,3 +614,42 @@ def decompose_domain(
...
+        match cmd.kind:
+            case CommandKind.C:
+                cmd_node = cmd.node
+                clifford_pattern = cmd.clifford.to_opengraph().to_pattern()
+                (output_node,) = clifford_pattern.output_nodes
+                # We avoid using `new_pattern.compose` here because
+                # pattern composition is linear in the size of each
+                # pattern, which would make transpilation run in
+                # quadratic time.
+                new_pattern.extend(clifford_pattern.reindex(lambda node: cmd_node if node == 0 else node + max_node))  # noqa: B023
```
matulni, line 645: I think the remapping function assumes that the input node is 0. This is true by construction of the `Clifford.to_opengraph` method, but it may be worth a comment.

graphix/optimization.py
```diff
@@ -614,3 +614,42 @@ def decompose_domain(
...
+                max_node += output_node
+                mapping[cmd.node] = max_node
+            case CommandKind.M | CommandKind.X | CommandKind.Z | CommandKind.S:
+                new_cmd = cmd.reindex(reindex)
+                new_pattern.add(new_cmd)
+            case _:
+                new_pattern.add(cmd)
```
matulni: I think `C` and `E` commands should also be remap. 

The following code fails:

```python
from graphix.optimization import remove_local_clifford_commands
from graphix import Pattern, Clifford
from graphix.command import N, C, E

p =  Pattern(input_nodes=[], cmds=[N(0), N(1), C(0, clifford=Clifford.H), E((0, 1))])
p.check_runnability()

p2 = remove_local_clifford_commands(p)
p2.check_runnability() # RunnabilityError: E((0, 1)): node 0 is already measured.
```

In general, is there any circumstance under which we wouldn't want to remap a command ?
If someone added a new command kind (for some obscure reason) this function would inadvertently become  outdated.
I think we should do something like this:

```suggestion
            case CommandKind.N:
                pass
            case _:
                new_cmd = cmd.reindex(reindex)
                new_pattern.add(new_cmd)
```
 because `cmd.reindex` is an abstract method of `BaseCommand`. The `case CommandKind.N` is just for code readability, since if we find a command `N(i)` , `i` will never in `mapping` in a runnable pattern.

Note, however, that this is not enough if one node of the `E` command acts on an input node. The following code fails with the fix I suggest because we need to remap the input nodes as well:

```python
from graphix.optimization import remove_local_clifford_commands
from graphix import Pattern, Clifford
from graphix.command import N, C, E

p =  Pattern(input_nodes=[1], cmds=[N(0), N(1), C(0, clifford=Clifford.H), E((0, 1))])
p.check_runnability()

p2 = remove_local_clifford_commands(p)
p2.check_runnability() # RunnabilityError: N(1): node 1 is already active.
```

graphix/optimization.py
```diff
@@ -614,3 +614,42 @@ def decompose_domain(
 
     new_pattern.reorder_output_nodes(pattern.output_nodes)
     return new_pattern
+
+
+def remove_local_clifford_commands(pattern: Pattern) -> Pattern:
```
matulni, line 615: Would it make sense to have this as a method of `Pattern` that can act in place (like `remove_pauli_measurements`) ? I'm not sure what a good rule is to determine what should be methods and what should be just plain functions.
In any case, I'm happy with keeping the logic here.

graphix/optimization.py
```diff
@@ -614,3 +614,42 @@ def decompose_domain(
...
+                # quadratic time.
+                new_pattern.extend(clifford_pattern.reindex(lambda node: cmd_node if node == 0 else node + max_node))  # noqa: B023
+                max_node += output_node
+                mapping[cmd.node] = max_node
+            case CommandKind.M | CommandKind.X | CommandKind.Z | CommandKind.S:
+                new_cmd = cmd.reindex(reindex)
+                new_pattern.add(new_cmd)
+            case _:
+                new_pattern.add(cmd)
+    new_pattern.reorder_output_nodes(map(reindex, pattern.output_nodes))
```
matulni, line 651: If two Clifford act on the same node and they have not been combined into a single Clifford, we have to take care of removing the outputs of the intermediate Cliffords.
The following code fails:

```python
from graphix.optimization import remove_local_clifford_commands
from graphix import Pattern, Clifford
from graphix.command import N, C

p =  Pattern(input_nodes=[], cmds=[N(0), C(0, clifford=Clifford.H), C(0, clifford=Clifford.H)])
p.check_runnability()

p2 = remove_local_clifford_commands(p)
```

tests/test_optimization.py
```diff
@@ -131,3 +131,20 @@ def test_bug_482() -> None:
     )
     output_pattern = StandardizedPattern.from_pattern(input_pattern).to_space_optimal_pattern()
     assert input_pattern.output_nodes == output_pattern.output_nodes
+
+
+@pytest.mark.parametrize("jumps", range(1, 11))
+def test_remove_local_clifford_commands(fx_bg: PCG64, jumps: int) -> None:
```
matulni, line 140: I suggest adding this test to cover the issues I mentioned above:
```python
@pytest.mark.parametrize("clifford", Clifford)
def test_remove_local_clifford_commands_interleaved(fx_rng: Generator, clifford: Clifford) -> None:
    pattern = Pattern(input_nodes=[1], cmds=[N(0), C(0, clifford=clifford), E((0, 1)), C(0, clifford=clifford)])
    new_pattern = remove_local_clifford_commands(pattern)
    assert not any(cmd.kind == CommandKind.C for cmd in new_pattern)
    input_state = rand_state_vector(1, rng=fx_rng)
    state_ref = pattern.simulate_pattern(input_state=input_state, rng=fx_rng)
    state = new_pattern.simulate_pattern(input_state=input_state, rng=fx_rng)
    assert state.isclose(state_ref)
```

graphix/_db.py
```diff
@@ -226,3 +227,31 @@ class _CMTuple(NamedTuple):
     ("h", "x", "sdg"),
     ("h", "x", "s"),
 )
+
+CLIFFORD_PAULI_DECOMPOSITION = (
```
matulni, line 231: I'm curious, how did you derive this equivalence, with the transpiler ?

Is there a reason why we hard-code the measurements? Couldn't we instead hard-code `StandardizedPattern` objects instead?
The transformations `StandardizedPattern` -> `Pattern` or `StandardizedPattern` -> `OpenGraph` are linear in the number of commands whereas `OpenGraph` -> `Pattern` is quadratic (because of the causal flow find).

I'm not sure if this would matter in any context, since we have 4 nodes at most. Probably your version is more readable.

matulni (commented):
Thanks! I made some comments

graphix/command.py
thierry-martinez: Fixed in 225d0db.

graphix/optimization.py
thierry-martinez: Comments added in e1ed319bf.

graphix/optimization.py
thierry-martinez: Comments added in e1ed319bf.

graphix/optimization.py
thierry-martinez: Indeed, we have to reindex every command (including `C` themselves), and this is fixed in e1ed319bf. However, I don't understand your last example: there is already a `RunnabilityError` raised by `p.check_runnability()` because input nodes cannot be prepared.

graphix/optimization.py
thierry-martinez: I've added a method `Pattern.remove_local_clifford_commands` that acts in place. Usually, I prefer algorithms to be implemented in their own functions or classes, ideally in a dedicated module, and with convenience methods on `Pattern` to call them, rather than cramming everything into one large`Pattern` class. That said, the module `optimization` is becoming too big itself, and we could consider splitting it by a finer-grained groups; however, I think that would deserve a separate PR.

graphix/optimization.py
thierry-martinez: Fixed in e1ed319: the `node` field in `C` commands has to be re-indexed as well.

tests/test_optimization.py
thierry-martinez: I added more tests in `test_remove_local_clifford_commands_edge_cases`.

graphix/_db.py
thierry-martinez: I derived them by exhaustive search with the code below, that searches for the shortest sequence of Pauli measurements that implements each single-qubit Clifford gate.

I think that hard-coding only measurements is more compact, because we can derive the corresponding patterns from them. Hard-coding the full patterns would be a bit more efficient, since we would not have to recompute them at runtime. In practice, for auto-generated code in the code base I tend to value compactness, and I do not agree with the time-complexity argument: as you noted, the size is fixed (at most 4 nodes), so the overhead is just a constant multiplicative factor.

```python
from itertools import product

import networkx as nx
from graphix import Axis, Clifford, Sign, Command, PauliMeasurement, Pattern, OpenGraph
from graphix.opengraph import OpenGraphError
from graphix.random_objects import rand_state_vector
from numpy.random import default_rng

rng = default_rng()

pauli_measurements = [PauliMeasurement(axis, sign) for axis in Axis for sign in Sign]

input_states = [rand_state_vector(nqubits=1, rng=rng) for _ in range(10)]

all_output_states_ref = [
    [
        Pattern(input_nodes=[0], cmds=[Command.C(0, clifford)]).simulate_pattern(
            input_state=input_state
        )
        for input_state in input_states
    ]
    for clifford in Clifford
]

patterns = [None] * len(Clifford)


def explore(n):
    graph = nx.path_graph(n + 1)
    for measurement_list in product(pauli_measurements, repeat=n):
        measurements = dict(zip(range(n), measurement_list))
        og = OpenGraph(
            graph=graph, input_nodes=[0], output_nodes=[n], measurements=measurements
        )
        try:
            pattern = og.to_pattern()
        except OpenGraphError:
            continue
        for clifford, output_states_ref in zip(Clifford, all_output_states_ref):
            if patterns[clifford.value] is not None:
                continue
            if all(
                pattern.simulate_pattern(input_state=input_state, rng=rng).isclose(
                    output_state_ref
                )
                for input_state, output_state_ref in zip(
                    input_states, output_states_ref
                )
            ):
                patterns[clifford.value] = measurement_list


n = 0
explore(0)
while any(pattern is None for pattern in patterns):
    n += 1
    explore(n)

print(patterns)
```

graphix/optimization.py
matulni: Sorry, I made an error copy-pasting the example. I meant:

```python
from graphix.optimization import remove_local_clifford_commands
from graphix import Pattern, Clifford
from graphix.command import N, C, E

p =  Pattern(input_nodes=[1], cmds=[N(0), C(0, clifford=Clifford.H), E((0, 1))])
p.check_runnability()

p2 = remove_local_clifford_commands(p)
p2.check_runnability() # RunnabilityError: E((0, 1)): node 0 is already measured.
```

I believe this occurs because input nodes have not been remap (I didn't have time to check after your last edits).

thierry-martinez: Thanks! The problem was that, in your proposed fix, the case for the `N` command simply performed a `pass` instead of adding the unmodified `N` command to `new_pattern`. Removing the special-case handling for `N` resolves the issue.

graphix/_db.py
thierry-martinez: I added the code in `test_db.generate_clifford_pauli_decomposition` in 1baf792.

graphix/pattern.py
```diff
@@ -162,13 +162,47 @@ def replace(self, cmds: list[CommandType], input_nodes: list[int] | None = None)
...
+            that changes are performed in place.
+
+        Returns
+        -------
+        Pattern
+            The re-indexed pattern. Equal to ``self`` if ``copy`` is ``False``.
+        """
+        if f is None:
+            # Suggested in issue #519
+            f = {node: i for i, node in enumerate(sorted(self.extract_nodes()))}
```
matulni: I have two suggestions:
- Allow to pass a positive integer, so that fresh node labels are generated starting from the passed integer instead of always 0.  We just have to accept an `int` for `f`, and do `node: i + f` in the dict comprehension. 
- Decouple the remapping of nodes (managed by `f`) from assigning sequential labels to nodes starting from a given integer. The goal would be to combine both possibilities: remap some nodes and assign fresh sequential labels to other nodes starting from a given integer. In other words, support non-injective `f`. This is tricky if `f` is a `Callable`, so maybe it's outside the scope of this PR.

The motivation for both suggestions if to offload the work done by `compose` to this new method.

tests/test_db.py
```diff
@@ -58,3 +71,63 @@ def test_safety(self, i: int) -> None:
...
+                if all(
+                    pattern.simulate_pattern(input_state=input_state, rng=rng).isclose(output_state_ref)
+                    for input_state, output_state_ref in zip(input_states, output_states_ref, strict=True)
+                ):
+                    patterns[clifford.value] = measurement_list
+
+    n = 0
+    explore(0)
+    while any(pattern is None for pattern in patterns):
+        n += 1
```
matulni: Very nice. For clarity, and to avoid an infinite loop should we introduce a bug somewhere, I'd add:
```python
if n > 4:
    raise RuntimeError("Local Cliffords are guaranteed to have a decomposition in 3 or less measured nodes.")
```

tests/test_command.py
```diff
@@ -1 +1,29 @@
...
+    assert Command.E((2, 0)).reindex(reindex) == Command.E((2, 1))
+    assert Command.C(0, Clifford.H).reindex(reindex) == Command.C(1, Clifford.H)
+    assert Command.C(2, Clifford.S).reindex(reindex) == Command.C(2, Clifford.S)
+    assert Command.X(0, {2}).reindex(reindex) == Command.X(1, {2})
+    assert Command.X(2, {0}).reindex(reindex) == Command.X(2, {1})
+    assert Command.Z(0, {2}).reindex(reindex) == Command.Z(1, {2})
+    assert Command.Z(2, {0}).reindex(reindex) == Command.Z(2, {1})
+    assert Command.S(0, {2}).reindex(reindex) == Command.S(1, {2})
+    assert Command.S(2, {0}).reindex(reindex) == Command.S(2, {1})
+    assert Command.T().reindex(reindex) == Command.T()
```
matulni:
```python

import pytest

from graphix import Clifford
from graphix.command import BaseM, BaseN, Command

if TYPE_CHECKING:
    from graphix.command import BaseCommand


@pytest.mark.parametrize(
    ("cmd1", "cmd2"),
    [
        (BaseN(0), BaseN(1)),
        (BaseN(2), BaseN(2)),
        (Command.N(0), Command.N(1)),
        (Command.N(2), Command.N(2)),
        (BaseM(0), BaseM(1)),
        (BaseM(2), BaseM(2)),
        (Command.M(0), Command.M(1)),
        (Command.M(2), Command.M(2)),
        (Command.E((0, 2)), Command.E((1, 2))),
        (Command.E((2, 0)), Command.E((2, 1))),
        (Command.C(0, Clifford.H), Command.C(1, Clifford.H)),
        (Command.C(2, Clifford.S), Command.C(2, Clifford.S)),
        (Command.X(0, {2}), Command.X(1, {2})),
        (Command.X(2, {0}), Command.X(2, {1})),
        (Command.Z(0, {2}), Command.Z(1, {2})),
        (Command.Z(2, {0}), Command.Z(2, {1})),
        (Command.S(0, {2}), Command.S(1, {2})),
        (Command.S(2, {0}), Command.S(2, {1})),
        (Command.T(), Command.T()),
    ],
)
def test_reindex(cmd1: BaseCommand, cmd2: BaseCommand) -> None:
    def reindex(node: int) -> int:
        return 1 if node == 0 else node

    assert cmd1.reindex(reindex) == cmd2
```

matulni (approved):
Thanks LGTM! I made some small comments only

tests/test_db.py
thierry-martinez: Fixed in bfcdf0e (replaced `while` by `for`-loop).

tests/test_command.py
thierry-martinez: Thanks. Integrated in 4ea0cd5.

graphix/pattern.py
thierry-martinez: This goes a bit beyond the scope of this PR, but it fits something I wanted to implement, so I've proposed a change. I hope you'll agree to review it in the context of this PR.

In commit f138a93 I introduced a new method `Pattern.node_mapping` that addresses your suggestions:
- The method has an optional argument `start` that lets you choose the starting index for fresh nodes.
- It also accepts an optional initial `mapping`, which fixes the new indices for a subset of nodes, and an optional `preserves` argument for nodes that should not be remapped, and an optional `avoids` argument for indices that should not be used for fresh nodes.

Providing this functionality in a separate method is useful because it gives users a convenient way to obtain the actual mapping as well as the re-indexed pattern. This feels more natural than returning a pair, which is what `compose` currently does.

graphix/pattern.py
```diff
@@ -162,7 +165,66 @@ def replace(self, cmds: list[CommandType], input_nodes: list[int] | None = None)
...
+        Parameters
+        ----------
+        mapping: Mapping[Node, Node] | None, optional
+            An initial (partial) mapping that will be extended.
+            A ``ValueError`` is raised if the supplied mapping is not injective.
+        start: int, optional
+            The starting index from which fresh indices are allocated.
+        avoids: Iterable[Node] | None, optional
+            Indices that must not be assigned to a node.
+        preserves: AbstractSet[Node] | None, optional
```
matulni: Is there a reason why these two fields are not annotated with the same type ? I think the API would be more intuitive if we promote ``preserves`` to ``Iterable`` and cast it to set as we do for ``avoids``.

And related question for my understanding: is ``Iterable`` the type we want? Shouldn't it be ``Collection`` ?
If we pass something like ``itertools.count()``, the code would break but the type checker wouldn't complain.

graphix/pattern.py
```diff
@@ -162,7 +165,66 @@ def replace(self, cmds: list[CommandType], input_nodes: list[int] | None = None)
...
+            if duplicates:
+                reason = ", ".join(f"{antecedents} are mapped to {value}" for value, antecedents in duplicates.items())
+                raise ValueError(f"Initial mapping is not injective: {reason}")
+        used_indices = set(avoids or ())
+        used_indices.update(result.values())
+        if preserves:
+            used_indices.update(preserves)
+        candidate = start
+        for node in sorted(nodes):
+            if node in result or (preserves and node in preserves):
```
matulni: Nit: if we cast preserves to `set` like you do for `avoids` (as suggested in prev. comment) we can just do `node in preserves`

graphix/pattern.py
```diff
@@ -162,7 +165,66 @@ def replace(self, cmds: list[CommandType], input_nodes: list[int] | None = None)
...
+        ----------
+        mapping: Mapping[Node, Node] | None, optional
+            An initial (partial) mapping that will be extended.
+            A ``ValueError`` is raised if the supplied mapping is not injective.
+        start: int, optional
+            The starting index from which fresh indices are allocated.
+        avoids: Iterable[Node] | None, optional
+            Indices that must not be assigned to a node.
+        preserves: AbstractSet[Node] | None, optional
+            Nodes that should not appear in the mapping.  Their index will not be changed by :meth:`reindex`.
```
matulni: Suggest adding `Notes` section:
"In case of conflict between ``mapping`` and ``avoids`` or ``preserves``, elements of ``mapping`` will prevail in the resulting dictionary".

graphix/pattern.py
```diff
@@ -1855,3 +1964,60 @@ def shift_outcomes(outcomes: Mapping[int, Outcome], signal_dict: Mapping[int, Ab
...
+
+    Returns
+    -------
+    dict[V, list[K]]
+        A dictionary whose keys are the duplicated values and whose values are
+        lists of the keys that map to them.  If the input mapping is injective,
+        an empty dictionary is returned.
+    """
+    reversed_mapping = _reverse_mapping(m)
+    return {v: antedents for v, antedents in reversed_mapping.items() if len(antedents) > 1}
```
```suggestion
    return {v: antecedents for v, antecedents in reversed_mapping.items() if len(antecedents) > 1}
```

matulni (approved):
Very nice, LGTM!

graphix/pattern.py
thierry-martinez: Notes added in d44fc45.

graphix/pattern.py
thierry-martinez: Indeed, I switched to `Collection` for both `preserves` and `avoids` in d44fc45.

graphix/command.py
```diff
@@ -101,6 +114,10 @@ class `M`, with given plane, angles, and domains. The base class `BaseM` allows
     node: Node
     kind: ClassVar[Literal[CommandKind.M]] = dataclasses.field(default=CommandKind.M, init=False)
 
+    @override
+    def reindex(self, f: Callable[[Node], Node]) -> BaseM:
+        return BaseM(f(self.node))
```
pranav97nair: Is there a reason why we use `dataclasses.replace` in `BaseN` but we directly instantiate a new `BaseM` object here?

tests/test_db.py
```diff
@@ -58,3 +71,63 @@ def test_safety(self, i: int) -> None:
         with pytest.raises(ValueError):
             # Cannot create writeable view
             v.flags.writeable = True
+
+
+def generate_clifford_pauli_decomposition(rng: Generator) -> tuple[tuple[PauliMeasurement, ...], ...]:
+    """Compute the value of CLIFFORD_PAULI_DECOMPOSITION.
+
+    This function ensures that the length of the decomposition is
+    optimal by search exhaustively by increasing length.
```
pranav97nair:
```suggestion
    optimal by searching exhaustively for increasing length.
```

pranav97nair (commented):
Nice work, Thierry! I approve with just a couple very minor comments.

graphix/command.py
thierry-martinez: Fixed in 9ffb158.

pranav97nair (approved):

PR #516, opened 2026-05-27
Take `enable-cache` and `python-version` into account when setting up `uv` in CI

thierry-martinez:
The `setup-uv` composite action ignored the `enable-cache` and `python-version` inputs. Cache was still enabled by default, which is the intended behaviour, but the system Python version was used instead of the one specified. As a result, type‑checking and coverage ran with Python 3.12 rather than the requested Python 3.14.

This commit fixes the `setup-uv` composite action so that it respects both the `enable-cache` and `python-version` inputs.

### DIFF
--- a/.github/actions/setup-uv/action.yml
+++ b/.github/actions/setup-uv/action.yml
@@ -1,6 +1,14 @@
 name: Set up uv
 description: Install the pinned version of uv and cache its package cache
-
+inputs:
+  enable-cache:
+    description: Enable cache
+    required: false
+    type: boolean
+  python-version:
+    description: Python version
+    required: false
+    type: string
 runs:
   using: composite
   steps:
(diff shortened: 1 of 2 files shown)


thierry-martinez:
- The use of Python 3.12 for type-checking was noticed during the review of #512 (see comment https://github.com/TeamGraphix/graphix/pull/512#discussion_r3302942519).
- In the type-checking workflow, `python-version` was set to 3.13 (and was ignored) with an outdated comment about Qiskit/qiskit-aer#2378, which is now closed (see comment https://github.com/thierry-martinez/graphix-pyzx/pull/1#discussion_r3309462656). The `python-version` field has been updated to 3.14.

matulni (approved):
Thanks LGTM!

PR #518, opened 2026-06-01
Numba-jit statevector backend

matulni:
This PR replaces the statevector backend with a new version orders of magnitude more efficient. It is largely inspired from Ref. [1].

Key features:

- Functions modifying the quantum state (exponentially expensive) are jit compiled with numba.

- Modifications of the quantum state are done in-place so data is not unnecessarily copied during the simulation. To that purpose, `StatevectorBackend` has a new class method constructor `with_capacity` which preallocates a given space in memory. Typically, we pass it a parameter `n = Pattern.max_space()` to allocate a complex-type array with size `2**n`. The `PatternSimualator` class handles this internally, so the API for pattern or circuit simulation remains the same. 

- Adaptive parallelization: jit-compiled functions are parallelized for qubit counts larger than ``NUM_QUBIT_PARALLEL``. This compile constant (empirically determined to be 15), is the number above which the multi-thread overhead does not compensate the speed gains.

In addition:

- Support for symbolic simulation with the statevector backend is moved to the `graphix-symbolic` plugin. See accompanying PR: https://github.com/TeamGraphix/graphix-symbolic/pull/9
- Parameter names of methods in `DenseState` are modified to follow a consistent naming. This effectively changes the public API.
- Test suit is extended and  files `test_statevec.py` and `test_statevectorbackend.py` are unified in a single file.

## Discussion

Below I show the profiling of simulating a QFT circuit on 23 qubits (from the Munich Benchmark suite) . With the current transpiler, this pattern has ``max_space=24`` and the following command count: ``{'N': 2523, 'E': 3023, 'M': 2523, 'Z': 23, 'X': 23}``:

image

Execution time is dominated by the measurement process (in particular, `remove_qubit` subcall), so further optimization efforts should focus on this part of the simulation pipeline.
Current implementation is probably as good as we can do with numba. I recommend reading this very insighful thread: https://stackoverflow.com/questions/79948374/improving-efficiency-of-numba-jit-function (credits to Jérôme Richard @zephyr111).

If we want to push this further, it may be worth to write specialized code for measurements on the XY, XZ and YZ planes. Currently, measurement needs four passes over ``psi`` (see method `graphix.sim.base_backend.DenseStateBackend.measure`):
1) Compute expectation value of projector (``expectation_single``).
2) Evolve the statevector with the appropriate projector (``evolve_single``).
3) Compute norm of 0 and 1 branches of the resulting statevector (``remove_qubit``).
4) Normalize the selected statevector (``remove_qubit``).

Step 1) is not necessary if we use the constant branch selector (this is legit for deterministic patterns, and btw, the pattern above runs in 78.9s with the 0-branch selector).
I suspect it is possible to merge steps 2) and 3) in a single pass which may give a ~30% speed improvement.

P.S. Comments by @thierry-martinez  in [the preliminary discussion](https://github.com/QatQode/graphix-statevec-jit/pull/1) have been addressed.

### References

[1] McGuffin, M. J., Robert J-M., and Ikeda K. "How to Write a Simulator for Quantum Circuits from Scratch: A Tutorial.", 2025 (arXiv:2506.08142).

## UPDATE (01/06/2026)

In commit [3ddc462](https://github.com/TeamGraphix/graphix/pull/518/commits/3ddc462d01cb0908db7ff6acef3417061c39fb97) I reimplemented the method `StatevectorBackend.measure` to merge ``evolve_single`` and ``remove_qubit`` into a single call ``project_qubit`` as discussed above. Effectively, we are fusing two loops over ``psi`` which is advantageous in memory-bound programs. New implementation of qubit measurement is about a 23% faster:

image

In the new implementation of the qubit-removal (step 3), we compute the squared norm of the two subvectors in the same loop. This is not the most efficient choice, since most of the times we only need the norm of one subvector. However, I observed that the difference is negligible in the example with 24 qubits, and calculating it separately leads to much more code repetition. Note that in both cases, we have to read all elements of ``psi`` to apply the projector operator which is probably more costly than doing the arithmetic operation to compute the squared norm.


tests/test_tnsim.py
```diff
@@ -402,8 +402,8 @@ def test_evolve(self, fx_bg: PCG64, jumps: int, fx_rng: Generator) -> None:
         pattern.remove_pauli_measurements()
         state = circuit.simulate_statevector().statevec
         tn_mbqc = pattern.simulate_pattern(backend="tensornetwork", rng=fx_rng)
-        random_op3 = random_op(3, rng)
-        random_op3_exp = random_op(3, rng)
+        random_op3 = rand_unit(2**3, rng)
+        random_op3_exp = rand_unit(2**3, rng)
```
thierry-martinez, line 406: Could you explain this change? Could we systematically replace `random_op(n, rng)` by `rand_unit(2**n, rng)` and remove `random_op` definition?

graphix/sim/statevec.py
```diff
@@ -1,91 +1,143 @@
...
+            self._psi = data._psi.copy()
+            self._max_qubits = data.max_qubits
+            self._nqubit = data.nqubit
+
+            if max_qubits is not None:
+                if max_qubits < data.max_qubits:
+                    raise ValueError(
+                        f"`max_qubits` can't be smaller than the capacity of input state: {max_qubits} < {data.max_qubits}."
+                    )
+                self.ensure_capacity(max_qubits)
```
thierry-martinez: The implemented behavior is:
- Copy `data._psi` as‑is.
- Require the user‑supplied `max_qubits` to be greater than `data.max_qubits`.

I propose the following, which I think is both more useful and more intuitive:
- Require the user‑supplied `max_qubits` to be greater than `data.nqubit`.
- Truncate `data._psi` at the user‑supplied `max_qubits` when necessary.

graphix/sim/statevec.py
```diff
@@ -172,63 +241,213 @@ def add_nodes(self, nqubit: int, data: Data) -> None:
...
+
+        Parameters
+        ----------
+        required_qubits : int
+            Minimum number of qubits the state vector must support. If expansion
+            is needed, ``self._psi`` is extended to size ``2**required_qubits``.
+        """
+        if required_qubits > self.max_qubits:
+            offset = (1 << required_qubits) - len(self._psi)
+            self._psi = np.concatenate([self._psi, np.empty(offset, dtype=self._psi.dtype)])
```
thierry-martinez: I've only performed micro-benchmarks but it seems much more efficient:
```suggestion
            new_psi = np.empty(1 << required_qubits, dtype=self._psi.dtype)
            new_psi[:len(self._psi)] = self._psi
            self._psi = new_psi
```

tests/test_tnsim.py
matulni: In the old API, `Statevec.expectation_value` was normalizing the statevector before computing the norm. Now, for efficiency, we don't normalize the statevector, because we _assume_ that the statevector always has norm 1. This is guaranteed upon instantianting a new `Statevec` object, and requires that `Statevec.evolve` only takes unitary operators, otherwise the norm is not conserved (same applies to `Statevec.evolve_single`, but not relevant here). I think this restriction is reasonable, as we expect gates to be unitary operations.

In general, `Statevec.expectation_value` doesn't need to be restricted to unitary operators, so I think we can keep `random_op` for the other instances.

graphix/sim/statevec.py
matulni: Done in [696e2a2](https://github.com/TeamGraphix/graphix/pull/518/commits/696e2a204abd827443369bf34b1becc1d5323309)
Makes sense, we allocate once instead of twice.

graphix/sim/statevec.py
matulni: Done in [696e2a2](https://github.com/TeamGraphix/graphix/pull/518/commits/696e2a204abd827443369bf34b1becc1d5323309) and added tests.

thierry-martinez (approved):
LGTM! Thanks.

graphix/sim/density_matrix.py
```diff
@@ -32,6 +32,19 @@
...
+)
+CNOT_TENSOR = np.array(
+    [[[[1, 0], [0, 0]], [[0, 1], [0, 0]]], [[[0, 0], [0, 1]], [[0, 0], [1, 0]]]],
+    dtype=np.complex128,
+)
+SWAP_TENSOR = np.array(
+    [[[[1, 0], [0, 0]], [[0, 0], [1, 0]]], [[[0, 1], [0, 0]], [[0, 0], [0, 1]]]],
+    dtype=np.complex128,
+)
+
```
pranav97nair: Why not use `Ops.CZ`, `Ops.CNOT` and `Ops.SWAP`?

graphix/sim/density_matrix.py
```diff
@@ -186,33 +199,33 @@ def add_nodes(self, nqubit: int, data: Data) -> None:
...
             op : np.ndarray
                 2*2 matrix.
-            i : int
+            qubit : int
                 Index of qubit to apply operator.
         """
-        assert i >= 0
-        assert i < self.nqubit
+        assert qubit >= 0
+        assert qubit < self.nqubit
```
pranav97nair:
```suggestion
        assert 0 <= qubit < self.nqubit
```

graphix/sim/statevec.py
```diff
@@ -1,91 +1,147 @@
...
-    [[[[1, 0], [0, 0]], [[0, 1], [0, 0]]], [[[0, 0], [0, 1]], [[0, 0], [1, 0]]]],
-    dtype=np.complex128,
-)
-SWAP_TENSOR = np.array(
-    [[[[1, 0], [0, 0]], [[0, 0], [1, 0]]], [[[0, 1], [0, 0]], [[0, 0], [0, 1]]]],
-    dtype=np.complex128,
-)
+NUM_QUBIT_PARALLEL = 15
+"""This compilation constant determines the number of qubits above which matrix operations
+are multi-threaded. For lower counts, the overhead does not compensate parallelization.
```
pranav97nair: This is not very clear. Do you mean the advantage from parallelization is lost/negligible due to the overhead?

graphix/sim/statevec.py
```diff
@@ -100,19 +156,20 @@ def __init__(
         elif isinstance(data, Iterable):
             input_list = list(data)
         else:
-            raise TypeError(f"Incorrect type for data: {type(data)}")
+            raise TypeError(f"Incorrect type for data: {type(data)}.")
 
         if len(input_list) == 0:
```
pranav97nair:
```suggestion
        if (length := len(input_list)) == 0:
```
And then you can remove the assignments of `length` in the `elif` blocks.

tests/test_statevec.py
```diff
@@ -6,217 +6,420 @@
...
+            BasicStates.MINUS_I,
+        ],
+    )
+    def test_measurement_into_each_xyz_basis(self, state: PlanarState) -> None:
+        n = 3
+        k = 0
+        statevector = state.to_statevector_numpy()
+        m_op = np.outer(statevector, statevector.T.conjugate())
+        sv = Statevector(nqubit=n)
+        sv.evolve_single(m_op, k)
```
pranav97nair:
```suggestion
```

graphix/sim/statevec.py
pranav97nair: Remove the `align` wrapper here since Sphinx doesn't allow for nested Latex equation environments inside its `.. math` block, apparently.

pranav97nair (approved):
Thanks for this very nice addition, Mateo. I have just made a few very minor comments.

graphix/sim/density_matrix.py
matulni: Done in [d047155](https://github.com/TeamGraphix/graphix/pull/518/commits/d0471559eb397712bbc5db753fd3462b882991e0)

graphix/sim/statevec.py
matulni: Reworded in [d047155](https://github.com/TeamGraphix/graphix/pull/518/commits/d0471559eb397712bbc5db753fd3462b882991e0)

graphix/sim/statevec.py
matulni: Done in [d047155](https://github.com/TeamGraphix/graphix/pull/518/commits/d0471559eb397712bbc5db753fd3462b882991e0)

tests/test_statevec.py
matulni: Nice catch, thanks!
Removed in [d047155](https://github.com/TeamGraphix/graphix/pull/518/commits/d0471559eb397712bbc5db753fd3462b882991e0)

graphix/sim/statevec.py
matulni: Thanks, fixed in [d047155](https://github.com/TeamGraphix/graphix/pull/518/commits/d0471559eb397712bbc5db753fd3462b882991e0)

graphix/sim/density_matrix.py
matulni: Indeed, and with `Ops` we don't need to do a `reshape`. Fixed in [d047155](https://github.com/TeamGraphix/graphix/pull/518/commits/d0471559eb397712bbc5db753fd3462b882991e0)

I moved the `TENSOR` definitions to the test file.

PR #523, opened 2026-06-03
Add pretty printing for statevectors and density matrices

simsaidan:
First, implements function that pretty prints a complex number. Then, adds to statevecotr and density matrix. The density matrix is printed in matrix form.

Resolves #501 as part of UnitaryHack.

Before submitting, please check the following:

- Make sure you have tests for the new code and that test passes (run `nox`)
- If applicable, add a line to the [unreleased] part of CHANGELOG.md, following [keep-a-changelog](https://keepachangelog.com/en/1.0.0/).
- Format added code by `ruff`
  - See `CONTRIBUTING.md` for more details
- Make sure the checks (github actions) pass.

Then, please fill in below:

**Context (if applicable):**

**Description of the change:**

**Related issue:**

PR #503, opened 2026-05-12
Add pretty printers for Statevec and DensityMatrix

donglrd:
## Summary\n- add complex-number pretty formatting for fractions, square-root constants, and phase exponentials\n- add statevector ket expansion rendering via `Statevec.draw()`\n- add density-matrix outer-product rendering via `DensityMatrix.draw()`\n- cover the new formatting helpers and draw methods with tests\n\nCloses #501\n\n## Testing\n- `pytest tests/test_pretty_print.py -q`\n- `python3 -m ruff check graphix/pretty_print.py graphix/sim/statevec.py graphix/sim/density_matrix.py tests/test_pretty_print.py`

PR #524, opened 2026-06-03
Pretty-printing for Statevec and DensityMatrix (closes #501)

### LINKED ISSUE #501: Pretty-print for `Statevec` and `DensityMatrix`
_This issue is part of unitaryHACK 2026. You have to be [registered](https://unitaryhack.dev/register/) to participate. Please read carefully the [unitaryHack rules](https://unitaryhack.dev/hacker-guide/) before submitting a PR for this issue._

_A note about [AI Slop](https://en.wikipedia.org/wiki/AI_slop): while we are open to collaboration with LLMs for unitaryHACK, fully AI-generated PRs are not acceptable. It is at the maintainers' discretion whether or not LLM-generated PRs are the right fit for the issues, and those that appear fully AI-generated may be immediately rejected. Please read carefully the [AI policy](https://unitaryhack.dev/ai-guide/) if you are using LLMs for your work._

## Context

Simulating a quantum circuit or an MBQC pattern leads to an output state that is represented either as a statevector (`Statevec` class) or a density matrix (`DensityMatrix` class). These classes embed a numpy array that the users may want to print to check the computation. The `Statevec` class provides a method `to_dict()` to represent the statevector as a dictionary where values close to zero has been filtered.

```python
>>> c = Circuit(2)
>>> c.h(0)
>>> c.cz(0, 1)
>>> c.simulate_statevector().statevec.to_dict()
{'00': np.complex128(0.7071067811865474+0j), '01': np.complex128(0.7071067811865474+0j)}
```

We would like a method to pretty-print this statevector in a more user-friendly way: instead of the floating-point complex number `0.7071067811865474+0j`, we would prefer to pretty-print it as $\frac {\sqrt 2} 2$.
 
## Tasks

- Define a function in `pretty_print.py` to pretty-print a complex number. This function should take an `output` parameter of type `OutputFormat`. We would like common values to be rendered nicely: that includes fractions (`0.25` should be printed as $\frac 1 4$), square roots (`0.70710678` should be printed as $\frac {\sqrt 2} 2$) and exponentials (`0.5 + 0.8660254j` should be printed as $\mathrm e^{\mathrm i \frac \pi 3}$.
- Define a function in `pretty_print.py` to pretty-print a statevector using the previously defined function and `Statevec.to_dict`. Define a method `draw` in `Statevec` that calls this function. Use the ket-notation with an `ENCODING` parameter as for `Statevec.to_dict`.
- Define a function in `pretty_print.py` to pretty-print a density matrix. Define a method `draw` in `DensityMatrix` that calls this function.

## Acceptance criteria

For the PR to be considered for review:

- It should pass the CI, including linting and type-checking. 

- New functions should be tested and documented appropriately following [numpy style](https://numpydoc.readthedocs.io/en/latest/format.html).

- Unitary Hack [AI use guidelines](https://unitaryhack.dev/ai-guide/) should be respected.

- Author should be ready to engage in a feedback loop.

For the PR to be accepted:

- All provided feedback should have been implemented.

Vinny010:
Closes #501.

Adds human-friendly rendering of amplitudes and states:

- `complex_to_str` — renders common values exactly (`1/4`, `√2/2`, `e^(iπ/3)`) via a square-then-rationalize heuristic, reusing the existing `angle_to_str` for the exponential phase. Supports ASCII / Unicode / LaTeX.
- `statevec_to_str` + `Statevec.draw` — ket notation, honouring the `to_dict` `encoding` parameter (e.g. `√2/2|00⟩ + √2/2|01⟩`).
- `density_matrix_to_str` + `DensityMatrix.draw` — column-aligned grid (ASCII/Unicode) or LaTeX `pmatrix`.
- Tests covering the issue examples plus edge cases (zero, negative, pure imaginary, LaTeX, decimal fallback, symbolic), with numpy-style docstrings.

Passes `ruff check`, `ruff format`, `mypy --strict`, and `pytest` locally (167 tests in the affected suites).

**Design points I'd value feedback on:**
- `draw()` returns a `str`, consistent with the existing `to_unicode` / `to_latex` style.
- The complex formatter prefers exponential form over cartesian when both magnitude and phase are recognized.

Happy to adjust either.

_Developed with LLM assistance, reviewed and tested by me._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Added new attribute `OpenGraph.output_cliffords`
   - Added `clifford` abstract method to `AbstractMeasurement`. Implemented it for `Plane` and `Axis`.
 
-- #524: Added pretty-printing for `Statevec` and `DensityMatrix` via new functions `statevec_to_str` and `density_matrix_to_str` (and the corresponding `__str__` methods), with ASCII, Unicode, and LaTeX output and recognition of common amplitudes (rational multiples, square roots, and factored uniform magnitudes).
+- #524: Added `Statevec.draw` and `DensityMatrix.draw` methods (and the underlying `statevec_to_str` and `density_matrix_to_str` functions) for pretty-printing states and density matrices in ASCII, Unicode, and LaTeX.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Added new attribute `OpenGraph.output_cliffords`
   - Added `clifford` abstract method to `AbstractMeasurement`. Implemented it for `Plane` and `Axis`.
 
+- #524: Added pretty-printing for `Statevec` and `DensityMatrix` via new functions `statevec_to_str` and `density_matrix_to_str` (and the corresponding `__str__` methods), with ASCII, Unicode, and LaTeX output and recognition of common amplitudes (rational multiples, square roots, and factored uniform magnitudes).
+
 ### Fixed
 
 - #454, #481: Ensure `Pattern.minimize_space` only reduces max-space and does not increase it.
--- a/graphix/pretty_print.py
+++ b/graphix/pretty_print.py
@@ -570,17 +570,29 @@ def _real_to_str(x: float, output: OutputFormat, max_denominator: int, atol: flo
     return _render_real(*rec, output)
 
 
+def _render_imaginary(signed_num: int, inner: int, den: int, output: OutputFormat) -> str:
+    """Render ``signed_num * sqrt(inner) / den * i`` with the unit leading the numerator.
+
+    The imaginary unit is placed at the front of the numerator (e.g. ``i/2``,
+    ``i√2/2``, ``3i/4``) so that it reads as ``i`` times a real magnitude, with a
+    unit coefficient collapsing to a bare ``±i``.
+    """
+    sign = "-" if signed_num < 0 else ""
+    magnitude = abs(signed_num)
+    unit = _imaginary_unit(output)
+    coefficient = "" if magnitude == 1 else f"{magnitude}"
+    numerator = f"{coefficient}{unit}{_sqrt_str(inner, output)}"
+    if den == 1:
+        return f"{sign}{numerator}"
+    return f"{sign}{_fraction_str(numerator, str(den), output)}"
+
+
 def _imaginary_to_str(x: float, output: OutputFormat, max_denominator: int, atol: float) -> str | None:
     """Render a purely imaginary value ``x * i``."""
     rec = _recognize_sqrt(x, max_denominator, atol)
     if rec is None:
         return None
-    signed_num, inner, den = rec
-    unit = _imaginary_unit(output)
-    # A unit coefficient collapses to just ``±i``.
-    if abs(signed_num) == 1 and inner == 1 and den == 1:
-        return f"{'-' if signed_num < 0 else ''}{unit}"
-    return f"{_render_real(signed_num, inner, den, output)}{unit}"
+    return _render_imaginary(*rec, output)
 
 
 def _recognize_angle_over_pi(theta: float, max_denominator: int, atol: float) -> Fraction | None:
@@ -700,17 +712,23 @@ def complex_to_str(
     if abs(z.real) <= atol:
         return _imaginary_to_str(z.imag, output, max_denominator, atol) or _decimal_to_str(z, output, precision)
     exponential = _exponential_to_str(z, output, max_denominator, atol)
-    if exponential is not None:
+    # The exponential form is the clearest representation on the unit circle, but for
+    # other moduli a Cartesian form (e.g. ``1 + i`` rather than ``√2·e^(iπ/4)``) reads
+    # better, so it is preferred when both parts are recognized.
+    if exponential is not None and math.isclose(math.hypot(z.real, z.imag), 1.0, abs_tol=atol):
         return exponential
     cartesian = _cartesian_to_str(z.real, z.imag, output, max_denominator, atol)
     if cartesian is not None:
         return cartesian
+    # Fall back to a radius-prefixed exponential when the Cartesian parts are not nice.
+    if exponential is not None:
+        return exponential
     return _decimal_to_str(z, output, precision)
 
 
 def _ket_str(ket: str, output: OutputFormat) -> str:
     if output == OutputFormat.LaTeX:
-        return rf"\lvert {ket}\rangle"
+        return rf"\ket{{{ket}}}"
     if output == OutputFormat.Unicode:
         return f"|{ket}⟩"
     return f"|{ket}>"
--- a/graphix/pretty_print.py
+++ b/graphix/pretty_print.py
@@ -509,8 +509,6 @@ def _recognize_sqrt(x: float, max_denominator: int, atol: float, rtol: float) ->
         squarefree integer, encoding ``x = signed_num * sqrt(inner) / den``.
         Returns ``None`` when ``x`` is not recognized as such a value.
     """
-    if x == 0:
-        return 0, 1, 1
     square = Fraction(x * x).limit_denominator(max_denominator)
     if not math.isclose(x * x, float(square), rel_tol=rtol, abs_tol=atol):
         return None
@@ -749,6 +747,37 @@ def _needs_parentheses(coefficient: str) -> bool:
     return " + " in coefficient or " - " in coefficient
 
 
+def _factor_uniform_magnitude(coefficients: Sequence[str], kets: Sequence[str]) -> str | None:
+    """Factor a magnitude shared by every amplitude, e.g. ``√2/2(|00⟩ + |11⟩)``.
+
+    Returns the factored string when all amplitudes have the same non-unit magnitude
+    (up to sign), or ``None`` when no such common factor exists (in which case the
+    caller falls back to the term-by-term rendering).
+    """
+    if len(coefficients) < 2:
+        return None
+    magnitudes: list[str] = []
+    signs: list[str] = []
+    for coefficient in coefficients:
+        # A compound (Cartesian) coefficient has no single magnitude to factor out.
+        if _needs_parentheses(coefficient):
+            return None
+        if coefficient.startswith("-"):
+            signs.append("-")
+            magnitudes.append(coefficient[1:])
+        else:
+            signs.append("+")
+            magnitudes.append(coefficient)
+    common = magnitudes[0]
+    # Nothing to factor when the magnitude is the unit coefficient, or it is not shared.
+    if common == "1" or any(magnitude != common for magnitude in magnitudes):
+        return None
+    body = kets[0] if signs[0] == "+" else f"-{kets[0]}"
+    for sign, ket in zip(signs[1:], kets[1:], strict=True):
+        body += f" {sign} {ket}"
+    return f"{common}({body})"
+
+
 def statevec_to_str(
     statevec: Statevec,
     output: OutputFormat,
@@ -788,17 +817,22 @@ def statevec_to_str(
     Returns
     -------
     str
-        The formatted statevector, e.g. ``√2/2|00⟩ + √2/2|01⟩``.
+        The formatted statevector, e.g. ``√2/2(|00⟩ + |01⟩)``.
     """
     amplitudes = statevec.to_dict(encoding, rtol=rtol, atol=atol)
     if not amplitudes:
         return "0"
+    coefficients = [
+        complex_to_str(amplitude, output, max_denominator=max_denominator, atol=atol, rtol=rtol, precision=precision)
+        for amplitude in amplitudes.values()
+    ]
+    kets = [_ket_str(ket, output) for ket in amplitudes]
+    # When every amplitude shares the same magnitude, factor it out, e.g. ``√2/2(|00⟩ + |11⟩)``.
+    factored = _factor_uniform_magnitude(coefficients, kets)
+    if factored is not None:
+        return factored
     result = ""
-    for index, (ket, amplitude) in enumerate(amplitudes.items()):
-        coefficient = complex_to_str(
-            amplitude, output, max_denominator=max_denominator, atol=atol, rtol=rtol, precision=precision
-        )
-        ket_str = _ket_str(ket, output)
+    for index, (coefficient, ket_str) in enumerate(zip(coefficients, kets, strict=True)):
         if coefficient == "1":
             term = ket_str
         elif coefficient == "-1":
--- a/graphix/pretty_print.py
+++ b/graphix/pretty_print.py
@@ -8,7 +8,7 @@ import string
 from enum import Enum
 from fractions import Fraction
 from math import pi
-from typing import TYPE_CHECKING, SupportsFloat
+from typing import TYPE_CHECKING, SupportsComplex, SupportsFloat
 
 # `assert_never` introduced in Python 3.11
 from typing_extensions import assert_never
@@ -26,6 +26,8 @@ if TYPE_CHECKING:
     from graphix.flow.core import PauliFlow, XZCorrections
     from graphix.fundamentals import Angle
     from graphix.pattern import Pattern
+    from graphix.sim.density_matrix import DensityMatrix
+    from graphix.sim.statevec import _ENCODING, Statevec
 
 
 class OutputFormat(Enum):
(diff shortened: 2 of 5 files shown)


graphix/pretty_print.py
```diff
@@ -439,3 +441,375 @@ def xzcorr_to_str(xzcorr: XZCorrections[AbstractMeasurement], output: OutputForm
...
+    else:
+        imag = f"{_render_real(abs(signed_num), inner, den, output)}{unit}"
+    return f"{re_str}{connector}{imag}"
+
+
+def _decimal_to_str(z: complex, output: OutputFormat) -> str:
+    """Fallback formatting using rounded decimals."""
+    unit = _imaginary_unit(output)
+    if abs(z.imag) <= _DEFAULT_ATOL:
+        return f"{z.real:.4g}"
```
thierry-martinez: We should be able to configure the precision of the printed value.

graphix/pretty_print.py
```diff
@@ -439,3 +441,375 @@ def xzcorr_to_str(xzcorr: XZCorrections[AbstractMeasurement], output: OutputForm
...
+    d = 2
+    while d * d <= inner:
+        while inner % (d * d) == 0:
+            inner //= d * d
+            outer *= d
+        d += 1
+    return outer, inner
+
+
+def _recognize_real(x: float, max_denominator: int, atol: float) -> tuple[int, int, int] | None:
```
thierry-martinez: The name `_recognize_real` feels overly generic; would `_recognize_sqrt` be a better fit?

thierry-martinez (commented):
Thank you very much for your contribution!

Codecov has detected that some parts of your code are not tested.

As you may have noticed, PR #503 and PR #523 address the same issue (#501). We invite you to cross‑review each other's work: PR reviews are an inherent part of software development, just as important as writing code. Moreover, this review process will help you position your PR relative to the others, allowing us to determine which contribution best addresses the issue.

Vinny010:
Thanks for the review, @thierry-martinez! Pushed [eebb104](https://github.com/TeamGraphix/graphix/pull/524/commits/eebb1048b5d103f5c77dba4ec391cbd69d45b654) addressing both inline points:

- **Configurable precision** — added a `precision` keyword argument to `complex_to_str`, `statevec_to_str`, `density_matrix_to_str`, and the matching `Statevec.draw` / `DensityMatrix.draw` methods. It controls the number of significant digits used by the decimal fallback. The default (`4`) preserves the previous behaviour. New regression test covering several precisions.
- **Rename** — renamed `_recognize_real` to `_recognize_sqrt`. The helper recognizes a real number as `signed_num * sqrt(inner) / den`, and the rationals are covered as the special case `inner == 1` — added a sentence to the docstring to make that explicit.

I'll have a look at #503 and #523 and follow up with comments there.

Vinny010:
@thierry-martinez — small process note: looks like the workflow run on `eebb104` is still awaiting approval (no check runs have fired since the response push). Whenever you have a moment to re-trigger it, happy to address anything else that comes up.

graphix/pretty_print.py
```diff
@@ -439,3 +441,393 @@ def xzcorr_to_str(xzcorr: XZCorrections[AbstractMeasurement], output: OutputForm
...
+        return exponential
+    cartesian = _cartesian_to_str(z.real, z.imag, output, max_denominator, atol)
+    if cartesian is not None:
+        return cartesian
+    return _decimal_to_str(z, output, precision)
+
+
+def _ket_str(ket: str, output: OutputFormat) -> str:
+    if output == OutputFormat.LaTeX:
+        return rf"\lvert {ket}\rangle"
```
thierry-martinez: I suggest you to use `\ket{{ket}}` instead.

tests/test_pretty_print.py
```diff
@@ -202,3 +207,121 @@ def test_xzcorr_str() -> None:
...
+        (1, "1"),
+        (-1, "-1"),
+        (2, "2"),
+        (0.5, "1/2"),
+        (-0.25, "-1/4"),
+        (2**-0.5, "√2/2"),
+        (math.sqrt(3) / 2, "√3/2"),
+        (1j, "i"),
+        (-1j, "-i"),
+        (0.5j, "1/2i"),
```
thierry-martinez: I would prefer `i/2`.

tests/test_pretty_print.py
```diff
@@ -202,3 +207,121 @@ def test_xzcorr_str() -> None:
...
+        (-(2**-0.5) * 1j, "-√2/2i"),
```
thierry-martinez: I would prefer `-i√2/2`.

tests/test_pretty_print.py
```diff
@@ -202,3 +207,121 @@ def test_xzcorr_str() -> None:
...
+    assert complex_to_str(0.25, OutputFormat.LaTeX) == r"\frac{1}{4}"
+    assert complex_to_str(0.5 + math.sqrt(3) / 2 * 1j, OutputFormat.LaTeX) == r"\mathrm{e}^{\mathrm{i} \frac{\pi}{3}}"
+
+
+def test_complex_to_str_fallback_and_symbolic() -> None:
+    # An unrecognized value falls back to a rounded decimal.
+    assert complex_to_str(0.123456, OutputFormat.ASCII) == "0.1235"
+    # A non-numeric object is stringified rather than raising.
+    assert complex_to_str("alpha", OutputFormat.ASCII) == "alpha"
+
```
thierry-martinez: I suggest merging these tests with `test_complex_to_str_unicode_values` by replacing expected with a `Mapping[OutputFormat, str]`. This would allow us to check multiple output formats within the same parametrized test, making it clearer to compare the output formats and their test coverage.

tests/test_pretty_print.py
```diff
@@ -202,3 +207,121 @@ def test_xzcorr_str() -> None:
...
+    dm = DensityMatrix(data=[BasicStates.ZERO])
+    assert dm.draw(OutputFormat.ASCII) == "[ 1  0 ]\n[ 0  0 ]"
+    assert dm.draw(OutputFormat.LaTeX) == r"\begin{pmatrix}1 & 0 \\ 0 & 0\end{pmatrix}"
+
+
+def test_complex_to_str_exponential_with_radius() -> None:
+    # |z| != 1: the radius prefixes the exponential form (1 + i = √2 e^{iπ/4}).
+    assert complex_to_str(1 + 1j, OutputFormat.Unicode) == "√2·e^(iπ/4)"
+    assert complex_to_str(1 + 1j, OutputFormat.ASCII) == "sqrt(2)*e^(i*pi/4)"
+    assert complex_to_str(1 + 1j, OutputFormat.LaTeX) == r"\sqrt{2} \mathrm{e}^{\mathrm{i} \frac{\pi}{4}}"
```
thierry-martinez: I am not sure `√2·e^(iπ/4)` is any simpler to read than `1 + i`. More generally, I tend to think that rational multiples of `√2·e^(±iπ/4)` are more readable when written as Cartesian form.

tests/test_pretty_print.py
```diff
@@ -202,3 +207,121 @@ def test_xzcorr_str() -> None:
...
+
+
+def test_complex_to_str_complex_decimal_fallback() -> None:
+    # Neither part is a recognized value -> rounded decimal real and imaginary parts.
+    assert complex_to_str(0.123456 + 0.234567j, OutputFormat.Unicode) == "0.1235+0.2346i"
+
+
+def test_complex_to_str_imaginary_formats() -> None:
+    assert complex_to_str(0.5j, OutputFormat.LaTeX) == r"\frac{1}{2}\mathrm{i}"
+    assert complex_to_str(0.5j, OutputFormat.ASCII) == "1/2i"
```
thierry-martinez: I suggest merging these tests with `test_complex_to_str_unicode_values` as well.

thierry-martinez (commented):
Thank you for your update, @Vinny010. I think we are on a good track to merge this PR. In this respect, please add a comment in the discussion of #501 so that I can assign the issue to you.

I have also made a few remarks (as a side note, I did not use the "Request changes" flag since we've chosen not to use it in the Graphix project, to avoid blocking the PR from being merged if other maintainers approve it). Note that we have a two‑approval rule for non‑trivial PRs: therefore, once I or another maintainer approves this PR, we'll need a second approval before merging.

graphix/pretty_print.py
thierry-martinez: The notation has been changed in d3f8a56, but the LaTeX notation for kets is not covered by any test.

tests/test_pretty_print.py
```diff
@@ -279,10 +277,17 @@ def test_density_matrix_draw() -> None:
...
+    # radius-prefixed exponential (1 + i rather than √2·e^(iπ/4)).
+    assert complex_to_str(1 + 1j, OutputFormat.Unicode) == "1 + i"
+    assert complex_to_str(1 + 1j, OutputFormat.ASCII) == "1 + i"
+    assert complex_to_str(1 + 1j, OutputFormat.LaTeX) == r"1 + \mathrm{i}"
+    # When the Cartesian parts are not recognized, the radius-prefixed exponential is
+    # used as a last resort before the decimal fallback.
+    z = 2 * (math.cos(math.pi / 5) + math.sin(math.pi / 5) * 1j)
+    assert complex_to_str(z, OutputFormat.Unicode) == "2·e^(iπ/5)"
+    assert complex_to_str(z, OutputFormat.ASCII) == "2*e^(i*pi/5)"
+    assert complex_to_str(z, OutputFormat.LaTeX) == r"2 \mathrm{e}^{\mathrm{i} \frac{\pi}{5}}"
```
thierry-martinez: This test and all the following tests up to the end of the file can be added to the cases tested by `test_complex_to_str_values`.

graphix/sim/density_matrix.py
```diff
@@ -117,6 +118,39 @@ def __str__(self) -> str:
         """Return a string description."""
         return f"DensityMatrix object, with density matrix {self.rho} and shape {self.dims()}."
 
+    def draw(
+        self,
+        output: OutputFormat = OutputFormat.Unicode,
+        *,
+        max_denominator: int = 1000,
+        atol: float = 1e-9,
```
thierry-martinez, line 126: Everywhere an `atol` parameter is present, we should be able to pass an `rtol` parameter as well.

graphix/pretty_print.py
```diff
@@ -439,3 +441,411 @@ def xzcorr_to_str(xzcorr: XZCorrections[AbstractMeasurement], output: OutputForm
...
+    connector = " - " if signed_num < 0 else " + "
+    unit = _imaginary_unit(output)
+    if abs(signed_num) == 1 and inner == 1 and den == 1:
+        imag = unit
+    else:
+        imag = f"{_render_real(abs(signed_num), inner, den, output)}{unit}"
+    return f"{re_str}{connector}{imag}"
+
+
+def _decimal_to_str(z: complex, output: OutputFormat, precision: int) -> str:
```
thierry-martinez: `atol` should be passed as argument instead of relying on a default value.

thierry-martinez (commented):
Thank you for your changes. Just minor comments left. Please have a look at the test coverage: there are still some gaps.

Vinny010:
Thanks @thierry-martinez — all four addressed in 58eb451.

- **`rtol` parameter.** Added wherever `atol` is exposed: `complex_to_str`, `density_matrix_to_str` and `DensityMatrix.draw` (it was indeed missing there), threaded down into the `math.isclose` recognition calls as `rel_tol`. `Statevec.draw`/`statevec_to_str` already had it; it now also reaches the per-amplitude recognition, not just the amplitude-dropping. Default is `0.0`, so existing behaviour is unchanged.
- **`_decimal_to_str`.** Now takes `atol` as an argument instead of reading the module-level `_DEFAULT_ATOL`.
- **LaTeX ket coverage.** The `\ket{...}` notation is now exercised in the statevec draw tests (both the bare basis state and the coefficient-prefixed Bell case).
- **Test consolidation.** Merged the remaining `complex_to_str` value tests (radius/Cartesian/decimal-fallback/integer-times-surd) into the parametrized `test_complex_to_str_values`. I kept two things out of it deliberately: `test_complex_to_str_precision_is_configurable` (it varies the `precision` argument rather than testing a default-args value→string mapping) and the `Statevec`/`DensityMatrix` `draw` tests (different entry points). Happy to fold those in too if you'd rather have a single table.
- **Coverage for `rtol`.** Added `test_complex_to_str_rtol_controls_recognition` (a value just off 1/2 is a decimal at the default tolerance, but recognized as `1/2` with a looser `rtol`) and a `DensityMatrix.draw(rtol=...)` test.

graphix/pretty_print.py
```diff
@@ -439,3 +441,427 @@ def xzcorr_to_str(xzcorr: XZCorrections[AbstractMeasurement], output: OutputForm
...
+
+
+def statevec_to_str(
+    statevec: Statevec,
+    output: OutputFormat,
+    *,
+    encoding: _ENCODING = "MSB",
+    max_denominator: int = _DEFAULT_MAX_DENOMINATOR,
+    atol: float = _DEFAULT_ATOL,
+    rtol: float = 0.0,
```
thierry-martinez, line 826:
```suggestion
    rtol: float = _DEFAULT_RTOL,
```

graphix/pretty_print.py
```diff
@@ -439,3 +441,427 @@ def xzcorr_to_str(xzcorr: XZCorrections[AbstractMeasurement], output: OutputForm
...
+        decimal representation (default: ``4``).
+
+    Returns
+    -------
+    str
+        The formatted statevector, e.g. ``√2/2|00⟩ + √2/2|01⟩``.
+    """
+    amplitudes = statevec.to_dict(encoding, rtol=rtol, atol=atol)
+    if not amplitudes:
+        return "0"
```
thierry-martinez, line 862: This line is not covered by any test.

graphix/pretty_print.py
```diff
@@ -439,3 +441,427 @@ def xzcorr_to_str(xzcorr: XZCorrections[AbstractMeasurement], output: OutputForm
...
+    return f"{re_str}{connector}{imag}"
+
+
+def _decimal_to_str(z: complex, output: OutputFormat, precision: int, atol: float) -> str:
+    """Fallback formatting using rounded decimals with ``precision`` significant digits."""
+    unit = _imaginary_unit(output)
+    if abs(z.imag) <= atol:
+        return f"{z.real:.{precision}g}"
+    if abs(z.real) <= atol:
+        return f"{z.imag:.{precision}g}{unit}"
```
thierry-martinez, line 666: This line is not covered by any tests.

graphix/pretty_print.py
```diff
@@ -439,3 +441,427 @@ def xzcorr_to_str(xzcorr: XZCorrections[AbstractMeasurement], output: OutputForm
...
+
+def _exponential_to_str(z: complex, output: OutputFormat, max_denominator: int, atol: float, rtol: float) -> str | None:
+    """Render ``z`` as ``r e^{iθ}`` when both ``r`` and ``θ / π`` are recognized."""
+    theta = math.atan2(z.imag, z.real)
+    angle_frac = _recognize_angle_over_pi(theta, max_denominator, atol, rtol)
+    if angle_frac is None or angle_frac == 0:
+        return None
+    radius = _real_to_str(math.hypot(z.real, z.imag), output, max_denominator, atol, rtol)
+    if radius is None:
+        return None
```
thierry-martinez, line 628: This line is not covered by any tests.

graphix/pretty_print.py
```diff
@@ -439,3 +441,427 @@ def xzcorr_to_str(xzcorr: XZCorrections[AbstractMeasurement], output: OutputForm
...
+    if den == 1:
+        return f"{sign}{numerator}"
+    return f"{sign}{_fraction_str(numerator, str(den), output)}"
+
+
+def _imaginary_to_str(x: float, output: OutputFormat, max_denominator: int, atol: float, rtol: float) -> str | None:
+    """Render a purely imaginary value ``x * i``."""
+    rec = _recognize_sqrt(x, max_denominator, atol, rtol)
+    if rec is None:
+        return None
```
thierry-martinez, line 607: This line is not covered by any tests.

graphix/pretty_print.py
```diff
@@ -439,3 +441,427 @@ def xzcorr_to_str(xzcorr: XZCorrections[AbstractMeasurement], output: OutputForm
...
+def _fraction_str(num: str, den: str, output: OutputFormat) -> str:
+    if output == OutputFormat.LaTeX:
+        return rf"\frac{{{num}}}{{{den}}}"
+    return f"{num}/{den}"
+
+
+def _render_real(signed_num: int, inner: int, den: int, output: OutputFormat) -> str:
+    """Render ``signed_num * sqrt(inner) / den`` produced by :func:`_recognize_sqrt`."""
+    if signed_num == 0:
+        return "0"
```
thierry-martinez, line 564: This line is not covered by any tests.

graphix/pretty_print.py
```diff
@@ -439,3 +441,427 @@ def xzcorr_to_str(xzcorr: XZCorrections[AbstractMeasurement], output: OutputForm
...
+
+    Returns
+    -------
+    tuple[int, int, int] or None
+        ``(signed_num, inner, den)`` with ``den > 0`` and ``inner`` a positive
+        squarefree integer, encoding ``x = signed_num * sqrt(inner) / den``.
+        Returns ``None`` when ``x`` is not recognized as such a value.
+    """
+    if x == 0:
+        return 0, 1, 1
```
thierry-martinez: This line is not covered by any tests.

graphix/pretty_print.py
```diff
@@ -439,3 +441,427 @@ def xzcorr_to_str(xzcorr: XZCorrections[AbstractMeasurement], output: OutputForm
...
+        Non-negative integer to decompose.
+
+    Returns
+    -------
+    tuple[int, int]
+        ``(outer, inner)`` such that ``outer ** 2 * inner == n`` and ``inner`` is
+        squarefree. ``n == 0`` returns ``(0, 1)``.
+    """
+    if n == 0:
+        return 0, 1
```
thierry-martinez, line 475: This line is not covered by any tests.

graphix/sim/statevec.py
```diff
@@ -520,6 +521,64 @@ def to_prob_dict(
...
+            The formatted statevector.
+
+        Examples
+        --------
+        >>> from graphix.transpiler import Circuit
+        >>> circuit = Circuit(2)
+        >>> circuit.h(0)
+        >>> circuit.cz(0, 1)
+        >>> print(circuit.simulate_statevector().statevec.draw())
+        √2/2|00⟩ + √2/2|01⟩
```
thierry-martinez: Uniform magnitude should be factored out: 
```suggestion
        √2/2(|00⟩ + |01⟩)
```

thierry-martinez (commented):
I extracted the coverage results from Codecov. You can view them here: https://app.codecov.io/gh/TeamGraphix/graphix/pull/524?src=pr&el=tree

Vinny010:
Thanks @thierry-martinez — follow-up addressed in 5b7c389.

**Uniform magnitude.** Statevec output now factors a magnitude shared by every amplitude: `√2/2(|00⟩ + |11⟩)` instead of `√2/2|00⟩ + √2/2|11⟩`, with the signs kept inside the parentheses (`1/2(|00⟩ - |01⟩ + |10⟩ + |11⟩)`). Compound (Cartesian) amplitudes and non-uniform magnitudes fall back to the term-by-term form. Updated the affected tests and the docstring / doctest examples.

**Uncovered lines.** Added tests reaching each flagged branch:
- tiny non-zero real → `"0"` (the zero branches of `_squarefree_decomposition` and `_render_real`),
- a purely imaginary decimal (`0.234567j` → `0.2346i`, the `_imaginary_to_str` → None and `_decimal_to_str` imaginary branches),
- a recognized phase with an unrecognized radius (`π·e^(iπ/4)`, the `_exponential_to_str` radius-None branch),
- the empty-statevector `"0"` (via a tolerance large enough to drop every amplitude).

For the `_recognize_sqrt` `x == 0` guard I removed it rather than test it: the general path already returns `(0, 1, 1)` for `x == 0`, so it was dead code.

(I checked the new branches execute by tracing the outputs; `pytest-cov` isn't in my local env, so if your coverage run still flags anything I missed, point me at the line and I'll cover it.)

Vinny010:
Note on the red CI here: the 7 failures are all in `tests/test_visualization.py` — the `pytest-mpl` image-comparison tests (`test_og_draw`, the `test_*_flow_draw` set, `test_xzcorr_draw`, `test_draw_graph_reference`), each failing with "Image files did not match". The other 3342 tests pass.

Those render graph/flow figures and don't touch anything in this diff, which only changes amplitude/statevector text formatting — nothing in `visualization.py` calls `complex_to_str`/`statevec_to_str`/`density_matrix_to_str`. The same `--mpl` suite passed on #526's CI run a day earlier (11 Jun), so this looks like matplotlib-version pixel drift in the reference images rather than a regression from this branch.

Happy to help if a baseline refresh is wanted — if you let me know the matplotlib version the CI pins to, I can regenerate the references against it; or if you'd rather keep baseline changes out of this PR, that's fine too and I'll leave it to you.

thierry-martinez:
Yes, indeed, the CI broke because of a new Matplotlib release. It was fixed in #541, which has now been merged. Could you merge the current `master` into your branch?

thierry-martinez:
> I checked the new branches execute by tracing the outputs; pytest-cov isn't in my local env, so if your coverage run still flags anything I missed, point me at the line and I'll cover it.

As I mentioned in my previous comment, you can view the code‑coverage results here: https://app.codecov.io/gh/TeamGraphix/graphix/pull/524

Vinny010:
Thanks for the codecov link — that pinned it to a single line. The patch coverage was 99.5%, with one miss: `pretty_print.py:847`, the `result += f" - {term[1:]}"` separator in the term-by-term statevec fallback.

It was a side effect of the new uniform-magnitude factoring: the negative-amplitude case I had a test for (`[0.5, -0.5, 0.5, 0.5]`) now goes through the factored path (`1/2(|00⟩ - |01⟩ + ...)`), so it no longer reaches the fallback loop, and the only remaining fallback test (the binomial) has no negative term. Fixed in 78a48e7 by giving `test_statevec_draw_non_uniform_not_factored` a negative second amplitude (`√3/2|0⟩ - 1/2|1⟩`), which flows through the fallback and exercises that branch. Patch coverage should be 100% now.

pranav97nair (commented):
Hi @Vinny010 , thanks for your work on this PR! This will be a nice addition to Graphix. I have a couple points to add.

Currently, the statevector draw function factors out the magnitude when it is shared by all component amplitudes. I would however like to see this factoring even in the case when there is a relative phase, as this does not change the magnitude of the affected component. A basic example:
```python
plus = Statevec(data=BasicStates.PLUS)
assert plus.draw() == "√2/2(|0⟩ + |1⟩)" # Passes

plus_i = Statevec(data=BasicStates.PLUS_I)
assert plus_i.draw() == "√2/2(|0⟩ + i|1⟩)" # Fails
```
Could you update your factoring function to account for this case and enforce factoring when all magnitudes are equal up to a relative phase $e^{i\varphi}$ ?

Also, could you include a test checking the impact of the `max_denominator` parameter of the new `draw()` functions? As it stands, I believe none of the tests use anything other than the default value.

Vinny010:
Thanks @pranav97nair — both addressed in 0faf587.

**Relative-phase factoring.** `_factor_uniform_magnitude` now works on the amplitudes rather than the coefficient strings: it factors a shared modulus `r` and keeps each per-component phase `amplitude / r` inside the parentheses. So your example now holds:

```python
Statevec(data=BasicStates.PLUS).draw()    # √2/2(|0⟩ + |1⟩)
Statevec(data=BasicStates.PLUS_I).draw()  # √2/2(|0⟩ + i|1⟩)
Statevec(data=BasicStates.MINUS_I).draw() # √2/2(|0⟩ - i|1⟩)
```

It still falls back to term-by-term when the moduli genuinely differ, when the shared modulus is trivial (`1`), or when the amplitudes are non-numeric (symbolic), so nothing regresses there. Added tests for the phase cases (Unicode + LaTeX) and direct edge-case coverage of the helper.

**`max_denominator` test.** Added `test_draw_max_denominator`, which checks both draw functions: with `max_denominator=1`, `DensityMatrix.draw` renders `0.5` instead of `1/2` and `Statevec.draw` renders `0.7071` instead of `√2/2` (the denominators can no longer be recognized).

Local run is green (ruff / mypy / 185 tests + doctests). Let me know if you'd like the relative phase rendered differently (e.g. an explicit `e^{iφ}` form for non-Pauli phases).

Vinny010:
Hi @thierry-martinez and @pranav97nair — a quick "ready for re-review" ping with the deadline tomorrow in mind.

Since your last feedback I've:
- Addressed the `pretty_print.py:847` coverage line (the fallback path is no longer reachable for the negative-amplitude case, explained in my Jun 13 reply)
- Implemented relative-phase factoring per @pranav97nair's example (commit `0faf587`) — `PLUS`, `PLUS_I`, `MINUS_I` now render with the shared modulus factored out

CI is fully green (24/24) and mergeable state is clean. Whenever you have a moment, would appreciate a final look — happy to make any last tweaks.

Thanks for all the review work on this one!

thierry-martinez (approved):
You have my approval! Please wait for a second approval (probably from @pranav97nair) and make sure to update from the branch (letting the CI run) before merging. The issue #501 is already assigned to you, so you'll receive the bounty. Thank you very much for your contribution!

Vinny010:
Thank you so much for the approval and the careful review throughout, @thierry-martinez — really appreciate it.

Going to update the branch from master now so CI can re-run, then ping @pranav97nair to take a final look. Excited to get this one across the line.

tests/test_pretty_print.py
```diff
@@ -202,3 +208,218 @@ def test_xzcorr_str() -> None:
...
+def test_draw_max_denominator() -> None:
+    # `max_denominator` caps the denominators the recognition will accept.
+    dm = DensityMatrix(data=[BasicStates.PLUS])
+    assert dm.draw(OutputFormat.Unicode) == "[ 1/2  1/2 ]\n[ 1/2  1/2 ]"
+    # With max_denominator=1, 1/2 can no longer be recognized and falls back to a decimal.
+    assert dm.draw(OutputFormat.Unicode, max_denominator=1) == "[ 0.5  0.5 ]\n[ 0.5  0.5 ]"
+    # Same effect on the statevector draw (√2/2 -> 0.7071).
+    sv = Statevec([2**-0.5, 2**-0.5])
+    assert sv.draw(OutputFormat.Unicode) == "√2/2(|0⟩ + |1⟩)"
+    assert sv.draw(OutputFormat.Unicode, max_denominator=1) == "0.7071(|0⟩ + |1⟩)"
```
pranav97nair, line 405: The denominator of the desired fraction is 2 for both the density matrix and the state vector. Changing the `max_denominator` argument to 2 for the state vector case (line 400) makes the test fail as expected as it correctly converts the amplitude to `√2/2`. However, the density matrix case (line 396) does not fail until you pass `max_denominator=4`. 

Can you address this discrepancy? Specifically, 
```python
dm.draw(OutputFormat.Unicode, max_denominator=2)
```
should output `"[ 1/2  1/2 ]\n[ 1/2  1/2 ]"`.

pranav97nair (commented):
Hi @Vinny010 , thanks for the updates and your prompt responses to our feedback throughout! I have one last thing to point out, please take a look.

Vinny010:
Thanks for the catch, @pranav97nair — that discrepancy was hiding a real semantic gap in how `max_denominator` propagated through `_recognize_sqrt`.

Pushed `f0ea250` to address it. The root cause was that the recognition only went through the squared form (`Fraction(x*x).limit_denominator(...)`), which effectively required `q ** 2 <= max_denominator` to recognise simple rationals — so `DensityMatrix.draw` needed `max_denominator=4` to render `1/2`, while `Statevec.draw` rendered `√2/2` at `max_denominator=2` because the squared form `1/2` happened to fit.

I added a direct-rational branch at the top of `_recognize_sqrt`: if `Fraction(x).limit_denominator(max_denominator)` already matches `x` within tolerance, return `(signed_num, 1, den)` directly. The squared-form path remains as the fallback for genuine `√` values. Net effect: `max_denominator` now consistently bounds the denominator of the rendered value across both draw paths.

Extended `test_draw_max_denominator` with the `max_denominator=2` case for both `DensityMatrix` and `Statevec` to lock that in:

```python
assert dm.draw(OutputFormat.Unicode, max_denominator=2) == "[ 1/2  1/2 ]\n[ 1/2  1/2 ]"
assert sv.draw(OutputFormat.Unicode, max_denominator=2) == "√2/2(|0⟩ + |1⟩)"
```

The full local `test_pretty_print.py` suite still passes (118/118), and I also re-ran `test_density_matrix.py` and `test_statevec_backend.py` as a sanity sweep — no regressions. Let me know if anything else needs tightening.

pranav97nair (approved):
Thanks for the update !

pranav97nair:
Hi @Vinny010 , before we merge can you please update `CHANGELOG.md` with an entry at the top briefly summarizing what is being added in this PR?

Vinny010:
Done — added a CHANGELOG entry under `## Unreleased` / `### Added` (commit e597b27) summarizing the new `statevec_to_str` / `density_matrix_to_str` pretty-printing. Thanks again for the thorough review, and ready to merge whenever you are!

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Added new attribute `OpenGraph.output_cliffords`
   - Added `clifford` abstract method to `AbstractMeasurement`. Implemented it for `Plane` and `Axis`.
 
+- #524: Added pretty-printing for `Statevec` and `DensityMatrix` via new functions `statevec_to_str` and `density_matrix_to_str` (and the corresponding `__str__` methods), with ASCII, Unicode, and LaTeX output and recognition of common amplitudes (rational multiples, square roots, and factored uniform magnitudes).
+
```
pranav97nair, line 27: I would prefer this to be more concise - you can take out the part about `__str__` methods and recognition of common amplitudes, and please mention the `draw` functions.

Vinny010: Good call — updated in 1c690f2. Trimmed it down and led with the Statevec.draw / DensityMatrix.draw methods (keeping the underlying statevec_to_str / density_matrix_to_str functions). Thanks!

pranav97nair: Thanks! I think we are ready to merge as soon as the CI finishes running.

thierry-martinez:
Merged! Thank you again for your contribution.

PR #526, opened 2026-06-03
Extract Pauli flow from XZ-corrections (closes #432)

### LINKED ISSUE #432: Pauli-flow extraction from pattern
_This issue is part of unitaryHACK 2026. You have to be [registered](https://unitaryhack.dev/register/) to participate. Please read carefully the [unitaryHack rules](https://unitaryhack.dev/hacker-guide/) before submitting a PR for this issue._

_A note about [AI Slop](https://en.wikipedia.org/wiki/AI_slop): while we are open to collaboration with LLMs for unitaryHACK, fully AI-generated PRs are not acceptable. It is at the maintainers' discretion whether or not LLM-generated PRs are the right fit for the issues, and those that appear fully AI-generated may be immediately rejected. Please read carefully the [AI policy](https://unitaryhack.dev/ai-guide/) if you are using LLMs for your work._

## Context

Theorems 1, 2 and 4 in Ref. [^1] prescribe how to recover an $XZ$-correction strategy (and therefore,  an MBQC pattern) from a causal, g- or Pauli flow.  The class `XZCorrections` has methods `to_causal_flow` and `to_gflow` which reverse this prescription to construct, respectively,  `CausalFlow` and `GFlow` objects from a given set of $XZ$ corrections. However, the method `to_pauli_flow` is missing.

Extracting a Pauli flow from a deterministic $XZ$ correction strategy does not seem trivial, since the anachronical corrections do not appear in the pattern (compare Eq. in Theorem 2 vs Eq. in Theorem 4 in [^1]).

### Examples

A simple pattern with causal flow implementing a Hadamard gate:

```python
from graphix.command import N, E, M, X
from graphix import Measurement, Pattern
p = Pattern(input_nodes=[0], cmds=[N(1), E((0, 1)), M(0, Measurement.XY(0)), X(1, {0})], output_nodes=[1])
cf = p.extract_causal_flow()
print(cf)  # c(0) = {1}; {0} < {1}
```

A more complicated pattern with gflow:

```python
from graphix.command import N, E, M, X, Z
from graphix import Measurement, Pattern

p = Pattern(
    input_nodes=[0],
    cmds=[
        N(1), N(2), N(3),
        E((0, 1)), E((0, 2)), E((1, 2)), E((1, 3)),
        M(0, Measurement.XY(0.1)), X(2, {0}), X(3, {0}),
        M(1, Measurement.XZ(0.2)), Z(2, {1}), Z(3, {1}), X(2, {1})],
    output_nodes=[2, 3])

gf = p.extract_gflow()
print(gf) # g(0) = {2, 3}, g(1) = {1, 2}; {0, 1} < {2, 3}
```
This pattern does not have a causal flow because measurement planes are not restricted to $XY$. Calling `p.extract_causal_flow()` will raise a `FlowPropositionError`.

The following pattern does not have a causal flow or a gflow, but it's compatible with a Pauli flow:

```python
from graphix.command import N, E, M, X, Z
from graphix import Measurement, Pattern

p = Pattern(
    input_nodes=[0],
    cmds=[
        N(1), N(2), N(3), E((0, 1)), E((1, 2)), E((2, 3)),
        M(0, Measurement.X), X(3, {0}),
        M(1, Measurement.X), Z(3, {1}),
        M(2, Measurement.X), X(3, {2})],
    output_nodes=[3])

# p.extract_causal_flow() # Raises exception
# p.extract_gflow() # Raises exception

pf = p.extract_opengraph().extract_pauli_flow()
print(pf) # p(0) = {1, 3}, p(1) = {2}, p(2) = {3}; {0} < {1} < {2} < {3}
```

The question that we are trying to address is how to extract the Pauli flow _directly_ from the pattern, without passing through the underlying open graph. This is because, in general, it is not guaranteed that _a_ Pauli flow of the underlying open graph of the pattern generates the pattern itself because the Pauli flow of an open graph is not unique.

To generate patterns with Pauli flow but no gflow (useful for testing), one can start from an `OpenGraph` with Pauli flow, and convert it to a pattern. In general, any open graph with Pauli measurements does not have a causal flow or a gflow, but it may have a Pauli flow:

```python
from graphix import OpenGraph, Measurement
import networkx as nx
og = OpenGraph(
        graph=nx.Graph([(0, 2), (2, 4), (3, 4), (4, 6), (1, 4), (1, 6), (2, 3), (3, 5), (2, 6), (3, 6)]),
        input_nodes=[0],
        output_nodes=[5, 6],
        measurements={
            0: Measurement.XY(0.1),
            1: Measurement.XZ(0.1),
            2: Measurement.Y,
            3: Measurement.XY(0.1),
            4: Measurement.Z,
        },
    )
og.to_pattern()
```

Method `draw` in `Pattern`, `PauliFlow` and `XZCorrections` objects may be useful to understand the problem better.

## Task

Implement a method `XZCorrections.to_pauli_flow`, and define a method `extract_pauli_flow` in `Pattern` that calls it for convenience (analogous to `Pattern.extract_causal_flow` and `Pattern.extract_gflow`).

## Acceptance criteria

For the PR to be considered for review:

- It should pass the CI, including linting and type-checking. 

- New functions should be tested and documented appropriately following [numpy style](https://numpydoc.readthedocs.io/en/latest/format.html).

- Unitary Hack [AI use guidelines](https://unitaryhack.dev/ai-guide/) should be respected.

- Author should be ready to engage in a feedback loop.

- PR description should explicitly explain how the difficulty of dealing with anachronical  corrections in the Pauli flow was tackled.

For the PR to be accepted:

- All provided feedback should have been implemented.

[^1]: Browne, Kashefi, Mhalla and Perdrix, 2007 ([arXiv:quant-ph/0702212](https://arxiv.org/abs/quant-ph/0702212)).

Vinny010:
Closes #432.

Implements `XZCorrections.to_pauli_flow` and the convenience method `Pattern.extract_pauli_flow` (analogous to `extract_causal_flow` / `extract_gflow`), reconstructing a Pauli flow **directly from the pattern's XZ-corrections** (Theorem 4 of Browne et al. 2007) rather than from the underlying open graph — whose Pauli flow is not unique and is not guaranteed to generate the pattern.

## How the anachronical corrections were tackled (the crux of the issue)

The hard part, as the issue notes, is that a Pauli flow's correction sets may contain **anachronical corrections**: corrections targeting **X/Y Pauli-measured nodes that lie in the present or past** of the corrected node. `PauliFlow.to_corrections` discards these (the `correcting_set & future` filter), so they never appear in the pattern and cannot be read off the corrections — they must be *reconstructed*.

For each measured node `i`, reconstruction is cast as a **linear system over GF(2)**:

- **Pinned future part.** For nodes in the future of `i`, membership in the correction set `p(i)` is fixed by the observed X-corrections of `i` (and must reproduce the Z-corrections via the odd neighbourhood). These become constants of the system.
- **Free anachronical variables.** The unknowns are exactly the anachronical candidates — non-future nodes measured along the X or Y axes (permitted in `p(i)` by P1) — and, where the local proposition allows it, `i` itself.
- **Equations.** The odd-neighbourhood constraints: the Z-corrections on the future nodes, the vanishing of the odd neighbourhood on past non-(Y/Z) nodes (P2), the membership/odd-neighbourhood coupling on past Y nodes (P3), and the local proposition on `i` (P4–P9, handling the XY/XZ/YZ planes and the X/Y/Z axes).

The system is solved with a small GF(2) Gaussian-elimination helper (`_solve_gf2`); failure to solve at any node means no Pauli flow is compatible with the corrections. The resulting flow is then validated by `PauliFlow.check_well_formed`.

## Correctness

The decisive check is the **round trip**: for the reconstructed `pf`, `pf.to_corrections()` must reproduce the pattern's X- and Z-corrections exactly (this is what guarantees it generates *this* pattern). Tests verify well-formedness **and** the round trip on:

- the three worked examples of the issue (causal, gflow and Pauli; the reconstructed correction functions match the expected `p(0) = {1, 3}, p(1) = {2}, p(2) = {3}` etc., including the anachronical node `1`),
- a Pauli-measured open graph, and
- a randomized family of open graphs that admit a Pauli flow.

There are also unit tests for the GF(2) solver. Passes `ruff`, `mypy --strict`, `pyright`, and `pytest` locally (new tests plus the existing flow / open-graph / pattern suites).

*Developed with LLM assistance, reviewed and tested by me.*


graphix/flow/core.py
```diff
@@ -1391,3 +1436,164 @@ def _check_flow_general_properties(flow: PauliFlow[_AM_co]) -> None:
     o_set = set(flow.og.output_nodes)
     if first_layer != o_set or not first_layer:
         raise PartialOrderLayerError(PartialOrderLayerErrorReason.FirstLayer, layer_index=0, layer=first_layer)
+
+
+def _solve_gf2(matrix: list[list[int]], rhs: list[int], n_vars: int) -> list[int] | None:
```
thierry-martinez: There is already a GF(2)-linear-system solver implemented in `_linalg.solve_f2_linear_system`.

tests/test_pauli_flow_extraction.py
```diff
@@ -0,0 +1,200 @@
...
+            4: Measurement.Z,
+        },
+    )
+    _assert_round_trip(og.to_pattern())
+
+
+_MEASUREMENTS: list[Callable[[Generator], Measurement]] = [
+    lambda r: Measurement.XY(round(float(r.random()), 3)),
+    lambda r: Measurement.XZ(round(float(r.random()), 3)),
+    lambda r: Measurement.YZ(round(float(r.random()), 3)),
```
thierry-martinez: Why do you choose to apply `round` there rather than leave the random float unchanged?

tests/test_pauli_flow_extraction.py
```diff
@@ -0,0 +1,200 @@
...
+        outputs = nodes[:n_out]
+        inputs = nodes[n_out : n_out + n_in]
+        measurements = {
+            m: _MEASUREMENTS[int(rng.integers(0, len(_MEASUREMENTS)))](rng) for m in nodes if m not in outputs
+        }
+        try:
+            pattern = OpenGraph(
+                graph=graph, input_nodes=inputs, output_nodes=outputs, measurements=measurements
+            ).to_pattern()
+        except Exception:  # noqa: BLE001, S112  open graph without a flow -> not a valid test case
```
thierry-martinez: What are the relevant exceptions here? We should at least catch `OpenGraphError`. Are there any others?

graphix/flow/core.py
```diff
@@ -266,6 +266,51 @@ def to_gflow(self: XZCorrections[_PM_co]) -> GFlow[_PM_co]:
...
+        -----
+        See Theorem 4 in Ref. [1] and Definition 5 therein for the Pauli-flow propositions.
+
+        References
+        ----------
+        [1] Browne et al., 2007 New J. Phys. 9 250 (arXiv:quant-ph/0702212).
+        """
+        correction_function = _reconstruct_pauli_correction_function(self)
+        pf: PauliFlow[_AM_co] = PauliFlow(self.og, correction_function, self.partial_order_layers)
+        pf.check_well_formed()  # Raises a `FlowError` if the reconstructed flow is not well formed.
```
thierry-martinez: Is it possible to fail at this point? Can you give an example where the `correction_function` isn’t well‑formed by construction?

thierry-martinez (commented):
Thank you very much for this clean solution. Here are my initial comments on your PR.

Vinny010:
Thanks for the thoughtful review, @thierry-martinez! Pushed [9a73dc7](https://github.com/TeamGraphix/graphix/pull/526/commits/9a73dc732c6fd909076c69d403ea9579b6cfca46) addressing each point:

- **`_linalg.solve_f2_linear_system` reuse** — good catch. Switched to it. The per-node augmented matrix `[A | b]` is reduced to row echelon form with `MatGF2.gauss_elimination(ncols=n_vars)` (which propagates the row operations to the RHS column), inconsistency is detected by scanning for a `[0…0 | 1]` row, and the reduced system is then handed to the existing solver. `_solve_gf2` is removed.
- **`round` on the random angle** — purely cosmetic (kept the test failure messages readable when I was iterating). Dropped.
- **Exception narrowing in the randomized test** — narrowed to `OpenGraphError`, which is the only documented raise condition of `OpenGraph.to_pattern` when no flow exists.
- **`pf.check_well_formed()` on line 311** — by construction the GF(2) equations of `_reconstruct_pauli_correction_function` encode P2–P9 exactly, and restricting the anachronical candidates to X/Y-measured nodes guarantees P1; the general flow properties are satisfied by construction too. So I cannot construct an example where the check fails — it’s effectively a regression guard for the algorithm. Added a comment explaining that. Happy to drop the call entirely if you’d prefer.

Let me know if you’d like any other changes.

veena-vee989:
Hello - you are currently over the PR limit for unitaryHACK, so your PRs will be disregarded until you align with the rules. Reach out to hack@unitary.foundation to be reinstated.

graphix/flow/core.py
```diff
@@ -1391,3 +1447,132 @@ def _check_flow_general_properties(flow: PauliFlow[_AM_co]) -> None:
...
+    return correction_set
+
+
+def _reconstruct_pauli_correction_function(xz: XZCorrections[AbstractMeasurement]) -> dict[int, set[int]]:
+    """Reconstruct a Pauli-flow correction function from XZ-corrections.
+
+    See :meth:`XZCorrections.to_pauli_flow`.
+    """
+    og = xz.og
+    adjacency: dict[int, set[int]] = {n: set(og.graph.neighbors(n)) for n in og.graph.nodes}
```
matulni:
```suggestion
    adjacency: dict[int, set[int]] = {n: og.neighbors({n}) for n in og.graph.nodes}
```

graphix/flow/core.py
```diff
@@ -1391,3 +1447,132 @@ def _check_flow_general_properties(flow: PauliFlow[_AM_co]) -> None:
...
+    # row operations propagate to the right-hand side. Inconsistent systems leave a row
+    # ``[0...0 | 1]`` after reduction, which signals that no Pauli flow exists.
+    augmented = np.array([[*row, c] for row, c in zip(matrix, rhs, strict=True)], dtype=np.uint8).view(MatGF2)
+    augmented = augmented.gauss_elimination(ncols=n_vars)
+    lhs = MatGF2(augmented[:, :n_vars])
+    b = augmented[:, n_vars]
+    for i in range(lhs.shape[0]):
+        if not lhs[i].any() and b[i] != 0:
+            return None
+    solution = solve_f2_linear_system(lhs, MatGF2(b))
```
matulni, line 1560:
```suggestion
    # `solve_f2_linear_system` does not check if a solution exists or if `lhs` is in REF.
    solution = solve_f2_linear_system(lhs, MatGF2(b))
```

graphix/flow/core.py
```diff
@@ -1391,3 +1447,132 @@ def _check_flow_general_properties(flow: PauliFlow[_AM_co]) -> None:
...
+
+    correction_set = set(fixed_in_p)
+    correction_set.update(v for v, bit in zip(variables, solution, strict=True) if int(bit))
```
matulni:
```suggestion
    correction_set.update(v for v, bit in zip(variables, solution, strict=True) if bit)
```

graphix/flow/core.py
```diff
@@ -1391,3 +1447,132 @@ def _check_flow_general_properties(flow: PauliFlow[_AM_co]) -> None:
...
+    # Self-membership in the correction set, dictated by the local proposition (P4-P9).
+    self_fixed_in = label in {Plane.XZ, Plane.YZ, Axis.Z}
+    self_is_var = label in {Axis.X, Axis.Y} and node in non_inputs
+    if self_fixed_in and node not in non_inputs:
+        return None  # `node` must correct itself but is an input node.
+
+    # Anachronical candidates: non-future, X/Y-measured, non-input nodes (other than `node`).
+    nonfuture_others = nodes - future - {node}
+    candidates = sorted(a for a in nonfuture_others if a in non_inputs and labels.get(a) in {Axis.X, Axis.Y})
+    variables = [*candidates, node] if self_is_var else list(candidates)
```
matulni: I don't think sorting is necessary, but if it is, please add a test which would fail if candidates are not sorted.

```suggestion
    variables = [a for a in nonfuture_others | {node} if a in non_inputs and labels.get(a) in {Axis.X, Axis.Y}]
```

With the suggestion above, `self_is_var` is no longer necessary.

graphix/flow/core.py
```diff
@@ -1391,3 +1447,132 @@ def _check_flow_general_properties(flow: PauliFlow[_AM_co]) -> None:
...
+
+    # P3: a non-future Y-measured node `g` must lie outside the closed odd neighbourhood of the
+    # correction set, i.e. its membership and odd-neighbourhood membership must coincide.
+    for g in nonfuture_others:
+        if labels.get(g) == Axis.Y:
+            row = row_at(g)
+            if g in var_index:
+                row[var_index[g]] ^= 1
+            matrix.append(row)
+            rhs.append(const_at(g))
```
matulni: I believe the following is equivalent and more clear:

```suggestion
    for g in nonfuture_others:
        lab_g = labels.get(g)

        # `nonfuture_others` never contains output nodes
        assert lab_g is not None

        # P2: the odd neighbourhood vanishes on non-future, non-(Y/Z) nodes.
        if lab_g not in {Axis.Y, Axis.Z}:
            matrix.append(row_at(g))
            rhs.append(const_at(g))

        # P3: a non-future Y-measured node `g` must lie outside the closed odd neighbourhood of the
        # correction set, i.e. its membership and odd-neighbourhood membership must coincide.
        elif lab_g == Axis.Y:
            row = row_at(g)
            if g in var_index:
                row[var_index[g]] ^= 1
            matrix.append(row)
            rhs.append(const_at(g))
```

graphix/flow/core.py
```diff
@@ -1391,3 +1447,132 @@ def _check_flow_general_properties(flow: PauliFlow[_AM_co]) -> None:
...
+        fixed_in_p.add(node)
+
+    def const_at(g: int) -> int:
+        return len(adjacency[g] & fixed_in_p) % 2
+
+    def row_at(g: int) -> list[int]:
+        return [1 if v in adjacency[g] else 0 for v in variables]
+
+    matrix: list[list[int]] = []
+    rhs: list[int] = []
```
matulni, line 1498: Could you please add a general comment explaining that `matrix` encodes the constraints on the odd neighbourhood of the correction function, and explain the meaning of having a 0 or 1 at the corresponding `rhs` ?

graphix/flow/core.py
```diff
@@ -1391,3 +1447,132 @@ def _check_flow_general_properties(flow: PauliFlow[_AM_co]) -> None:
...
+        for node in layer:
+            future_of[node] = set(accumulated)
+        accumulated |= set(layer)
+
+    correction_function: dict[int, set[int]] = {}
+    for node in og.measurements:
+        correction_set = _solve_pauli_correction_set(xz, node, future_of[node], adjacency, labels)
+        if correction_set is None:
+            # No correction set reconciles the XZ-corrections with the Pauli-flow propositions.
+            raise FlowError
```
matulni: It may be worth adding a new `FlowGenericErrorReason` in `flow.exceptions.py` to pass your comment as an exception message.

graphix/flow/core.py
```diff
@@ -266,6 +268,60 @@ def to_gflow(self: XZCorrections[_PM_co]) -> GFlow[_PM_co]:
...
+        [1] Browne et al., 2007 New J. Phys. 9 250 (arXiv:quant-ph/0702212).
+        """
+        correction_function = _reconstruct_pauli_correction_function(self)
+        pf: PauliFlow[_AM_co] = PauliFlow(self.og, correction_function, self.partial_order_layers)
+        # Defensive: by construction the GF(2) equations of `_reconstruct_pauli_correction_function`
+        # encode propositions P2-P9 exactly, and the anachronical candidates are restricted to X/Y
+        # axes which guarantees P1; the general flow properties are also satisfied by construction
+        # (the correction function is defined on the measured nodes and its image is included in
+        # the non-input nodes). This check is kept as a regression guard.
+        pf.check_well_formed()
```
matulni: I agree with @thierry-martinez. This seems to be a purely defensive test, I'd be inclined to remove it and eventually add a more comprehensive test suite.

tests/test_pauli_flow_extraction.py
```diff
@@ -0,0 +1,184 @@
...
+]
+
+
+@pytest.mark.filterwarnings("ignore:Open graph with non-inferred Pauli measurements.")
+def test_extract_pauli_flow_randomized_round_trip() -> None:
+    # Generate random open graphs; those that admit a Pauli flow (so that `to_pattern`
+    # succeeds) are converted to a pattern, and the reconstructed flow is checked to be
+    # well formed and to reproduce the pattern's corrections.
+    tested = 0
+    for seed in range(400):
```
matulni: Having the seed as an internal parameter makes it difficult to debug. Could you please pass it as `pytest.mark.parametrize` variable ?

tests/test_pauli_flow_extraction.py
matulni, line 1: In addition to these tests, could you please mirror the tests `test_extract_causal_flow_rnd_circuit`, `test_extract_gflow_rnd_circuit`, `test_extract_causal_flow`, and `test_extract_gflow` in `test_pattern.py`?

tests/test_pauli_flow_extraction.py
```diff
@@ -0,0 +1,184 @@
...
+        except OpenGraphError:
+            # The randomly drawn open graph does not admit a flow (the only documented
+            # raise condition of `OpenGraph.to_pattern`) -> not a valid test case.
+            continue
+        _assert_round_trip(pattern)
+        tested += 1
+    assert tested >= 30  # ensure the randomized sweep actually exercised the extraction
+
+
+def test_to_pauli_flow_raises_when_no_flow_exists() -> None:
```
matulni: Please use `pytest.mark.parametrize`. I'd appreciate if you could add a few more failure tests.

matulni (commented):
Thanks for this nice work! Here are some comments from my first pass. There's some work before merging, but in the mean time I'm ok with assigning you the issue.

For my own curiosity, would you be willing to share the promtps and skills (or even the reasoning log) of the LLM you used to code this ? This is of course not mandatory,  but I'm curious to see how the solution was found. Thanks!

tests/test_pauli_flow_extraction.py
```diff
@@ -0,0 +1,184 @@
...
+
+def test_extract_pauli_flow_pauli_example() -> None:
+    # The flow must include the anachronical correction (node 1 in p(0)) that does not
+    # appear in the pattern, in order to satisfy the X-axis proposition (P7).
+    pattern = _pauli_pattern()
+    assert _correction_function(pattern) == {0: {1, 3}, 1: {2}, 2: {3}}
+    _assert_round_trip(pattern)
+
+
+@pytest.mark.filterwarnings("ignore:Open graph with non-inferred Pauli measurements.")
```
thierry-martinez: I believe that such warnings should not be triggered here; and if they are, we should address them, because whether Pauli measurements are inferred (or not) affects the Pauli flow.

tests/test_pauli_flow_extraction.py
```diff
@@ -0,0 +1,184 @@
...
+    lambda r: Measurement.XY(float(r.random())),
+    lambda r: Measurement.XZ(float(r.random())),
+    lambda r: Measurement.YZ(float(r.random())),
+    lambda _r: Measurement.X,
+    lambda _r: Measurement.Y,
+    lambda _r: Measurement.Z,
+]
+
+
+@pytest.mark.filterwarnings("ignore:Open graph with non-inferred Pauli measurements.")
```
thierry-martinez: Same comment as above: these warnings should not be triggered, and they are relevant with respect to Pauli flows.

graphix/flow/core.py
thierry-martinez: Actually, the current implementation contains a case where this check can fail:`Pattern().extract_xzcorrections().to_pauli_flow()` raises `graphix.flow.exceptions.PartialOrderError: The partial order cannot be empty.`
However, this seems to be a bug in `check_well_formed` (see #531).
I agree with @matulni : these sanity checks should be part of the test-suite and should not be performed systematically in production.

Vinny010:
Thanks, @thierry-martinez — both points addressed in the latest commit.

**Non-inferred Pauli warning suppressions.** You're right that these shouldn't be silenced. I removed both `@pytest.mark.filterwarnings("ignore:Open graph with non-inferred Pauli measurements.")` markers. I also checked *why* they were there: it turns out the warning doesn't actually fire for either test — the open graphs use explicit `PauliMeasurement` (`Measurement.Y`/`Measurement.Z`) and non-Pauli Bloch angles, so there are no "secretly Pauli" Bloch measurements. I verified this by promoting that exact warning to an error (`-W "error:Open graph with non-inferred Pauli measurements.:UserWarning"`) and the whole file still passes. So the suppressions were dead weight masking a semantically relevant signal — gone now.

**`check_well_formed()` in production.** Agreed with you and @matulni — removed the `pf.check_well_formed()` call from `XZCorrections.to_pauli_flow`. The reconstructed flow is well formed by construction (the GF(2) system encodes P2-P9, the anachronical candidates are X/Y-restricted for P1, and the image/domain conditions hold), and the well-formedness is now exercised only in the test-suite via the existing `assert pf.is_well_formed()` in the round-trip helper.

This also removes the production manifestation of #531: `Pattern().extract_xzcorrections().to_pauli_flow()` no longer raises `PartialOrderError` and instead returns the trivial flow. I added `test_to_pauli_flow_empty_pattern` to lock that in. (The underlying `check_well_formed` behaviour on an empty partial order is still the separate concern tracked in #531.)

Vinny010:
Thanks for the careful review, @matulni — all addressed in the latest push.

**Reconstruction (`core.py`).**
- Took your `variables` suggestion: it now iterates `nonfuture_others | {node}` with the single filter, which drops both the sorting and `self_is_var`. (No sorting-dependent test needed since the order no longer matters.)
- Merged the P2 and P3 loops into the single `for g in nonfuture_others` form you proposed, with the `assert lab_g is not None` — confirmed sound: output nodes are layer 0 of the partial order, so they're in every measured node's `future` and never reach `nonfuture_others`.
- Added the general comment explaining that each row is the indicator of which `variables` lie in a node's neighbourhood, so `row · p` is its odd-neighbourhood parity, and that `rhs` is the required parity (0/1) after folding in the fixed part via `const_at`.
- Applied the `adjacency = {n: og.neighbors({n}) ...}`, the `solve_f2_linear_system` comment, and the `if bit` tidy-ups.

**Typed error.** Added `FlowGenericErrorReason.NoPauliFlow` and the reconstruction now raises `FlowGenericError(NoPauliFlow)` instead of a bare `FlowError`, carrying your "no correction set reconciles the XZ-corrections" message.

**Defensive check.** Removed `pf.check_well_formed()` from the production `to_pauli_flow` path (agreeing with you and @thierry-martinez); the well-formedness is asserted in the test-suite instead.

**Tests.**
- The randomized round-trip is now `@pytest.mark.parametrize("seed", range(400))` — one independently reproducible case per seed, skipping draws with no edges / no flow.
- The failure test is parametrized and asserts the typed `FlowGenericErrorReason.NoPauliFlow`, with added XZ- and YZ-input cases alongside the original Z-input and isolated-XY ones.
- Mirrored the extraction tests in `test_pattern.py`: `test_extract_pauli_flow_rnd_circuit` (the random-circuit round trip) and `test_extract_pauli_flow` over the worked cases. For the worked cases I assert the round trip on every `has_gflow` case (a gflow always induces a Pauli flow) and skip the non-gflow ones, since a Pauli flow is strictly more general and its absence can't be inferred from `has_gflow` alone — happy to add a dedicated `has_pflow` field if you'd prefer the negative cases covered too.

On the LLM question — I did use an AI coding assistant (Claude) and then verified everything myself by running the full test suite, which is how the edge cases (the anachronical corrections, the empty-pattern path) got shaken out. I'd rather keep my specific prompts and workflow to myself, but appreciate the curiosity — and I'm glad to keep iterating on the PR the normal way.

Vinny010:
Hi @thierry-martinez @matulni — quick ready-for-re-review ping with the unitaryHACK deadline tomorrow.

I addressed your Jun 10 review feedback in my Jun 10–11 replies. CI is fully green (24/24) and mergeable state is clean. Whenever you have a moment, would appreciate a final look — happy to make any last tweaks if anything else needs attention.

Thanks for the careful review on this one!

thierry-martinez (commented):
After careful consideration, we've decided to award you the bounty for this PR as well. Please leave a comment on issue #432 so we can assign it to you. In the meantime, take a look at the code‑coverage results: https://app.codecov.io/gh/TeamGraphix/graphix/pull/526

graphix/flow/core.py
```diff
@@ -1391,3 +1447,145 @@ def _check_flow_general_properties(flow: PauliFlow[_AM_co]) -> None:
...
+    # Reduce the augmented matrix ``[A | b]`` to row echelon form together so that the
+    # row operations propagate to the right-hand side. Inconsistent systems leave a row
+    # ``[0...0 | 1]`` after reduction, which signals that no Pauli flow exists.
+    augmented = np.array([[*row, c] for row, c in zip(matrix, rhs, strict=True)], dtype=np.uint8).view(MatGF2)
+    augmented = augmented.gauss_elimination(ncols=n_vars)
+    lhs = MatGF2(augmented[:, :n_vars])
+    b = augmented[:, n_vars]
+    for i in range(lhs.shape[0]):
+        if not lhs[i].any() and b[i] != 0:
+            return None
```
matulni:
```suggestion
    for coeffs, const in zip(lhs, b, strict=True):
        if not np.any(coeffs) and const:
            return None
```

matulni (approved):
Thanks for the nice job, LGTM!

(Please wait for a second approval).

graphix/flow/core.py
Vinny010: Applied in e7fa8ab — thanks, that reads much better. Iterating the rows directly with `zip(lhs, b, strict=True)` and `np.any` drops the index bookkeeping entirely. And thanks for the approval!

matulni:
Hi @Vinny010, before we merge could you please update the CHANGELOG?

Vinny010:
Done — added a CHANGELOG entry under `## Unreleased` / `### Added` (commit 0ac8592) for the new `XZCorrections.to_pauli_flow` method and the `NoPauliFlow` error reason. Thanks, @matulni — ready to merge on your side.

graphix/flow/core.py
```diff
@@ -1391,3 +1447,145 @@ def _check_flow_general_properties(flow: PauliFlow[_AM_co]) -> None:
...
+    # neighbourhood of `g`, i.e. whether `g` belongs to the odd neighbourhood of that part. The
+    # matching `rhs` is the required parity of that membership (1 = `g` must be in the odd
+    # neighbourhood, 0 = it must not), after folding in `const_at(g)`, the fixed part's contribution.
+    matrix: list[list[int]] = []
+    rhs: list[int] = []
+
+    # Odd-neighbourhood constraints on the future nodes (the Z-corrections of `node`).
+    for g in future:
+        matrix.append(row_at(g))
+        rhs.append((1 if g in z_corr else 0) ^ const_at(g))
```
matulni, line 1503: Sorry for the never-ending stream of comments, but I think the following is equivalent and reduces the number of equations. Do you agree @Vinny010, @thierry-martinez ?

If not, would it be possible to add a test that fails in this scenario @Vinny010 ?

```suggestion
    for g in future:
        lab_g = labels.get(g)
        if lab_g and lab_g not in {Axis.Y, Axis.Z}:
            matrix.append(row_at(g))
            rhs.append((1 if g in z_corr else 0) ^ const_at(g))
```

Vinny010: No need to apologise — the thoroughness is appreciated. Agreed, and verified before applying (9aa5fb0):

- Ran the full suite with the change: the 400-seed randomized round-trip (which draws every measurement basis, so Y/Z-measured nodes and outputs do appear as future nodes), the worked examples (`test_extract_pauli_flow_pauli_opengraph` has a Y node, a Z node, and two outputs), and the no-flow cases all pass; `ruff` and `mypy` stay clean.
- Since `to_corrections()` reproducing the XZ-corrections is exactly the correctness criterion, the absence of any round-trip failure is good evidence those equations are redundant for output / Pauli-Y/Z future nodes — so I couldn't construct the failing scenario.

My reading of *why* they're redundant: a future node that is an output or measured in the Y/Z Pauli basis doesn't impose an independent Z-correction constraint on `p(node)` here — its odd-neighbourhood relationship is already determined by the rest of the system rather than by this equation. I'm leaning mainly on the empirical round-trip for that, though, so please correct me if the formal justification is different. Either way the smaller system is a nice cleanup. Thanks!

Vinny010: I have to walk back my earlier "agreed" — after testing this more widely, the reduction is **not** equivalent, so I've reverted it (27fd0b7) and added a regression test. Apologies for the premature sign-off.

What I found: instrumenting the dropped branch across the suite (and an ~8k-seed sweep) showed that the skipped future-node equations are, in practice, **all for output nodes** — a Y/Z-measured node never actually lands in a `future` set in any generated flow, so that part of the condition is never exercised. But the **output** equations are sometimes load-bearing: when a Z-correction targets an output node, dropping its equation relaxes the system and the reconstructed flow stops reproducing the Z-corrections. The 0–399 seed range in the existing test just happened not to contain such a case.

Minimal deterministic counterexample, distilled from a randomized failure and added as `test_extract_pauli_flow_output_zcorrection`:

```python
og = OpenGraph(
    graph=nx.Graph([(0, 1), (0, 4), (0, 5), (0, 7), (1, 3), (1, 4), (2, 4), (2, 5), (3, 7), (4, 6), (6, 7)]),
    input_nodes=[],
    output_nodes=[3, 0, 5],
    measurements={1: Measurement.X, 2: Measurement.X, 4: Measurement.YZ(0.3), 6: Measurement.Z, 7: Measurement.Y},
)
_assert_round_trip(og.to_pattern())
```

With the full future loop this round-trips; with the reduction, `to_corrections()` no longer matches `z_corrections` (`z_ok=False`). So I've kept the original loop. Thanks for pushing on it — good instinct to ask for the failing test.

matulni: Thanks, good catch. That statement also checks consistency with the $z$ map

thierry-martinez (commented):
Sorry for the delay; I am still checking that the test suite is comprehensive enough.

One problem I have with the current test suite is that the following implementation of `XZCorrections.to_pauli_flow` passes all the tests:
```python
    def to_pauli_flow(self) -> PauliFlow[_AM_co]:
        pf = self.og.find_pauli_flow()
        if pf is None:
            raise FlowGenericError(FlowGenericErrorReason.NoPauliFlow)
        return pf
```

It is obvious that this implementation is incorrect: the Pauli flow implemented by the pattern is not necessarily the same as the Pauli flow inferred from the underlying open graph, since the Pauli flow is not unique.

Could you come up with a test that exercises this property?

Vinny010:
Thanks @thierry-martinez — addressed in 377b25b on the rebased branch.

You were right that the existing round-trip tests don't distinguish the real reconstruction from the trivial `pf = self.og.find_pauli_flow()` impl: every test case happens to feed in patterns whose flow already coincides with the maximally-delayed flow returned by `find_pauli_flow`, so the trivial impl rounds-trips perfectly on all of them.

The new test `test_extract_pauli_flow_pins_the_pattern_specific_flow` pins down a small open graph (path 0-1-2-3, all X-measured, output 3) that admits two distinct, well-formed Pauli flows whose `to_corrections()` outputs genuinely differ:

- `pf_with_anachronical` = {0: {1}, 1: {2}, 2: {3}} — node 0's correction is anachronical-only ({1}), no future X-correction on the output.
- `pf_with_future` = {0: {1, 3}, 1: {2}, 2: {3}} — node 0 X-corrects the output.

It then builds the XZ-corrections of each, asserts the reconstruction returns the chosen flow exactly, and asserts both round-trip back to their original corrections. The discriminator: any trivial impl that ignores the XZ-corrections returns the same flow for both inputs, so it can match at most one of them — the assertion fires for whichever case the trivial impl misses.

I verified this empirically by monkey-patching the trivial impl in: with that swap, the five existing scalar tests (`causal`, `gflow`, `pauli_example`, `pauli_opengraph`, `output_zcorrection`) still pass, and only the new test fails — confirming it discriminates exactly the property you flagged.

Also: rebased the branch onto current `master` (small CHANGELOG conflict against #524, resolved by keeping both entries). CI green locally — 84 / 84 pauli-flow-extraction tests + the rest of the flow + pattern suites (422 / 422 in the slice I ran).

thierry-martinez (approved):
Thank you again for your contribution. Sorry for the reviewing delay. You've my approval. Could you solve the merge conflict before we merge?

PR #541, opened 2026-06-13
Pin `matplotlib==3.10.9` for extra `dev` dependencies

thierry-martinez:
This commit adds a pin to `matplotlib==3.10.9` in the `dev` extra dependencies.

Matplotlib 3.11.0 introduces visual changes that break our graphical-regression tests.  As we do for other packages whose new releases can break the test suite, we pin a precise version in the `dev` extra to ensure reproducible CI while keeping the package unpinned in the default dependencies so users can choose whichever version they prefer.

We pin `3.10.9` instead of the newer `3.11.0` because the latter drops support for Python 3.11, which we continue to support until its official end-of-life (October 2026):
https://devguide.python.org/versions/


thierry-martinez:
For instance, for `test_causal_flow_draw`, the figure produced by `matplotlib==3.11.0` is the following:
result
and the difference with the baseline is:
result-failed-diff

mgarnier59 (approved):

PR #549, opened 2026-06-26
Fix #531: output nodes in partial order layers

### LINKED ISSUE #531: [Bug]: `_check_flow_general_properties` rejects empty graphs
`_check_flow_general_properties` rejects empty graphs with `PartialOrderError(PartialOrderErrorReason.Empty)`, even though the formal definitions of flows do not exclude empty graphs. As a result, flow‑extraction procedures return an empty flow object for an empty graph, but that object then fails the well‑formedness check.

```python
def test_empty_graph() -> None:
    og = OpenGraph(graph=nx.Graph(), input_nodes=[], output_nodes=[], measurements={})
    pf = og.extract_causal_flow()
    pf.check_well_formed() # raises PartialOrderError(PartialOrderErrorReason.Empty)
```

**Edit:** I copy-pasted the wrong exception. In this example, `pf.check_well_formed()` raises `graphix.flow.exceptions.PartialOrderLayerError: The first layer of the partial order must contain all the output nodes of the open graph and cannot be empty. First layer: frozenset()`.

matulni:
This commit fixes #531 where it was noted that there exist well-formed flows **without** output nodes.
- The flow-finding routines are modified so that they never include an empty output nodes layer in the partial order. This design choice is in accordance with the behaviour for `XZCorrections`.
- The check-wellformedness methods do not systematically reject flows on open graphs without output nodes.


graphix/flow/_find_gpflow.py
```diff
@@ -583,7 +583,7 @@ def compute_partial_order_layers(correction_matrix: CorrectionMatrix[_AM_co]) ->
     Returns
     -------
     layers : tuple[frozenset[int], ...]
-        Partial order between corrected qubits in a layer form. The frozenset `layers[i]` comprises the nodes in layer `i`. Nodes in layer `i` are "larger" in the partial order than nodes in layer `i+1`. Output nodes are always in layer 0.
+        Partial order between corrected qubits in a layer form. The frozenset `layers[i]` comprises the nodes in layer `i`. Nodes in layer `i` are "larger" in the partial order than nodes in layer `i+1`. Output nodes are always in layer 0 when they exist.
```
thierry-martinez: This function is not part of the documentation, but we should try to keep the docstring syntax valid anyway. In particular, backquotes should be doubled (this impacts many other occurrences).
```suggestion
        Partial order between corrected qubits in a layer form. The frozenset ``layers[i]`` comprises the nodes in layer ``i``. Nodes in layer ``i`` are "larger" in the partial order than nodes in layer ``i+1``. Output nodes are always in layer 0 when they exist.
```

graphix/flow/core.py
```diff
@@ -585,20 +583,21 @@ def to_corrections(self) -> XZCorrections[_AM_co]:
...
+                for measured_node in layer:
+                    correcting_set = self.correction_function[measured_node]
+                    # Conditionals avoid storing empty correction sets
+                    if x_corrected_nodes := correcting_set & future:
+                        x_corrections[measured_node] = frozenset(x_corrected_nodes)
+                    if z_corrected_nodes := self.og.odd_neighbors(correcting_set) & future:
+                        z_corrections[measured_node] = frozenset(z_corrected_nodes)
 
-            future |= layer
+                future |= layer
```
thierry-martinez: This loop can be written more simply by iterating from layer 0, eliminating both the special case and the initial non‑emptiness test. Just initialize `future` with an empty set. The first iteration will be a no‑op because `future` is empty; on the next step `future |= layer` will copy the contents of layer 0 into `future`, without needing an explicit call to `copy`.

matulni: You're right, fixed in [015d75a](https://github.com/TeamGraphix/graphix/pull/549/commits/015d75a5da7d096a84923fb6e891dbc82668dcd3)

graphix/flow/_find_gpflow.py
matulni: Agreed. I did a regex search and replace for the `core.py` and `find_gpflow.py` files in [759b165](https://github.com/TeamGraphix/graphix/pull/549/commits/759b165727bf8b0f221e0b2c06e7f5852f952154) and [dbc222c](https://github.com/TeamGraphix/graphix/pull/549/commits/dbc222c37d89cc44fae647564bf824ecac10ffab), but I think beyond that is outside the scope of this PR.

thierry-martinez (approved):
LGTM! Thanks.

pranav97nair (approved):
Thanks for this update, Mateo! I have no other comments to add.

PR #536, opened 2026-06-10
fix(visualization): keep layer arrow in frame for linear flows

shantoshdurai:
Fixes #535

Linear open graphs place every node on one row, so `ymin == ymax`. Layer labels use fixed point offsets below `plot_lims.ymin`, but the ymin adjustment in `_draw_layers` ran before axis limits were set, so the data/display transform was degenerate and the "Layer" arrow ended up below the plot frame.

**Changes**
- Set preliminary plot limits before `_draw_layers` so point-based offsets map correctly
- Expand `ymin` to include layer label height (`va='top'`) plus padding
- Enforce a minimum vertical axis span in `_set_plot_lims` when node rows collapse to one line
- Add regression test `test_linear_causal_flow_layer_label_in_frame`

**Verify**
```bash
pytest tests/test_visualization.py::test_linear_causal_flow_layer_label_in_frame -q
```

PR #559, opened 2026-07-12
Fix #535: add visual padding below measurement-order arrow

### LINKED ISSUE #535: [Bug]: Measurement order arrow outside of frame in flow visualisation
The "Layer" arrow in flow visualisation appears outside the plot frame for linear graphs.
In general, we expect it to be at a fixed, constant distance between the lower open graph nodes and the plot frame.

Incorrect behavior:

```python
from graphix import OpenGraph, PauliExponential, Plane
import networkx as nx

og = OpenGraph(graph=nx.Graph([(0, 1), (1, 2), (2, 3)]), input_nodes=[0], output_nodes=[3], measurements=dict.fromkeys(range(3), Plane.XY))
flow = og.extract_causal_flow()

flow.draw()
```
Image

Correct behaviour (for non-linear open graphs):

```python
from graphix import OpenGraph, PauliExponential, Plane
import networkx as nx

og = OpenGraph(graph=nx.Graph([(0, 2), (2, 4), (1, 3), (3, 5), (2, 3)]), input_nodes=[0, 1], output_nodes=[4, 5], measurements=dict.fromkeys(range(4), Plane.XY))
flow = og.extract_causal_flow()

flow.draw()
```

Image

The potential fix should make methods `GraphVisualizer._set_plot_lims` and `GraphVisualizer._draw_layers` in the `visualization.py` module more robust.

thierry-martinez:
This commit adds padding beneath the measurement-order arrow when visualizing open graphs with layers, fixing #535 by keeping both the arrow and its label inside the visualization box.

This commit simplifies the approach taken in #536. We reuse the fact that the coordinate transformation employed in `_draw_layers` depends on the current plot limits set on the figure.

In this commit, we use `mpl` to track regression for #535. Because changing the coordinate‑transformation logic in `_draw_layers` affects other tests (independent of the padding itself), all baseline files have been updated accordingly.

This commit only fixes the bug reported in #535. It does not prevent other objects, such as flow arrows or the legend box, from appearing outside the visualization frame.


matulni (approved):
Thanks for the fix, LGTM!

(Reminder: CHANGELOG.md)

PR #557, opened 2026-07-07
Namespace homogenisation 

matulni:
This PR introduces a homogeneous and pythonic naming convention for transformation methods and accessors.

## Transformation methods

We fix the following convention:

- `.to_<object>` when the transformation can only return an `object` or raise an exception.  If `object` corresponds to an existing class, we use the name of the class without underscores and small case. Ex.: `to_opengraph`, `to_causalflow`, `to_xzcorrections`, etc.

Non-exhaustive list of method names (unless specified, the method name remained the same in the commit):
  - `Pattern.to_opengraph` (formerly `Pattern.extract_opengraph`)
  - `Pattern.to_xzcorrections` (formerly `Pattern.extract_xzcorrections`)
  - `Pattern.to_qasm3`
  - `Pattern.to_causalflow` (formerly `Pattern.extract_causal_flow`)
  - `Pattern.to_gflow` (formerly `Pattern.extract_gflow`)
  - `Pattern.to_pauliflow` (formerly `Pattern.extract_pauli_flow`)
  - `StandardizedPattern.to_pattern`
  - `StandardizedPattern.to_space_optimal_pattern`
  - `StandardizedPattern.to_opengraph` (formerly `StandardizedPattern.extract_opengraph`)
  - `StandardizedPattern.to_xzcorrections` (formerly `StandardizedPattern.extract_xzcorrections`)
  - `XZCorrections.to_pattern`
  - `XZCorrections.to_causalflow` (formerly `XZCorrections.to_causal_flow`)
  - `XZCorrections.to_gflow`
  - `XZCorrections.to_pauliflow` (formerly `XZCorrections.to_pauli_flow`)
  - `XZCorrections.to_dag` (formerly `XZCorrections.extract_dag`)
  - `PauliFlow.to_xzcorrections` (formerly `PauliFlow.to_corrections`)
  - `GFlow.to_xzcorrections` (formerly `GFlow.to_corrections`)
  - `CausalFlow.to_xzcorrections` (formerly `CausalFlow.to_corrections`)
  - `ExtractionResult.to_circuit`
  - `PauliString.to_tableau`
  - `CliffordMap.to_tableau`
  - `<...>.to_bloch`
  - `AbstractMeasurement.to_plane_or_axis`
  - `AbstractPlanarMeasurement.to_plane`
  - `Measurement.to_pauli_or_bloch`
  - `<...>.to_ascii`
  - `<...>.to_unicode`
  - `<...>.to_latex`
[...]

PR #561, opened 2026-07-13
Fix #555: Method `Pattern.draw` attempts to compute Pauli flow from pattern.

### LINKED ISSUE #555: [Bug]: `Pattern.draw` should attempt to compute the Pauli flow from pattern
Currently, it only attempts to calculate causal and gflow.

matulni:
Fix #555: Method `Pattern.draw` attempts to compute Pauli flow from pattern before fall back to flow extraction from open graph.


graphix/pattern.py
```diff
@@ -1535,37 +1535,36 @@ def draw(
         Raises
         ------
         PatternError
-            If the underlying opengraph does not have flow.
+            If the underlying open graph does not have flow.
 
         Notes
         -----
-        If ``flow_from_pattern==True`` but the pattern is not compatible with a gflow, an attempt to be extract the flow from the underlying open graph will be made while warning the user.
+        If ``flow_from_pattern==True`` but the pattern is not compatible with a causal flow, gflow, or Pauli flow, an attempt to be extract the flow from the underlying open graph will be made while warning the user.
```
thierry-martinez: The following typo already existed before the current PR: `to extract the flow` (without **be**). I suggest the following wording:
```suggestion
        If ``flow_from_pattern == True`` but the pattern is not compatible with a causal flow, gflow, or Pauli flow, a warning will be issued and an attempt will be made to extract the flow from the underlying open graph. If the underlying open graph also lacks a flow, the ``PatternError`` exception is raised.
```

graphix/pattern.py
```diff
@@ -1535,37 +1535,36 @@ def draw(
...
-                                try:
-                                    flow = xz_corrections.to_gflow()
-                                except FlowError:
-                                    warn(
-                                        "The pattern is not consistent with a causal flow or a gflow. An attempt to be extract the flow from the underlying open graph will be made.",
-                                        stacklevel=stacklevel,
-                                    )
-
+                                warn(
+                                    "The pattern is not consistent with a flow. An attempt to be extract the flow from the underlying open graph will be made.",
```
thierry-martinez, line 1572: The following typo already existed before the current PR: `to extract the flow` (without *be*). Moreover, codecov reveals this execution path is never tested. 

```suggestion
                                    "The pattern is not consistent with a flow. An attempt to extract the flow from the underlying open graph will be made.",
```

graphix/pattern.py
```diff
@@ -1535,37 +1535,36 @@ def draw(
...
         Notes
         -----
-        If ``flow_from_pattern==True`` but the pattern is not compatible with a gflow, an attempt to be extract the flow from the underlying open graph will be made while warning the user.
+        If ``flow_from_pattern==True`` but the pattern is not compatible with a causal flow, gflow, or Pauli flow, an attempt to be extract the flow from the underlying open graph will be made while warning the user.
         """
         if annotations is None:
             og = self.extract_opengraph()
             gv = GraphVisualizer.from_opengraph(og=og, **options)
         else:
+            xz_corrections = self.extract_xzcorrections()
```
thierry-martinez: Nit-picking: could we use `xzcorrections` for the variable name as well, following the convention of #557.

graphix/pattern.py
```diff
@@ -1584,8 +1583,7 @@ def draw(
                     gv = GraphVisualizer.from_flow(flow=flow, **options)
 
                 case DrawPatternAnnotations.XZCorrections:
-                    xzcorrections = self.extract_xzcorrections()
-                    gv = GraphVisualizer.from_xzcorrections(xz_corr=xzcorrections, **options)
+                    gv = GraphVisualizer.from_xzcorrections(xz_corr=xz_corrections, **options)
```
thierry-martinez: Nit‑picking: I think it's preferable to pass `xz_corr` as a positional argument, since there is no ambiguity here and the name isn't necessarily stable.

tests/baseline/test_causal_flow_draw.png
thierry-martinez, line 1: These images are identical: they do not need to be updated.

graphix/pattern.py
matulni: Done in [213eef7](https://github.com/TeamGraphix/graphix/pull/561/commits/213eef7fde0e97b97a048a3f9c24c9dc6d55987c)

graphix/pattern.py
matulni: Done in [213eef7](https://github.com/TeamGraphix/graphix/pull/561/commits/213eef7fde0e97b97a048a3f9c24c9dc6d55987c)

tests/baseline/test_causal_flow_draw.png
matulni: They were all generated with `pytest --mpl-generate-path=tests/baseline`, is it possible to selectively generate a few of them only ?
I tried to amend the commit to remove the unnecessary updates, but now I'm not sure if I just removed them altogether.

graphix/pattern.py
matulni: Added test in [062d91e](https://github.com/TeamGraphix/graphix/pull/561/commits/062d91ebd4ea39c61e1bde92e35e1888090ba2cb)

tests/baseline/test_causal_flow_draw.png
thierry-martinez: You've removed them indeed in 923119 but you've restored them since, thanks!
We can generate only the images related to a given test by running `pytest --mpl-generate-path=tests/baseline -k test-name`.

thierry-martinez (approved):
LGTM! Thanks.

graphix/pattern.py
```diff
@@ -1542,37 +1542,36 @@ def draw(
         Raises
         ------
         PatternError
-            If the underlying opengraph does not have flow.
+            If the underlying open graph does not have flow.
 
         Notes
         -----
-        If ``flow_from_pattern==True`` but the pattern is not compatible with a gflow, an attempt to be extract the flow from the underlying open graph will be made while warning the user.
+        If ``flow_from_pattern==True`` but the pattern is not compatible with a causal flow, gflow, or Pauli flow, an attempt to be extract the flow from the underlying open graph will be made while warning the user.
```
pranav97nair: Typo:
```suggestion
        If ``flow_from_pattern==True`` but the pattern is not compatible with a causal flow, gflow, or Pauli flow, an attempt to extract the flow from the underlying open graph will be made while warning the user.
```

pranav97nair (approved):
LGTM, thanks!

PR #562, opened 2026-07-13
Fix #553: Handle visualization of empty graphs

### LINKED ISSUE #553: [Bug]: Handle visualization of empty graph properly
Currently, visualizing an empty graph (or an empty pattern) raises an exception.
```python
>>> OpenGraph(nx.Graph(), (), (), {}).draw()
[...]
  File "/home/tmartine/miniforge3/lib/python3.13/site-packages/graphix/visualization.py", line 710, in _
    l_max = max(layers.values())
ValueError: max() iterable argument is empty

>>> Pattern().draw()
[...]
  File "/home/tmartine/miniforge3/lib/python3.13/site-packages/numpy/linalg/_linalg.py", line 144, in _raise_linalgerror_singular
    raise LinAlgError("Singular matrix")
numpy.linalg.LinAlgError: Singular matrix
```

In #549, the situation will be handled more gracefully by raising a `ValueError` with a dedicated message. However, it would be preferable to allow visualization of empty graphs without any failures.

matulni:
Visualization methods yield an empty canvas instead of raising a `ValueError` when called on an empty graph.

### DIFF
--- a/graphix/visualization.py
+++ b/graphix/visualization.py
@@ -216,14 +216,6 @@ class GraphVisualizer:
         GraphVisualizer
         """
         options = VisualizationOptions(**kwargs)
-        if not og.graph.nodes:
-            return GraphVisualizer(
-                og=og,
-                pos={},
-                edge_paths={},
-                options=options,
-                _source=_Source.OG,
-            )
         pos = _compute_positions(og)
         pos = _scale_positions(pos, options.node_distance)
         edge_paths = _compute_edge_paths(og, pos)
@@ -254,14 +246,6 @@ class GraphVisualizer:
         GraphVisualizer
         """
         options = VisualizationOptions(**kwargs)
-        if not flow.og.graph.nodes:
-            return GraphVisualizer(
-                og=flow.og,
-                pos={},
-                edge_paths={},
-                options=options,
-                _source=_Source.Flow,
-            )
         pos = _compute_positions(flow)
         pos = _scale_positions(pos, options.node_distance)
         edge_paths = _compute_edge_paths(flow.og, pos)
@@ -297,14 +281,6 @@ class GraphVisualizer:
         GraphVisualizer
         """
         options = VisualizationOptions(**kwargs)
-        if not xz_corr.og.graph.nodes:
-            return GraphVisualizer(
-                og=xz_corr.og,
-                pos={},
-                edge_paths={},
-                options=options,
-                _source=_Source.XZCorr,
-            )
         pos = _compute_positions(xz_corr)
         pos = _scale_positions(pos, options.node_distance)
         edge_paths = _compute_edge_paths(xz_corr.og, pos)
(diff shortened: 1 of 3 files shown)


graphix/visualization.py
```diff
@@ -215,9 +215,15 @@ def from_opengraph(
...
-            raise ValueError("Open graph is empty.")
         options = VisualizationOptions(**kwargs)
+        if not og.graph.nodes:
+            return GraphVisualizer(
+                og=og,
+                pos={},
+                edge_paths={},
+                options=options,
+                _source=_Source.OG,
+            )
```
thierry-martinez: I would add the following lines at the beginning of `_compute_positions` for `OpenGraph`:
```python
    if not obj.graph.nodes:
        return {}
```
The idea is that we don't need a special execution path here and to re‑specify all of the arguments for `GraphVisualizer`.

graphix/visualization.py
```diff
@@ -247,9 +253,15 @@ def from_flow(
...
-            raise ValueError("Open graph is empty.")
         options = VisualizationOptions(**kwargs)
+        if not flow.og.graph.nodes:
+            return GraphVisualizer(
+                og=flow.og,
+                pos={},
+                edge_paths={},
+                options=options,
+                _source=_Source.Flow,
+            )
```
thierry-martinez: Do we need this? I think the default behavior should be sufficient.

graphix/visualization.py
```diff
@@ -284,9 +296,15 @@ def from_xzcorrections(
...
-            raise ValueError("Open graph is empty.")
         options = VisualizationOptions(**kwargs)
+        if not xz_corr.og.graph.nodes:
+            return GraphVisualizer(
+                og=xz_corr.og,
+                pos={},
+                edge_paths={},
+                options=options,
+                _source=_Source.XZCorr,
+            )
```
thierry-martinez: Do we need this? I think the default behavior should be sufficient.

graphix/visualization.py
matulni: You are right, much cleaner this way. Done in [9daa53a](https://github.com/TeamGraphix/graphix/pull/562/commits/9daa53ab26d376888c54e8b9e1dd330bccce933e)

graphix/visualization.py
matulni: No, indeed, removed in [9daa53a](https://github.com/TeamGraphix/graphix/pull/562/commits/9daa53ab26d376888c54e8b9e1dd330bccce933e)

graphix/visualization.py
matulni: Removed in [9daa53a](https://github.com/TeamGraphix/graphix/pull/562/commits/9daa53ab26d376888c54e8b9e1dd330bccce933e)

thierry-martinez (approved):
LGTM! Thanks.

mgarnier59 (approved):
Nothing to add! Thanks.

PR #563, opened 2026-07-13
Fix #556: unify default output format

### LINKED ISSUE #556: Fix a default output format for `__str__` methods.
It may be useful to have be able to fix a context-dependent output format for pretty print methods with having to import `OutputFormat`

thierry-martinez:
This commit unifies the default behavior of display routines by introducing the context variable `default_output_format`, which is set to `OutputFormat.ASCII` by default.

The `__str__` methods of `PauliFlow`, `XZCorrections`, `Pattern`, and `StateVec` now directly call the corresponding `*_to_str` functions using the default value for the `output` parameter. This default is resolved by `ensure_output_format`, which retrieves the value of `default_output_format`. Previously, `StateVec` produced output without relying on the pretty-printing functions, while `StateVec.draw` uses `OutputFormat.Unicode` by default and the other classes called `to_ascii` directly.

In `docs/source/conf.py`, the following entry has been added to `intersphinx_mapping`:

```python
    "python": ("https://docs.python.org/3", None),
```

This enables links to the Python documentation for `ContextVar`.

The documentation of the `pretty_print` module has been changed to:

```
.. automodule:: graphix.pretty_print
    :members:
```

so that all public members are included in the documentation.


matulni (approved):
Thanks, very elegant. 
Just one question, in general we don't need tot import `contextvars`, right? We could simply do:

```python
from graphix import Circuit, OutputFormat
from graphix.pretty_print import default_output_format

circuit = Circuit(1)
circuit.h(0)

default_output_format.set(OutputFormat.LaTeX)
print(circuit.transpile().pattern)
```

pranav97nair (approved):
This a very nice addition, thanks Thierry!

PR #564, opened 2026-07-13
Fix #474: `FocusedPauliFlow` and `FocusedGFlow` classes

thierry-martinez:
This commit adds `FocusedPauliFlow` and `FocusedGFlow` classes to encapsulate operations specific to focused flows. Correction matrices produce focused flows by construction. Having dedicated classes removes the need to dynamically recheck that a flow is focused during circuit extraction.

This change removes the override of the class method `CausalFlow.try_from_correction_matrix` (that raised an exception), since `CausalFlow` is outside the `FocusedPauliFlow` hierarchy.

### DIFF
--- a/graphix/flow/core.py
+++ b/graphix/flow/core.py
@@ -594,35 +594,6 @@ class PauliFlow(Generic[_AM_co]):
 
     _CF_PREFIX: str = "p"  # Correction function prefix for printing
 
-    @classmethod
-    def try_from_correction_matrix(cls, correction_matrix: CorrectionMatrix[_AM_co]) -> Self | None:
-        """Initialize a ``PauliFlow`` object from a matrix encoding a correction function.
-
-        Parameters
-        ----------
-        correction_matrix : CorrectionMatrix[_AM_co]
-            Algebraic representation of the correction function.
-
-        Returns
-        -------
-        Self | None
-            A Pauli flow if it exists, ``None`` otherwise.
-
-        Notes
-        -----
-        This method verifies if there exists a partial measurement order on the input open graph compatible with the input correction matrix. See Lemma 3.12, and Theorem 3.1 in Ref. [1]. Failure to find a partial order implies the non-existence of a Pauli flow if the correction matrix was calculated by means of Algorithms 2 and 3 in [1].
-
-        References
-        ----------
-        [1] Mitosek and Backens, 2024 (arXiv:2410.23439).
-        """
-        correction_function = correction_matrix.to_correction_function()
-        partial_order_layers = compute_partial_order_layers(correction_matrix)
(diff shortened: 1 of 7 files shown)


graphix/flow/core.py
```diff
@@ -935,6 +906,80 @@ def is_focused(self) -> bool:
                     return False
         return True
 
+    def to_focused(self) -> FocusedPauliFlow[_AM_co]:
+        """Return the Pauli flow as a focused Pauli flow after verifying that it is focused.
+
+        Notes
+        -----
+        If ``self`` is already an instance of :class:`FocusedPauliFlow`,
+        the check is bypassed and ``self`` is returned.
```
mgarnier59: I don't see where this happens, it looks like the test is done, no?

graphix/flow/core.py
```diff
@@ -935,6 +906,80 @@ def is_focused(self) -> bool:
...
+    """An immutable dataclass representing a focused Pauli flow.
+
+    An instance of ``FocusedPauliFlow`` is a :class:`PauliFlow` that
+    satisfies the :meth:`PauliFlow.is_focused` predicate. Once a
+    ``FocusedPauliFlow`` instance has been obtained, this property
+    does not need to be checked again.
+    """
+
+    @classmethod
+    def try_from_correction_matrix(cls, correction_matrix: CorrectionMatrix[_AM_co]) -> Self | None:
```
mgarnier59: Just to be sure: this is supposed to be the implementation that `FocusedGFlow` will use (via `super`) since `GFlow` doesn't implement it? In this way we guarantee that everything that comes out of the finding algorithm is of a `Focused` type?

graphix/flow/core.py
```diff
@@ -935,6 +906,80 @@ def is_focused(self) -> bool:
...
+        return FocusedPauliFlow(self.og, self.correction_function, self.partial_order_layers)
+
+
+class FocusedPauliFlow(PauliFlow[_AM_co], Generic[_AM_co]):
+    """An immutable dataclass representing a focused Pauli flow.
+
+    An instance of ``FocusedPauliFlow`` is a :class:`PauliFlow` that
+    satisfies the :meth:`PauliFlow.is_focused` predicate. Once a
+    ``FocusedPauliFlow`` instance has been obtained, this property
+    does not need to be checked again.
```
mgarnier59: I don't see where this is enforced? There is no `is_focused` for `Focused`classes

mgarnier59 (commented):
No comments on the nice work. I just want to be sure I understand what has been done. Let's iterate in a meeting or via comments!

Thanks!

mgarnier59 (approved):

Issue #493, opened 2026-04-29
API for pattern optimisations including pre-simulation should be streamlined.


emlynsg:
It is good both to be explicit with optimisations and to provide a simple API to allow regular users to run a common routine.

For future consideration and discussion:
- Add `transpile_swaps` field to `Circuit.transpile()` to provide easy access to the `transpile_swaps` method
- Add `optimize` method to `Pattern` which runs a chosen regular sequence of commands (likely some combination of `standardize` and  `minimize_space`)
- Add `presimulate_paulis` method to `Pattern` which runs the sequence of `remove_input_nodes`, `infer_pauli_measurements` and `perform_pauli_measurements`
- Add a method to `Pattern` which applies in-place the `infer_pauli_measurements` method (to avoid the requirement to write `pattern = pattern.infer_pauli_measurements()`.

PR #567, opened 2026-07-18
Rename `simulate_pattern` and `simulate_statevector` into `simulate`

thierry-martinez:
This PR renames the methods `Pattern.simulate_pattern` and `Circuit.simulate_statevector` to `simulate`.

The new names are more appropriate than the previous ones:

* `Pattern.simulate_pattern` unnecessarily repeated the type in the method name.
* `Circuit.simulate_statevector` has been inaccurate since #512 because the circuit simulator now supports density matrix simulation as well.

Moreover, using the same `simulate` method name for both `Pattern` and `Circuit` makes the API more consistent.

This is a step toward a cleaner simulation API, as proposed in #493.

This PR also updates the documentation. It replaces the explicit method listings for `Circuit` and `Pattern` with `:members:`, allowing more methods to be included automatically in the API reference. Along the way, a few docstrings have also been fixed.


matulni (commented):
Thanks a lot for this work, it improves the codebase!
Two comments:
- While we are at it, do you think it would be possible to rename `graphix.transpiler.SimulateResult.statevec` to `graphix.transpiler.SimulateResult.result` or `graphix.transpiler.SimulateResult.state` (it's more accurate, since it sometimes contains a density matrix object) ?
- Not from this PR, but I believe that the function `graphix.transpiler._check_target` is not used anywhere. This PR may be a good place to remove this dead code.

pranav97nair (approved):
LGTM, thanks for this update Thierry!

thierry-martinez:
Thank you, @matulni and @pranav97nair, for your reviews. @matulni, I've addressed your comments.

matulni (approved):

PR #568, opened 2026-07-18
Add in-place variants for `infer_pauli_measurements`, `to_bloch`,...

thierry-martinez:
Method `Pattern.infer_pauli_measurements` now operates in place by default, with an optional `copy` parameter.  If `copy=True`, a modified copy is returned instead.

Added the following methods to `Pattern`. By default, they modify the pattern in place, with an optional `copy` parameter.  If `copy=True`, a modified copy is returned instead.
  - `apply`, an in-place variant of `map`,
  - `blochify`, an in-place variant of `to_bloch`,
  - `substitute`, an in-place variant of `subs`,
  - `replace_parameters`, an in-place variant of `xreplace`.

This is a step toward the cleaner simulation API proposed in #493. It also makes the examples in the Graphix paper cleaner.


matulni (commented):
Thanks for the change, new functions are useful!

I'm not convinced by the naming of the parametric angle functions. The following are not very intuitive to me:

- ``subs`` and ``substitute`` have the same relationship as ``xreplace`` and ``replace_parameters``.
- ``substitute`` and ``replace_parameters`` act in-place.
- ``subs`` replaces one parameter, and ``xreplace`` replaces multiple parameters in parallel. (I'm aware that this is a sympy convention and that it's independent of this PR, but I always found it difficult to remember. In addition, with the new methods we drift away from sympy's convention).

Here are two alternatives:

A)
`substitute` -> `subs_in_place`
`replace_parameters` -> `xreplace_in_place`

It solves 1) and 2), but it's more verbose and not true when calling `subs_in_place(..., copy=True)`.

B)
`subs` -> `substituted`

`substitute` remains

`xreplace` -> `substituted_multi`

`replace_parameters` -> `substitute_multi`

I think this solves all three points but we don't follow sympy convention at all, and it requires a larger refactoring (we would need to update analogous methods in `OpenGraph`, etc.). If you think the -ed suffix may be easier to miss when reading the code, we could use `with_substitution` instead.

What do you think ?

pranav97nair (approved):
Thanks for this change. The new functions look good to me.

thierry-martinez:
Thank you, @matulni and @pranav97nair, for your reviews.

I consolidate the API changes:
  - All classes that have `subs` and `xreplace` methods (`Pattern`, `Circuit`, `OpenGraph`, `Measurement`, `PauliFlow`, `XZCorrections`, `DensityMatrix`) now derive from the `Parameterizable` class, which exposes the method `with_parameter` and `with_parameters`, with `subs` and `xreplace` as their respective aliases.
  - `Pattern` and `Circuit` now derive from the `InplaceParameterizable` subclass, which additionally provides the following methods:
    - `replace_parameter`, an in-place variant of `with_parameter` (or `subs`),
    - `replace_parameters`, an in-place variant of `with_parameters` (or `xreplace`).

Moreover, I added a class `AbstractStatevector` to allow `Statevector` defined in `graphix-symbolic` to be pretty-printed like the built-in `Statevector`.

graphix/parameter.py
```diff
@@ -348,21 +371,23 @@ def check_expression_or_float(value: object) -> ExpressionOrFloat:
...
+def with_parameter(value: _T, variable: Parameter, substitute: ExpressionOrSupportsFloat) -> _T | Expression | complex:
     """
     Substitute in `value`.
 
     If `value` is in instance of :class:`Expression`, then return
-    `value.subs(variable, substitute)` (coerced into a complex or a
+    `value.with_parameter(variable, substitute)` (coerced into a complex or a
     float if the result is a number).
 
     If `value` does not implement `subs`, `value` is returned
```
matulni:
```suggestion
    If ``value`` does not implement ``with_parameter``, ``value`` is returned
```

graphix/parameter.py
```diff
@@ -384,23 +409,25 @@ def subs(value: T, variable: Parameter, substitute: ExpressionOrSupportsFloat) -
...
+def with_parameters(value: _T, assignment: Mapping[Parameter, ExpressionOrSupportsFloat]) -> _T | Expression | complex:
     """
     Substitute in parallel in `value`.
 
     If `value` is an an instance of :class:`Expression`, then return
-    `value.xreplace(assignment)` (coerced into a complex if the result
+    `value.with_parameters(assignment)` (coerced into a complex if the result
     is a number).
 
     If `value` does not implement `xreplace`, `value` is returned
```
matulni:
```suggestion
    If ``value``` does not implement ``with_parameters``, ``value`` is returned
```

graphix/flow/core.py
```diff
@@ -537,10 +541,11 @@ def subs(self: XZCorrections[_M], variable: Parameter, substitute: ExpressionOrS
         -----
         See notes and examples in :func:`OpenGraph.subs`.
```
matulni:
```suggestion
        See notes and examples in :meth:`OpenGraph.with_parameter`.
```

graphix/flow/core.py
```diff
@@ -559,12 +564,12 @@ def xreplace(
         -----
         See notes and examples in :func:`OpenGraph.xreplace`.
```
matulni:
```suggestion
        See notes and examples in :meth:`OpenGraph.with_parameters`.
```

graphix/flow/core.py
```diff
@@ -879,10 +885,11 @@ def subs(  # noqa: PYI019 Annotating with ``Self`` is not possible since ``self`
         -----
         See notes and examples in :func:`OpenGraph.subs`.
```
matulni:
```suggestion
        See notes and examples in :meth:`OpenGraph.with_parameter`.
```

graphix/flow/core.py
```diff
@@ -901,7 +908,7 @@ def xreplace(  # noqa: PYI019
         -----
         See notes and examples in :func:`OpenGraph.xreplace`.
```
matulni:
```suggestion
        See notes and examples in :meth:`OpenGraph.with_parameters`.
```

graphix/parameter.py
```diff
@@ -348,21 +371,23 @@ def check_expression_or_float(value: object) -> ExpressionOrFloat:
...
-# The return type could be `T | complex` since `subs` returns `Expression` only
-# if `T == Expression`, but `mypy` does not handle this yet: https://github.com/python/mypy/issues/12989
-def subs(value: T, variable: Parameter, substitute: ExpressionOrSupportsFloat) -> T | Expression | complex:
+# The return type could be `_T | complex` since `with_parameter` returns `Expression` only
+# if `_T == Expression`, but `mypy` does not handle this yet: https://github.com/python/mypy/issues/12989
+def with_parameter(value: _T, variable: Parameter, substitute: ExpressionOrSupportsFloat) -> _T | Expression | complex:
     """
     Substitute in `value`.
 
     If `value` is in instance of :class:`Expression`, then return
```
pranav97nair:
```suggestion
    If `value` is an instance of :class:`Expression`, then return
```

graphix/parameter.py
```diff
@@ -384,23 +409,25 @@ def subs(value: T, variable: Parameter, substitute: ExpressionOrSupportsFloat) -
...
-# The return type could be `T | Expression | complex` since `subs` returns `Expression` only
-# if `T == Expression`, but `mypy` does not handle this yet: https://github.com/python/mypy/issues/12989
-def xreplace(value: T, assignment: Mapping[Parameter, ExpressionOrSupportsFloat]) -> T | Expression | complex:
+# The return type could be `_T | Expression | complex` since `subs` returns `Expression` only
+# if `_T == Expression`, but `mypy` does not handle this yet: https://github.com/python/mypy/issues/12989
+def with_parameters(value: _T, assignment: Mapping[Parameter, ExpressionOrSupportsFloat]) -> _T | Expression | complex:
     """
     Substitute in parallel in `value`.
 
     If `value` is an an instance of :class:`Expression`, then return
```
pranav97nair:
```suggestion
    If `value` is an instance of :class:`Expression`, then return
```

graphix/sim/statevec.py
```diff
@@ -48,7 +53,231 @@
...
+        other : :class:`graphix.sim.statevec.Statevector`
+            Statevector to compare with.
+
+        Returns
+        -------
+        float
+            Fidelity between the two statevectors.
+        """
+        inner = np.dot(self.flatten().conjugate(), other.flatten())
+        return float(np.abs(inner) ** 2)
```
matulni: This method will not work on a symbolic state vector. For instance, the test

```python
def test_fidelity() -> None:
    alpha = SympyParameter("alpha")
    circuit = Circuit(1)
    circuit.rz(0, alpha)
    result_sym = circuit.simulate_statevector(
        backend=StatevectorBackend(branch_selector=RandomBranchSelector(pr_calc=False), symbolic=True)
    ).statevec

    assert result_sym.fidelity(result_sym) == pytest.approx(1)
```

will raise the following error:

```
============================= test session starts ==============================
platform darwin -- Python 3.13.14, pytest-9.1.1, pluggy-1.6.0
benchmark: 5.2.3 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
Matplotlib: 3.10.9
Freetype: 2.6.1
rootdir: /Users/muldemol/QAT/projects/graphix/graphix-plugins/graphix-symbolic
configfile: pyproject.toml
plugins: benchmark-5.2.3, mock-3.15.1, cov-7.1.0, mpl-0.19.0
collected 5 items / 4 deselected / 1 selected

tests/test_sympy_parameter.py F                                          [100%]

=================================== FAILURES ===================================
________________________________ test_fidelity _________________________________

    def test_fidelity() -> None:
        alpha = SympyParameter("alpha")
        circuit = Circuit(1)
        circuit.rz(0, alpha)
        result_sym = circuit.simulate_statevector(
            backend=StatevectorBackend(branch_selector=RandomBranchSelector(pr_calc=False), symbolic=True)
        ).statevec
    
>       assert result_sym.fidelity(result_sym) == pytest.approx(1)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests/test_sympy_parameter.py:71: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <graphix_symbolic.statevec.Statevector object at 0x141474ec0>
other = <graphix_symbolic.statevec.Statevector object at 0x141474ec0>

    def fidelity(self, other: Statevector) -> float:
        r"""Calculate the fidelity against another statevector.
    
        The fidelity is defined as :math:`|\langle\psi_1|\psi_2\rangle|^2`.
    
        Parameters
        ----------
        other : :class:`graphix.sim.statevec.Statevector`
            Statevector to compare with.
    
        Returns
        -------
        float
            Fidelity between the two statevectors.
        """
        inner = np.dot(self.flatten().conjugate(), other.flatten())
>       return float(np.abs(inner) ** 2)
                     ^^^^^^^^^^^^^
E       TypeError: bad operand type for abs(): 'SympyExpression'

../../graphix/graphix/sim/statevec.py:95: TypeError
=========================== short test summary info ============================
FAILED tests/test_sympy_parameter.py::test_fidelity - TypeError: bad operand ...
======================= 1 failed, 4 deselected in 3.38s ========================
```
tested on your branch `in-place_methods` on your `graphix-symbolic` fork.

This bug existed in the codebase before we removed the symbolic simulation capabilities, I noticed it when migrating the symbolic backend to the `graphix-symbolic` package.
I think we could fix it writing a function to handle the symbolic abs operation, but I really doubt it is worth (I'm starting to question even if it's worth to maintain the symbolic backend at all).

graphix/sim/statevec.py
```diff
@@ -48,7 +53,231 @@
...
+
+    def _to_dict_map(
+        self,
+        f: Callable[[npt.NDArray[_ScalarT]], npt.NDArray[_ScalarT2]],
+        encoding: _ENCODING = "MSB",
+        *,
+        rtol: float = 0.0,
+        atol: float = 1e-8,
+    ) -> dict[str, _ScalarT2]:
+        mask = np.logical_not(np.isclose(np.abs(self.psi), 0, rtol=rtol, atol=atol))
```
matulni: This method won't work on symbolic instances for the same reason that the method `fidelity` fails.
As a result methods `draw`, `to_prob_dict` and `to_dict` also fail.

graphix/sim/statevec.py
```diff
@@ -48,7 +53,231 @@
...
+        """Return flattened state vector.
+
+        A view of only the first ``2**self.nqubit`` elements of ``self._psi`` is returned.
+        """
+        flat_psi = self.psi.flatten()
+        if flat_psi.dtype == np.object_:
+            return flat_psi.astype(np.object_, copy=False)
+        return flat_psi.astype(np.complex128, copy=False)
+
+    def fidelity(self, other: Statevector) -> float:
```
matulni: Why restrict `other` to `Statevector` ?
This would imply that we cannot compute the fidelity between two objects of a custom subclass, but that's the most likely case.

matulni (approved):
Thanks a lot for implementing this changes, it's a lot of work.

I'm happy with all the changes regarding the parametric substitution.

However, I believe that the new shared methods in the `AbstractStatevector` class won't work with the symbolic backend (I give a specific example below). I suspect that the code is not generic enough to work on the GPU backend either.

Making this methods work requires some additional work on the symbolic backend, which I don't think is worth at all. Then, as it is now, I don't think we need the `AbstractStatevector` class. I agree that in the future it could be practical if we implement others backend more useful than the symbolic one, but can see then.

Something I missed: is the `AbstractStatevector` class necessary to make the new work on the parametric substitution work ? Couldn't we leave it for a future PR ?

pranav97nair (approved):
Thanks for these changes, Thierry. The new abstract classes `Parameterizable` and `InplaceParametrizable` are nice additions. I've just found a couple minor typos.

PR #570, opened 2026-07-18
Fix type annotations in `test_flow_core`

thierry-martinez:
This commit removes the `# type: ignore[misc]` pragma from `test_flow_core`. It introduces a new `PatternTestCase` class, derived from `XZCorrectionsTestCase`, to refine the type of the `flow` field when the `pattern` field is present.


matulni (approved):
Thank you, LGTM!

pranav97nair (approved):
LGTM!

PR #571, opened 2026-07-22
Remove method `Pattern.graph`

matulni:
This commit removes the method `Pattern.graph` which is not correct as it does not standardize the pattern. Where appropriate it was replaced by `Pattern.to_opengraph().graph`, including unmerged branches in reverse dependencies. 

Method `Pattern.isolated_nodes` is rewritten to avoid constructing the open graph as it's not necessary.


COMPATIBILITY.md
```diff
@@ -4,6 +4,8 @@ Graphix does not guarantee backwards compatibility between 0.x version releases.
 
 ## Unreleased
 
+- #571: Method `Pattern.graph` (formely `Pattern.extract_graph`) is removed. Use instead `Pattern.to_opengraph().graph` which standardizes the pattern and fails gracefully if there are non-commutative CLiffords.
```
thierry-martinez:
```suggestion
- #571: Method `Pattern.graph` (formely `Pattern.extract_graph`) is removed. Use instead `Pattern.to_opengraph().graph` which standardizes the pattern and fails gracefully if there are non-commutative Cliffords.
```

graphix/pattern.py
```diff
@@ -1237,8 +1216,15 @@ def isolated_nodes(self) -> set[int]:
...
-        graph = self.graph()
-        return {node for node, d in graph.degree if d == 0}
+        nodes = set(self.input_nodes)
+        for cmd in self.__seq:
+            match cmd.kind:
+                case CommandKind.N:
+                    nodes.add(cmd.node)
+                case CommandKind.E:
+                    u, v = cmd.nodes
+                    nodes -= {u, v}
```
thierry-martinez: The following test passes on `master` but fails with this change.
```python
def test_isolated_nodes() -> None:
    pattern = Pattern(input_nodes=[0, 1], cmds=[E((0, 1)), E((0, 1))])
    assert pattern.isolated_nodes() == {0, 1}
```

COMPATIBILITY.md
matulni: Fixed in [248e6df](https://github.com/TeamGraphix/graphix/pull/571/commits/248e6df36397a68c358a8c88fddcc36d5d0f75c4)

graphix/pattern.py
matulni: Thanks. As discussed, I modified `isolated_nodes` to get the graph from the pattern's open graph in [248e6df](https://github.com/TeamGraphix/graphix/pull/571/commits/248e6df36397a68c358a8c88fddcc36d5d0f75c4).

I added your test as well.

thierry-martinez (approved):
LGTM! Thanks.

PR #576, opened 2026-07-31
Fix pretty-printing of Clifford gates

thierry-martinez:
This commit implements `clifford_to_str`, which relies on the QASM3 decomposition (into `I`, `X`, `Y`, `Z`, `S`, `S†`, and `H`), since it is more compact than the HSZ decomposition.

`CLIFFORD_LABEL` is removed from `_db` as names were ambiguous. The QASM3 decomposition has been made optimal and is now tested.

All members of the `Clifford` class now appear in the documentation and docstrings are fixed.


tests/test_db.py
```diff
@@ -131,3 +134,49 @@ def unwrap(v: T | None) -> T:
...
+        for clifford, decomposition in current_decompositions.items():
+            for instr_clifford, instr in QASM3_BASIS.items():
+                result = instr_clifford @ clifford
+                if decompositions[result.value] is not None:
+                    continue
+                result_decomposition = (*decomposition, instr)
+                decompositions[result.value] = result_decomposition
+                new_decompositions[result] = result_decomposition
+    assert all(decomposition is not None for decomposition in decompositions)
+    return tuple(map(unwrap, decompositions))
```
matulni, line 177: Very nice!

matulni (approved):
LGTM, thanks!

tests/test_db.py
thierry-martinez: Seeing the code again, I realized that lines 167 and 174 are making the job twice. Fixed in 21590b2. Thanks for the compliment!

pranav97nair (approved):
Nice work, thanks Thierry!

PR #585, opened 2026-08-13
Fix #583: typing issues with numpy 2.5

### LINKED ISSUE #583: [Bug]: `numpy>=2.5` causes the build to fail
`numpy>=2.5` causes the build to fail:
- https://github.com/TeamGraphix/graphix/actions/runs/31588265715/job/94087035559?pr=579
- https://github.com/TeamGraphix/graphix/actions/runs/31666880725/job/94343296433?pr=581
- https://github.com/TeamGraphix/graphix/actions/runs/31667056049/job/94343799834?pr=582

The current constraint in `pyproject.toml` is `numpy>=2`.

thierry-martinez:
This commit fixes the typing issues introduced by numpy 2.5, making the codebase compatible with newer versions of numpy while preserving compatibility with older versions (from numpy 2).

numpy 2.5 uses the type alias syntax to define `NDArray`, so we can no longer use `NDArray` as a base type.

The issue https://github.com/numpy/numpy/issues/31513 has been fixed, so we can stop ignoring the argument types of `np.einsum`.

`rng.choice` is now typed as `signedinteger[_32Bit | _64Bit]` instead of `Any`, so `rng.choice([0, 1])` is no longer a valid `Outcome` value. This commit introduces `rand_outcome` to generate a random `Outcome` value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `replace_parameter`, an in-place variant of `with_parameter` (or `subs`),
     - `replace_parameters`, an in-place variant of `with_parameters` (or `xreplace`).
 
+- #585: Added `rand_outcome`.
+
 ### Fixed
 
 - #454, #481: Ensure `Pattern.minimize_space` only reduces max-space and does not increase it.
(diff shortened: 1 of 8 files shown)


pranav97nair (approved):
Thanks for this update! Looks good to me.

matulni (approved):
Thanks, LGTM!

PR #586, opened 2026-08-14
Fix #584: swap transpilation of measured qubits

### LINKED ISSUE #584: [Bug]: Swap transpilation fails if qubit is swapped and measured.
**Edit** : Simplified test

The following snippet fails with an expected error on master branch:

```python
from graphix import Circuit, Axis

circuit = Circuit(2)
circuit.swap(0, 1)
circuit.m(1, Axis.Z)

circuit.transpile().pattern  # Raises KeyError: 1
```

Transpilation is successful when disabling swaps transpilation:

```python
import numpy as np
rng = np.random.default_rng(42)

pattern = circuit.transpile(transpile_swaps=False).pattern

state_qc = circuit.simulate(rng=rng).state
state_mbqc = pattern.simulate(rng=rng)
assert state_mbqc.isclose(state_qc)
```

matulni:
This commit proposes a rewrite of the method `graphix.transpiler.TranspileSwapsResult.extract_output_node_indices` to fix #584.

New tests have been added accordingly.

This commit also removed some duplicate tests from `test_transpiler.py`.


graphix/transpiler.py
```diff
@@ -1008,14 +1008,22 @@ def extract_outputs(self, kind: OutputKind) -> tuple[int, ...]:
...
+
+        bit_index = 0
+        qubit_indices_shifted = {}
+        for index in range(self.circuit.width):
+            if index not in qubit_indices_set:
+                bit_index += 1
+            else:
+                qubit_indices_shifted[index] = index - bit_index
+
+        return tuple(qubit_indices_shifted[qubit_index] for qubit_index in qubit_indices)
```
thierry-martinez: A bit simpler:

```suggestion
        qubit_indices = self.extract_outputs(OutputKind.Qubit)
        rank = {q: i for i, q in enumerate(sorted(qubit_indices))}
        return tuple(rank[q] for q in qubit_indices)
```

thierry-martinez (approved):
Thanks for fixing this and for removing the duplicated tests.

graphix/transpiler.py
matulni: Thanks, applied in [1311431](https://github.com/TeamGraphix/graphix/pull/586/commits/13114319066eabd3054bd653b461a9a5fae3b832)

pranav97nair (approved):
Nice catch and thanks for the fix, Mateo! I don't see any issues.

PR #591, opened 2026-08-25
Ensure RNG persistence during hybrid branch selection

pranav97nair:
This PR fixes a bug where simulating a pattern using a `FixedBranchSelector` on a partial `results` mapping and a `RandomBranchSelector` for the remaining qubits always uses an un-seeded default RNG, even if a seeded one is passed to the  `PatternSimulator`. Consider the following example:
```python
from numpy.random import default_rng
from graphix import Pattern, Measurement, RandomBranchSelector, FixedBranchSelector
from graphix.command import N, M
from graphix.measurements import Outcome

pattern = Pattern(cmds=[N(0), N(1), M(0), M(1, Measurement.XY(0.5))])

results: dict[int, Outcome] = {0: 0}
rbs = RandomBranchSelector()
fbs = FixedBranchSelector(results, default=rbs)
rng_seeded = default_rng(seed=42)

pattern.simulate(branch_selector=fbs, rng=rng_seeded)
```
The above code raises `UserWarning: Default random-number generator is used. Results may not be reproducible.` on `graphix/branch_selector.py:137` where the `FixedBranchSelector` calls 
```python
self.default.measure(qubit, f_expectation0)
```
without passing its `rng` parameter to the `default` branch selector. With this update, we now include the `rng` argument in the function call on this line which ensures that the RNG object provided to the `PatternSimulator` reaches the `RandomBranchSelector`.

In addition, a new test was added to verify this behaviour.

### DIFF
--- a/graphix/branch_selector.py
+++ b/graphix/branch_selector.py
@@ -134,7 +134,7 @@ class FixedBranchSelector(BranchSelector, Generic[_T]):
         if result is None:
             if self.default is None:
                 raise ValueError(f"Unexpected measurement of qubit {qubit}.")
-            return self.default.measure(qubit, f_expectation0, rng)
+            return self.default.measure(qubit, f_expectation0, rng, stacklevel=stacklevel + 1)
         return result
 
 
--- a/graphix/branch_selector.py
+++ b/graphix/branch_selector.py
@@ -134,7 +134,7 @@ class FixedBranchSelector(BranchSelector, Generic[_T]):
         if result is None:
             if self.default is None:
                 raise ValueError(f"Unexpected measurement of qubit {qubit}.")
-            return self.default.measure(qubit, f_expectation0)
+            return self.default.measure(qubit, f_expectation0, rng)
         return result
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - #586: Fixed #584 Swap transpilation of measured qubits.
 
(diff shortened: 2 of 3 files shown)


thierry-martinez (approved):
Nice catch and perfect fix! Could you update the changelog?

pranav97nair:
> Nice catch and perfect fix! Could you update the changelog?

Thanks for the review! I've included the change under version 0.4 as it is still in pre-release and I believe we can still update it before making it final. Let me know if you agree with this approach, or if I should put this under 'Unreleased' instead.

thierry-martinez:
Graphix 0.4 is still in pre-release but is already on PyPI: I would avoid making changes to it except for blocking bugs (bugs that would make Graphix really unusable or insecure). For this particular bug, I think we can wait for Graphix 0.4.1.

pranav97nair:
Okay makes sense. I've logged the change under `Unreleased`.

graphix/branch_selector.py
```diff
@@ -134,7 +134,7 @@ def measure(
         if result is None:
             if self.default is None:
                 raise ValueError(f"Unexpected measurement of qubit {qubit}.")
-            return self.default.measure(qubit, f_expectation0)
+            return self.default.measure(qubit, f_expectation0, rng)
```
thierry-martinez: It would be nice if the warnings, if any, were reported to the call site.
```suggestion
            return self.default.measure(qubit, f_expectation0, rng, stacklevel=stacklevel + 1)
```

pranav97nair: Done!

matulni (approved):
LGTM, thanks for the fix!

PR #592, opened 2026-08-25
Add all standard OpenQASM gates

thierry-martinez:
This commit adds the following gates specified in the OpenQASM 3.0 specification: SDG, T, TDG, SX, SXDG, CY, P, U, CP, CRX, CRY, CRZ, CU, CSWAP, and GPHASE.

Support for GPHASE is currently limited: the gate is ignored by the circuit simulator and the transpiler. It is useful, however, to have it internally in order to add control to the decompositions of the other gates.

Additionally, this commit adds the CJ gate, which is useful internally for adding control to decompositions that involve J gates.

These additions were motivated by the issue TeamGraphix/graphix-qasm-parser#14. The OpenQASM parser plugin can now parse all the gates and map them to native Graphix gates without translation, except for `u1`, `u2`, and `u3`, which are maintained for compatibility with OpenQASM 2.0 and are translated.


graphix/instruction.py
```diff
@@ -211,6 +242,51 @@ def visit(self, visitor: InstructionVisitor, *, copy: bool = False) -> SWAP:
...
+    The CSWAP gate applies the matrix
+
+    .. math::
+
+      \left[\begin{matrix}
+        1 & 0 & 0 & 0\\
+        0 & 1 & 0 & 0\\
+        0 & 0 & \cos \frac \theta 2 & -\mathrm i \sin \frac \theta 2\\
+        0 & 0 & -\mathrm i \sin \frac \theta 2 & \cos \frac \theta 2
+      \end{matrix}\right]
```
matulni: Is this matrix a typo ?

graphix/ops.py
```diff
@@ -167,6 +228,187 @@ def rz(theta: ParameterizedAngle) -> npt.NDArray[np.complex128] | npt.NDArray[np
...
+                [0, 1, 0, 0],
+                [0, 0, exp(1j * gamma_rad) * cos, -exp(1j * (gamma_rad + lambda_rad)) * sin],
+                [0, 0, exp(1j * (gamma_rad + phi_rad)) * sin, exp(1j * (gamma_rad + phi_rad + lambda_rad)) * cos],
+            ],
+            theta,
+        )
+
+    CH: ClassVar[npt.NDArray[np.complex128]] = controlled(H)
+
+    CSWAP: ClassVar[npt.NDArray[np.complex128]] = controlled(SWAP)
```
matulni: Nit: could we group all `ClassVar` together ? I think it would slightly improve readeability

graphix/ops.py
```diff
@@ -167,6 +228,187 @@ def rz(theta: ParameterizedAngle) -> npt.NDArray[np.complex128] | npt.NDArray[np
...
+        Returns
+        -------
+        operator : 2*2 np.asarray
+        """
+        cos, sin = cos_sin(angle_to_rad(theta) / 2)
+        phi_rad = angle_to_rad(phi)
+        lambda_rad = angle_to_rad(lambda_)
+        return Ops._cast_array(
+            [[cos, -exp(1j * lambda_rad) * sin], [exp(1j * phi_rad) * sin, exp(1j * (phi_rad + lambda_rad)) * cos]],
+            theta,
```
matulni: Why using only `theta` to cast the array ?  In principle  `theta` could be a `float` and `phi` or `lambda` still be expressions, so the matrix should be casted to `np.object_`, right?

graphix/ops.py
```diff
@@ -167,6 +228,187 @@ def rz(theta: ParameterizedAngle) -> npt.NDArray[np.complex128] | npt.NDArray[np
...
+        gamma_rad = angle_to_rad(gamma)
+        return Ops._cast_array(
+            [
+                [1, 0, 0, 0],
+                [0, 1, 0, 0],
+                [0, 0, exp(1j * gamma_rad) * cos, -exp(1j * (gamma_rad + lambda_rad)) * sin],
+                [0, 0, exp(1j * (gamma_rad + phi_rad)) * sin, exp(1j * (gamma_rad + phi_rad + lambda_rad)) * cos],
+            ],
+            theta,
+        )
```
matulni:
```suggestion
        gamma_rad = angle_to_rad(gamma)
        return controlled(exp(1j * gamma_rad) * Ops.u(theta, phi, lambda_))
```

graphix/transpiler.py
```diff
@@ -628,17 +983,36 @@ def evolve(op: Matrix, qargs: Iterable[int]) -> None:
...
                     evolve(Ops.CCX, [instr.controls[0], instr.controls[1], instr.target])
+                case instruction.InstructionKind.CSWAP:
+                    evolve(Ops.CSWAP, [instr.control, instr.targets[0], instr.targets[1]])
                 case instruction.InstructionKind.M:
                     result = _backend.measure(
                         instr.target, PauliMeasurement(instr.axis), rng=rng, stacklevel=stacklevel + 1
                     )
                     classical_measures.append(result)
+                case InstructionKind.GPHASE:
+                    gphase += instr.angle
```
matulni: What is the variable `gphase` for ?

graphix/instruction.py
```diff
@@ -159,24 +175,39 @@ def visit(self, visitor: InstructionVisitor, *, copy: bool = False) -> RZZ:
...
+
+
+@dataclass(repr=False)
+class CNOT(_KindChecker, ControlledSingleTargetInstruction):
+    """CNOT circuit instruction."""
+
+    kind: ClassVar[Literal[InstructionKind.CNOT]] = field(default=InstructionKind.CNOT, init=False)
+
+
+# CZ is not defined as a ControlledSingleTargetInstruction because of
```
matulni, line 209: Probably this PR is the place to change this, but I think this convention is very non-intuitive:
- `tuple` is an ordered sequence, so it does not truly reflect the symmetry.
- Methods of `Circuit` don't follow this convention. Both `Circuit.cz` and `Circuit.cnot` take two integers (with different parameter names, which is even worse).

Since this PR is already quite big, I suggest we leave it as is, and eventually discuss it further in an issue.

graphix/instruction.py
```diff
@@ -55,6 +55,22 @@ class InstructionKind(Enum):
     RX = enum.auto()
     RY = enum.auto()
     RZ = enum.auto()
+    SDG = enum.auto()
```
matulni, line 58: Now that we have an $S^\dagger$ gate, we could replace $S Z$ by $S^\dagger$ in `graphix.circ_ext.compilation.cm_berg_pass` so that circuit extraction returns circuits with less number of gates.

graphix/transpiler.py
```diff
@@ -921,9 +1320,118 @@ def decompose_rz(instr: instruction.RZ) -> Iterator[instruction.J]:
...
+        Instruction.GPHASE | Instruction.X | Instruction.Z | Instruction.J | Instruction.CNOT | Instruction.RZ
+    ],
+) -> Iterable[InstructionType]:
+    """Yield a controlled gate sequence from a gate sequence.
+
+    Parameters
+    ----------
+    control: int
+        The control qubit.
+    instrs: Iterable[Instruction.X | Instruction.Z | Instruction.J | Instruction.CNOT | Instruction.RZ]
```
matulni:
```suggestion
    instrs: Iterable[Instruction.GPHASE | Instruction.X | Instruction.Z | Instruction.J | Instruction.CNOT | Instruction.RZ]
```

graphix/transpiler.py
```diff
@@ -921,9 +1320,118 @@ def decompose_rz(instr: instruction.RZ) -> Iterator[instruction.J]:
...
+            case InstructionKind.Z:
+                yield instruction.CZ((control, instr.target))
+            case InstructionKind.J:
+                yield instruction.CJ(control=control, target=instr.target, angle=instr.angle)
+            case InstructionKind.CNOT:
+                yield instruction.CCX(target=instr.target, controls=(control, instr.control))
+            case InstructionKind.RZ:
+                yield instruction.CRZ(control=control, target=instr.target, angle=instr.angle)
+            case InstructionKind.GPHASE:
+                gphase += instr.angle
```
matulni: I would prefer `yield Instruction.P(target=control, angle=gphase)` at this point instead of doing it at the end for the following reasons:
- The number of gates is conserved.
- The resulting sequence of gates can be easily mapped to the original sequence of gates (one to one mapping, without permutations).
- We don't introduce unnecessary `Instruction.P` with `angle=0` unless there is such gate in the input sequence.

If needed we could introduce a function `merge_gphase_gates` to merge all `gphase` gates into a single one.

graphix/transpiler.py
```diff
@@ -921,9 +1320,118 @@ def decompose_rz(instr: instruction.RZ) -> Iterator[instruction.J]:
...
+            case _:
+                assert_never(instr.kind)
+    yield Instruction.P(target=control, angle=gphase)
+
+
+def decompose_cj(instr: Instruction.CJ) -> Iterator[InstructionType]:
```
matulni: Can't we narrow the return type?

```suggestion
def decompose_cj(instr: Instruction.CJ) -> Iterator[Instruction.RZ | Instruction.CNOT | Instruction.RY | Instruction.P]:
```

graphix/transpiler.py
```diff
@@ -921,9 +1320,118 @@ def decompose_rz(instr: instruction.RZ) -> Iterator[instruction.J]:
...
+    yield Instruction.CJ(control=instr.control, target=instr.target, angle=0)
+    yield Instruction.P(target=instr.control, angle=instr.gamma - instr.theta / 2)
+
+
+def insert_control(
+    control: int,
+    instrs: Iterable[
+        Instruction.GPHASE | Instruction.X | Instruction.Z | Instruction.J | Instruction.CNOT | Instruction.RZ
+    ],
+) -> Iterable[InstructionType]:
```
matulni: Here too we can narrow the return type:

```suggestion
) -> Iterator[Instruction.CNOT | Instruction.CZ | Instruction.CJ | Instruction.CCX | Instruction.CRZ | Instruction.P]:
```

graphix/transpiler.py
```diff
@@ -965,6 +1487,35 @@ def instructions_to_jcz(instrs: Iterable[InstructionType]) -> Iterator[instructi
...
+                )
+            case InstructionKind.CU:
+                yield from instructions_to_jcz(decompose_cu(instr))
+            case InstructionKind.CSWAP:
+                yield from instructions_to_jcz(
+                    insert_control(instr.control, decompose_swap(Instruction.SWAP(instr.targets)))
+                )
+            case InstructionKind.GPHASE:
+                # Global phase is currently ignored
+                pass
```
matulni: Since this function can return `Instruction.GPHASE` (for instance, from `decompose_rz`), I think it's more fair to do `yield instr` here.

graphix/transpiler.py
```diff
@@ -749,6 +1123,28 @@ def transpile_j_to_rzh(self) -> Circuit:
...
+    def transpile_rzz(self) -> Circuit:
+        """Return an equivalent circuit where all RZZ gates have been replaced with OpenQASM gates."""
+        new_circuit = Circuit(self.width)
+        for instr in self.instruction:
+            match instr.kind:
+                case InstructionKind.RZZ:
+                    new_circuit.extend(decompose_rzz(instr))
+                case _:
+                    new_circuit.add(instr)
+        return new_circuit
```
matulni: Maybe it would it make sense to group these methods and `transpile_j_to_rzh` into a single method `transpile_to_qasm_gates` ?

graphix/transpiler.py
```diff
@@ -921,9 +1320,118 @@ def decompose_rz(instr: instruction.RZ) -> Iterator[instruction.J]:
...
+    yield Instruction.CJ(control=instr.control, target=instr.target, angle=instr.theta)
+    yield Instruction.CJ(control=instr.control, target=instr.target, angle=instr.phi + ANGLE_PI / 2)
+    yield Instruction.CJ(control=instr.control, target=instr.target, angle=0)
+    yield Instruction.P(target=instr.control, angle=instr.gamma - instr.theta / 2)
+
+
+def insert_control(
+    control: int,
+    instrs: Iterable[
+        Instruction.GPHASE | Instruction.X | Instruction.Z | Instruction.J | Instruction.CNOT | Instruction.RZ
```
matulni, line 1344: Why don't we support `Instruction.U` here?

graphix/transpiler.py
```diff
@@ -921,9 +1320,118 @@ def decompose_rz(instr: instruction.RZ) -> Iterator[instruction.J]:
     """
     yield instruction.J(target=instr.target, angle=instr.angle)
     yield instruction.J(target=instr.target, angle=0)
+    yield instruction.GPHASE(-instr.angle / 2)
+
+
+def decompose_u(instr: instruction.U) -> Iterator[instruction.J | Instruction.GPHASE]:
+    """Yield a J decomposition of the U gate.
+
+    The U(θ, φ, λ) gate is decomposed into H·J(φ + 𝜋/2)·J(θ)·J(λ - 𝜋/2) (that is to say, J(0)·J(φ + 𝜋/2)·J(θ)·J(λ - 𝜋/2)).
```
matulni:
```suggestion
    The :math:`U(\theta, \phi, \lambda)` gate is decomposed as :math:`e^{-i \theta/2} \cdot H \cdot J(\phi + \pi/2) \cdot J(\theta) \cdot J(\lambda - \pi/2)`, or equivalently, :math:`J(0) \cdot J(\phi + \pi/2) \cdot J(\theta) \cdot J(\lambda - \pi/2)`.

```

graphix/qasm3_exporter.py
```diff
@@ -62,7 +62,7 @@ def circuit_to_qasm3_lines(circuit: Circuit, *, transpile: bool = True) -> Itera
         An iterator over the OpenQASM 3.0 lines that represent the circuit.
     """
     if transpile:
-        circuit = circuit.transpile_j_to_rzh().transpile_measurements_to_z_axis()
+        circuit = circuit.transpile_cj().transpile_rzz().transpile_j_to_rzh().transpile_measurements_to_z_axis()
```
matulni: I think it would be useful to group all these methods into a single one `Circuit.transpile_to_qasm_gates`

tests/test_instruction.py
```diff
@@ -53,43 +98,87 @@ def visit_axis(self, axis: Axis) -> Axis:
...
-    assert instr_copy != instruction
-    assert instr_visited == instr_copy
-
-
-@pytest.mark.parametrize("instruction", ALL_INSTRUCTIONS)
-def test_visit_angle(instruction: InstructionType) -> None:
-    if not hasattr(instruction, "angle"):
+    instr_visited = instr.visit(visitor, copy=True)
+    assert instr == instr_copy
+    if test_case.name != "GPHASE":
```
matulni, line 111: Why do we skip "GPHASE" ?

matulni (commented):
Thanks for this nice effort. I made some comments.

graphix/instruction.py
thierry-martinez: Yes, indeed, it was a copy-paste mistake. Thanks! I've removed this matrix and added more details about the basis.

graphix/ops.py
thierry-martinez: Done.

graphix/ops.py
thierry-martinez: I've changed `_cast_array` to take an n-ary tuple of parameters instead of just one.

graphix/transpiler.py
thierry-martinez: It was in anticipation of changing the global phase on the resulting state vector, but I've removed that since we haven't implemented that yet.

graphix/instruction.py
thierry-martinez: Done.

graphix/transpiler.py
thierry-martinez: Done.

graphix/transpiler.py
thierry-martinez: Done.

graphix/transpiler.py
thierry-martinez: Done.

graphix/qasm3_exporter.py
thierry-martinez: Done.

tests/test_instruction.py
thierry-martinez: I added this comment:
```
    # instr_visited differs from the original instr_copy iff a qubit
    # attribute has been modified by the visitor, and GPHASE gate has
    # no qubit attribute.
```

graphix/transpiler.py
thierry-martinez: Because we don't need it yet: at that point, I only defined `insert_control` for the gates that the transpiler needs for `instructions_to_jcz`. Adding cases that aren't used there would require specific tests to achieve coverage.

graphix/instruction.py
thierry-martinez: I would prefer to discuss that in a separate issue: https://github.com/TeamGraphix/graphix/issues/598.

graphix/instruction.py
matulni: Thanks!

graphix/ops.py
```diff
@@ -23,6 +24,27 @@
...
+    return np.block(
+        [
+            [np.eye(n), np.zeros((n, n))],
+            [np.zeros((n, n)), gate],
+        ]
+    )
+
+
 class Ops:
     """Basic single- and two-qubits operators."""
```
pranav97nair: We already included the CCX gate and now with the CSWAP we have two 3-qubit gates in `Ops`, so the docstring is outdated. I suggest updating to something like:
```suggestion
    """Elementary quantum gates acting on 1, 2, or 3 qubits."""
```

tests/test_qasm3_exporter_to_graphix_parser.py
```diff
@@ -48,22 +56,42 @@ def test_circuit_to_qasm3(fx_bg: PCG64, jumps: int) -> None:
...
         pytest.skip()
-    check_round_trip(Circuit(3, instr=[instruction]))
+    check_round_trip(Circuit(3, instr=[instr]))
 
 
 def test_j_to_qasm3() -> None:
-    circuit = Circuit(3, instr=[Instruction.J(target=0, angle=ANGLE_PI / 4)])
+    circuit = Circuit(1, instr=[Instruction.J(target=0, angle=ANGLE_PI / 4)])
     qasm = circuit_to_qasm3(circuit)
     parser = OpenQASMParser()
```
pranav97nair: As it stands, `graphix-qasm-parser` is not included in any of the dependency groups for Graphix. I think it would be useful to include any such dependencies that are only required to run certain tests into a `test` group, so we can install them with one `uv` command.

tests/test_qasm3_exporter_to_graphix_parser.py
```diff
@@ -48,22 +56,42 @@ def test_circuit_to_qasm3(fx_bg: PCG64, jumps: int) -> None:
...
+    qasm = circuit_to_qasm3(circuit)
+    parser = OpenQASMParser()
+    parsed_circuit = parser.parse_str(qasm)
+    assert parsed_circuit.instruction == circuit.transpile_to_qasm_gates().instruction
+
+
+def test_rzz_to_qasm3() -> None:
+    circuit = Circuit(2, instr=[Instruction.RZZ(control=0, target=1, angle=ANGLE_PI / 4)])
+    qasm = circuit_to_qasm3(circuit)
+    parser = OpenQASMParser()
```
pranav97nair, line 102: As it stands, `graphix-qasm-parser` is not included in any of the dependency groups for Graphix. Similarly to the dependencies related to Qiskit, I think we should include it in the `dev` group to allow the user to run the entire test suite after `uv sync --extra dev`.

matulni (approved):
Great addition, thanks a lot for the work. LGTM!

graphix/transpiler.py
```diff
@@ -459,6 +491,310 @@ def m(self, qubit: int, axis: Axis) -> None:
...
+        qubit : int
+            target qubit
+        angle : ParameterizedAngle
+            rotation angle in units of π
+        """
+        assert qubit in self.active_qubits
+        self.instruction.append(Instruction.P(target=qubit, angle=angle))
+
+    def u(self, qubit: int, theta: ParameterizedAngle, phi: ParameterizedAngle, lambda_: ParameterizedAngle) -> None:
+        """Apply an U gate.
```
pranav97nair:
```suggestion
        """Apply a universal single-qubit gate.
```

pranav97nair (approved):
Thanks for this substantial effort, Thierry! LGTM.

tests/test_qasm3_exporter_to_graphix_parser.py
thierry-martinez: I added this comment in the beginning of this test file:
```python
# `graphix-qasm-parser` depends on the `graphix` package, so we cannot have
# `graphix-qasm-parser` as a dependency of `graphix`, as this would create
# a dependency loop, which is disallowed by PyPI.
# Instead, `graphix-qasm-parser` is made optional and installed separately
# in `noxfile.py` (for `tests_all` and as a reverse dependency) and in the
# `cov` pipeline.
# The version of `graphix-qasm-parser` to use is defined in two places:
# - `nox`'s `tests_all` session and `cov` pipeline rely on the requirement
#   defined in `.github/qasm-parser-requirements.txt`;
# - `nox`'s reverse dependency check is declared in `noxfile.py` itself.
```

PR #593, opened 2026-08-26
Suppress all warnings emitted by the test suite

thierry-martinez:
This commit suppresses all warnings emitted by the test suite, ensuring that any new warnings introduced by future changes are easier to notice.

It also fixes the warning emitted by `remove_pauli_measurements.py` when isolated nodes are removed, ensuring that it is only emitted for non-Pauli nodes.

Finally, it ensures that warnings emitted by `CliffordMap.from_focused_flow` and `PauliFlow.extract_circuit` are reported to their call sites.


matulni (approved):
thanks for the effort!
LGTM!

pranav97nair (approved):
Thanks for this update, LGTM!

I noticed that `pytest` still throws a `PytestBenchmarkWarning` when running any tests that don't run benchmarks, so everything except `test_linalg.py`. This could be suppressed by updating the tests configuration in `pyproject.toml`, but Mateo suggested that we address this in a future PR to rework benchmarks more generally and separate them from the regular test suite. If you agree with that, I can create an issue on this to be addressed down the line.

thierry-martinez:
Merged. Thanks for the reviews! I open #606 for reintroducing the benchmarks in a dedicated pipeline.

PR #607, opened 2026-09-21
Fix #493: New base class `Simulable`

thierry-martinez:
The new base class `Simulable` is implemented by `Pattern`, `StandardizedPattern`, `OpenGraph`, `PauliFlow`, and `XZCorrections`. The `simulate` method runs the simulation on a pattern derived from the object, applying usual optimization passes by default. Optimizations can be disabled by passing `optimized=False`.


graphix/flow/core.py
```diff
@@ -661,6 +663,10 @@ def to_xzcorrections(self) -> XZCorrections[_AM_co]:
 
         return XZCorrections(self.og, x_corrections, z_corrections, self.partial_order_layers)
 
+    @override
+    def to_pattern(self: PauliFlow[Measurement]) -> Pattern:
```
matulni: Would it be worth to accept an optional parameter `total_measurement_order` and pass it directly to `to_pattern`?

graphix/pattern.py
```diff
@@ -1913,11 +1819,8 @@ def remove_pauli_measurements(
                 before the other measurements. If ``copy`` is ``False``,
                 the result is ``self``.
         """
-        from graphix.remove_pauli_measurements import PauliPushingCut, remove_pauli_measurements  # noqa: PLC0415
-
-        standardized_pattern = optimization.StandardizedPattern.from_pattern(self)
-        cut = PauliPushingCut.from_standardizedpattern(standardized_pattern, stacklevel=stacklevel + 1)
-        standardized_pattern = remove_pauli_measurements(cut, stacklevel=stacklevel + 1)
+        standardized_pattern = self.to_standardizedpattern()
+        standardized_pattern = standardized_pattern.remove_pauli_measurements(stacklevel=stacklevel + 1)
```
```suggestion
        standardized_pattern = self.to_standardizedpattern().remove_pauli_measurements(stacklevel=stacklevel + 1)
```

graphix/simulator.py
```diff
@@ -583,3 +589,185 @@ def _initialize_backend(
...
+        -------
+        Pattern
+            Optimized pattern.
+        """
+        standardized_pattern = self.to_standardizedpattern()
+        standardized_pattern = standardized_pattern.minimize_space()
+        standardized_pattern2 = (
+            standardized_pattern.infer_pauli_measurements()
+            .remove_pauli_measurements(stacklevel=stacklevel + 1)
+            .minimize_space()
```
matulni: Would it be worth to check if there are any `PauliMeasurement` after calling `remove_pauli_measurements` and `minimize_space` ? Those passes are potentially expensive, and a checking if there are 
Pauli measurements is simply linear in number of commands.

graphix/simulator.py
```diff
@@ -583,3 +589,185 @@ def _initialize_backend(
...
+            Stack level to use for warnings. Defaults to 1, meaning that warnings
+            are reported at this function's call site.
+
+        Returns
+        -------
+        Pattern
+            Optimized pattern.
+        """
+        standardized_pattern = self.to_standardizedpattern()
+        standardized_pattern = standardized_pattern.minimize_space()
```
matulni:
```suggestion
        standardized_pattern = self.to_standardizedpattern().minimize_space()
```

